PR #2, opened 2024-11-26
FIX: New template library support (specifically Ubuntu 24.04)

jglanz:
Fixed issues related to building with slightly newer template libraries

- including <cstdint> in `wast_to_wasm.cpp`
- forcing template inference to `path` instead of string in `clio`


dtaghavi:
Looks good, merging.

PR #11, opened 2025-03-17
Block State Log

brianjohnson5972:
Wcore 49
Converted Block log classes to use templates to allow storing the block state log in a parallel log and index to the existing block log.

### DIFF
--- a/libraries/chain/include/sysio/chain/block_log.hpp
+++ b/libraries/chain/include/sysio/chain/block_log.hpp
@@ -120,3 +120,5 @@ namespace sysio { namespace chain {
          std::unique_ptr<detail::block_log_impl<stored_type> > my;
    };
 } }
+
+#include <eosio/chain/block_log.cpp>
--- a/libraries/chain/include/sysio/chain/block_log.hpp
+++ b/libraries/chain/include/sysio/chain/block_log.hpp
@@ -121,4 +121,4 @@ namespace sysio { namespace chain {
    };
 } }
 
-#include <eosio/chain/block_log.cpp>
+#include <sysio/chain/block_log.cpp>
--- a/libraries/chain/include/sysio/chain/block_log.hpp
+++ b/libraries/chain/include/sysio/chain/block_log.hpp
@@ -6,7 +6,10 @@
 
 namespace sysio { namespace chain {
 
-   namespace detail { struct block_log_impl; }
+   namespace detail {
+      template<typename StoredType>
+      struct block_log_impl;
+   }
 
    /* The block log is an external append only log of the blocks with a header. Blocks should only
     * be written to the log after they irreverisble as the log is append only. The log is a doubly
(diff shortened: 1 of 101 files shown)


brianjohnson5972:
@bearcubsvet @valthon @dtaghavi

valthon:
Don't like the cpp file in the `include` directory.  Appears that we can move it back out and explicitly instantiate the two implementations without issue.  When doing so, I found (and hopefully fixed) an issue with `full_validate_blocks` not using the templated type.

```
diff --git c/libraries/chain/CMakeLists.txt i/libraries/chain/CMakeLists.txt
index 2c6dca1286..9fb96bf28f 100644
--- c/libraries/chain/CMakeLists.txt
+++ i/libraries/chain/CMakeLists.txt
@@ -94,6 +94,7 @@ add_library( sysio_chain
              controller.cpp
              authorization_manager.cpp
              resource_limits.cpp
+             block_log.cpp
              transaction_context.cpp
              sysio_contract.cpp
              sysio_contract_abi.cpp
diff --git c/libraries/chain/include/sysio/chain/block_log.cpp i/libraries/chain/block_log.cpp
similarity index 99%
rename from libraries/chain/include/sysio/chain/block_log.cpp
rename to libraries/chain/block_log.cpp
index c754d70a46..50df1d2113 100644
--- c/libraries/chain/include/sysio/chain/block_log.cpp
+++ i/libraries/chain/block_log.cpp
@@ -1,4 +1,5 @@
 #include <sysio/chain/block_header_state.hpp>
+#include <sysio/chain/block_log.hpp>
 #include <sysio/chain/block_log_config.hpp>
 #include <sysio/chain/exceptions.hpp>
 #include <sysio/chain/log_catalog.hpp>
@@ -1393,7 +1394,7 @@ namespace sysio { namespace chain {
       SYS_ASSERT(!is_currently_pruned(), block_log_exception, "pruned block log cannot be repaired");
       try {
          try {
-            signed_block entry;
+            StoredType entry;
             while (remaining() > 0 && block_num < last_block_num) {
                std::tie(block_num, block_id) = full_validate_block_entry<StoredType>(block_num, block_id, entry);
                if (block_num % 1000 == 0)
@@ -1856,4 +1857,8 @@ namespace sysio { namespace chain {
       }
    }

+   // Force instantiation of the two supported block_log implementations
+   template class block_log<block_header_state>;
+   template class block_log<signed_block>;
+
 }} // namespace sysio::chain
diff --git c/libraries/chain/include/sysio/chain/block_log.hpp i/libraries/chain/include/sysio/chain/block_log.hpp
index b02efc2667..fc00882d9f 100644
--- c/libraries/chain/include/sysio/chain/block_log.hpp
+++ i/libraries/chain/include/sysio/chain/block_log.hpp
@@ -119,6 +119,4 @@ namespace sysio { namespace chain {
    private:
          std::unique_ptr<detail::block_log_impl<stored_type> > my;
    };
-} }
-
-#include <sysio/chain/block_log.cpp>
+} }
\ No newline at end of file
```

libraries/chain/include/sysio/chain/block_log.hpp
```diff
@@ -82,9 +78,8 @@ namespace sysio { namespace chain {
 
          static std::filesystem::path repair_log( const std::filesystem::path& data_dir, uint32_t truncate_at_block = 0, const char* reversible_block_dir_name="" );
 
-         using chain_context = std::variant<genesis_state, chain_id_type>;
-         static std::optional<chain_context> extract_chain_context(const std::filesystem::path& data_dir,
```
valthon, line 86: why get rid of this `using` definition?

valthon (commented):
other than the previously mentioned patch, everything generally looks solid

brianjohnson5972:
@valthon thanks for tracking that down. I am not sure why when I looked into it last week that the errors seemed much worse (I think it was where I instantiated the explicit templates), but glad you fixed this, because it was obviously worse having all that in the include path for so many files.
 @dtaghavi I'd like a quick check from one of you

valthon (approved):

PR #16, opened 2025-04-04
wcore-74: Removing compression from packed_transaction::digest().

brianjohnson5972:
Removing compression since it is set internal to the node, on a transaction by transaction basis and is not retrievable for validation purposes.


brianjohnson5972:
@dtaghavi @qhool

dtaghavi (approved):
Looks fine to me

brianjohnson5972:
Chuy found that the compression is provided in the transaction trace, so this change isn't needed.

PR #20, opened 2025-06-05
renamed leap-util to sys-util

bearcubsvet:
Code builds and works; tested with `wire-install`

Screenshot 2025-06-05 at 3 31 20 PM


bearcubsvet:
> I'd like to know that the block_log_*** tests were able to run. You can run them by running "ctest -R block_log" and any tests that have that string in it will run.

mmmm forgot to check that, will do and will report

bearcubsvet:
Screenshot 2025-06-10 at 1 21 06 PM

PR #24, opened 2025-06-19
Improvements for integration tests

brianjohnson5972:
Changed label for tests that we expect to fail and marked some longer tests with medium_running_tests to reduce the runtime for this group


bearcubsvet:
> I appreciate switching the label for long running tests that are (currently) failing, as that's consistent with the other changes we've made to narrow down the "must pass" suite. However, I'm not in favor of adding `medium_running_tests` as it really feels redundant with the `long_running_tests` label. Too much nuance here is just asking for missed tests and confused future developers. Either it's a long-running test or it's fast enough to run in our normal set.

Do you have any flag name in mind @valthon ?

brianjohnson5972:
I think he wants medium and long to combine under long_running_tests.
I'm fine with that, but my reason for creating the new one is that both
take about 15 minutes to run, so it was to prevent having to lock up your
machine, or a cloud resource, for 30 minutes

brianjohnson5972:
You sold me on it

dtaghavi (changes requested):
Just remove the medium-running from README to match your latest commit, otherwise looks good to me.

bearcubsvet (approved):

dtaghavi (approved):

PR #27, opened 2025-07-16
wcore-48 - Added Root processing Snapshot

brianjohnson5972:
Added the code to write the S Root table to a snapshot that is being created and creating the table using a passed in snapshot.


valthon (approved):
would be nice if there was a test, but ...

PR #39, opened 2025-08-22
Remove deferred transaction support

heifner:


brianjohnson5972 (approved):

dtaghavi:
Wouldn't we want to recompile any and every contract deployed to our chain anyways?

dtaghavi:
Since we added new key types to our CDT, if those contracts need to be able to accept SIG_ED / SIG_EMs they'd need to be recompiled. Are there specific contracts you are thinking about?

PR #43, opened 2025-08-31
ubuntu 24 - build support

jglanz:
The core issue was the need for `llvm-11` source compilation with the correct options/features enabled.  To do so, required `gcc-10`, which may need to be built from source itself in the near future, once installed and CMAKE was configured to use it as the compiler and with an install prefix path configured, everything worked:

```sh
cmake \
-DCMAKE_C_COMPILER=gcc-10 \
-DCMAKE_CXX_COMPILER=g++-10 \
-DCMAKE_INSTALL_PREFIX=/opt/llvm/llvm-11 \
-DCMAKE_PREFIX_PATH=/opt/llvm/llvm-11 \
...
```
The scripts for building llvm properly are under `scripts/llvm-11`.

The Dockerfile was also updated to build on Ubuntu 24.04. Additionally it was made multi-stage for build performance and efficiency

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Wire Sysio
 
-Wire Sysio is a fork of Leap, a C++ implementation of the [Antelope](https://github.com/AntelopeIO) protocol. It contains blockchain node software and supporting tools for developers and node operators.
+Wire Sysio is a fork of Leap, a C++ implementation of the [Antelope](https://github.com/AntelopeIO) protocol. It
+contains blockchain node software and supporting tools for developers and node operators.
 
 ## Branches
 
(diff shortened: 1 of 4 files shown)


jglanz:
# Apple & Ubuntu 24 - Ready for review

- Apple Silicon support via/Docker
- Docker build script
  Example usage: `./scripts/docker-build.sh --
target=app-build-local --tag=wire/sysio:local`
- Docs updated regarding docker & host builds

etc/docker/Dockerfile
```diff
@@ -0,0 +1,136 @@
...
+        libssl-dev \
+        libusb-1.0-0-dev \
+        pkg-config \
+        python3 \
+        zlib1g-dev \
+        libbz2-dev \
+        liblzma-dev \
+        libncurses5-dev \
+        libzstd-dev \
+    		python3.12
```
heifner, line 54: I think we should remove `libboost-all-dev` from this list and below as boost is submoduled into Leap. Having it installed could cause confusion.
I would also add `jq` and `file` either here or in the ubuntu deps.

README.md
```diff
@@ -1,6 +1,7 @@
 # Wire Sysio
```
heifner, line 1: Seems to me we may want some indication this is for development use or experimental use. We typically do not want block producers running nodes in docker because of the poor performance. This is the main reason docker was not provided in Leap. It was removed because it is not a recommended env.

bearcubsvet:
@jglanz should modify .github/workflows/unit-test.yaml as it used to depend on the Dockerfile that was removed

I build step now would be: 
```
- name: Build Docker image
   run: |
    ./scripts/docker-build.sh --target=app-build-repo --tag=wire-sysio-build -- \
      --build-arg BRANCH=${{ github.head_ref }} \
      --build-arg REPO=${{ github.event.pull_request.head.repo.clone_url }}
  ```
  
 want to see that change trigger the action

File: scripts/docker-build.sh [hunk 3]
```diff
@@ -0,0 +1,117 @@
...
+      usage
+      exit 0
+      ;;
+    *)
+      echo "Unknown argument: $1" 1>&2
+      usage
+      exit 1
+      ;;
+  esac
+done
```
bearcubsvet, line 41: extend the arg parser to accept branch and repo 

```
    --repo=*)
      REPO="${1#--repo=}"
      shift
      ;;
    --branch=*)
      BRANCH="${1#--branch=}"
      shift
      ;;
      ```

PR #46, opened 2025-09-16
Update account creation to use gifted RAM from sysio

heifner:
* See https://wire-network.atlassian.net/wiki/spaces/WNC/pages/24084489/Account+Creation
* Create a policy for `sysio.acct` to track gifted RAM amount. 
* Failing tests are addressed in follow-on PRs.

Once approved I'll update wire-system-contracts with the contracts here.

### DIFF
--- a/libraries/testing/tester.cpp
+++ b/libraries/testing/tester.cpp
@@ -568,7 +568,7 @@ namespace sysio { namespace testing {
                                                 vector<permission_level>{{config::system_account_name, config::active_name}},
                                                 fc::mutable_variant_object()
                                                 ("account", a)
-                                                ("ram_bytes", 2800) // matches ram_newaccount of system contract
+                                                ("ram_bytes", newaccount_ram) // matches newaccount_ram of system contract
                                                 ("net_weight", -1)
                                                 ("cpu_weight", -1)));
          }
--- a/libraries/testing/include/sysio/testing/tester.hpp
+++ b/libraries/testing/include/sysio/testing/tester.hpp
@@ -165,6 +165,7 @@ namespace sysio { namespace testing {
          static const uint32_t DEFAULT_BILLED_CPU_TIME_US = 2000;
          static const fc::microseconds abi_serializer_max_time;
 
+         static constexpr uint64_t newaccount_ram = 2800; // Should match sysio.system native newaccount_ram
          static constexpr auto NODE_DADDY = "nodedaddy"_n;
          bool has_roa = false;
 


libraries/chain/sysio_contract.cpp
```diff
@@ -109,34 +108,23 @@ void apply_sysio_newaccount(apply_context& context) {
...
+                                                                       std::move(create.active), context.trx_context.is_transient() );
 
       context.control.get_mutable_resource_limits_manager().initialize_account(create.name, context.trx_context.is_transient());
 
-      // Determine if this is a system account
-      bool is_system_account = (create.name == config::system_account_name) ||
-                               (name_str.size() > 5 && name_str.find("sysio.") == 0);
-
-      // If it's not a system account, set CPU, NET, RAM to 0
-      if (!is_system_account) {
```
dtaghavi, line 124: Everything else looks good from my perspective. Im curious, if this is taken out are non-system accounts appropriately given 0 NET / CPU still?

heifner:
> Looks good to me. One consideration is when creating the new account and gifting the RAM. We _might_ want to add a `reslimit` row for this account, the one consideration I have for that is that `reslimit` wouldn't be a sum of all policies issued then, not sure if that's a concern. It would make `reslimit` accurate in terms of total allocated resources to an account though.

That seems like a good idea to me to update the `sysio.acct` `reslimit`.

libraries/testing/contracts/sysio.roa/sysio.roa.cpp
```diff
@@ -3,6 +3,45 @@
 
 namespace sysio {
 
+    bool is_sysio_account(const name& account) {
+        return account.prefix() == "sysio"_n;
+    }
+
+    roa::resources_t roa::update_reslimit(const name& owner, const asset& netWeight, const asset& cpuWeight, int64_t ram_bytes) {
```
valthon, line 10: I think this function should be named `increase_reslimit` to better match it's semantics.  When reading `update` I tend to assume the passed in arguments are absolute target values, not incremental increases.   I also usually expect an `update_` function to be idempotent, which this is definitely not.

I also wonder whether we want a separate `set_reslimit` function for the cases when we don't want to increment. Initial roa setup might be a good case for this, as it makes me a bit uncomfortable to think we might double-initialize in certain odd scenarios.

libraries/testing/tester.cpp
```diff
@@ -556,6 +557,23 @@ namespace sysio { namespace testing {
...
+      if (include_ram_gift) {
+         // if bios contract with setalimits available then provide ram for account which is similar to what sysio.system contract does
+         const auto& acnt = control->get_account(config::system_account_name);
+         auto abi = acnt.get_abi();
+         if (std::ranges::any_of(abi.actions, [](const auto& action) { return action.name == "setalimits"_n; })) {
+            trx.actions.emplace_back(get_action(config::system_account_name, "setalimits"_n,
+                                                vector<permission_level>{{config::system_account_name, config::active_name}},
+                                                fc::mutable_variant_object()
+                                                ("account", a)
+                                                ("ram_bytes", 2800) // matches ram_newaccount of system contract
```
valthon, line 571: could we use an explicit constant for this value?  Ideally, one that's shared w/ the system contract, but if that's not feasible I'd still like to see it as some extremely explicit const.

unittests/auth_tests.cpp
```diff
@@ -362,9 +362,7 @@ try {
    BOOST_CHECK_EXCEPTION(chain.create_account(name("aaaaaaaaaaaaa")), action_validate_exception,
                          fc_exception_message_is("account names can only be 12 chars long"));
 
-   // Creating a new account with non-privileged account, should fail
-   BOOST_CHECK_EXCEPTION(chain.create_account(name("dandy.joe"), name("joe")), action_validate_exception,
-                         fc_exception_message_is("Only privileged accounts can create new accounts"));
+   // Note: Only privileged accounts can create accounts enforced by sysio.system contract
```
valthon, line 365: assuming we switch back to enforcing this where it was originally, can we add this assert back in?

valthon (changes requested):

libraries/testing/contracts/sysio.roa/sysio.roa.cpp
heifner: https://github.com/Wire-Network/wire-sysio/pull/46/commits/321593acee748039a3bdf3af0edfaf81d55e9f4e

valthon (approved):

PR #49, opened 2025-09-17
WIRE token contract

heifner:
* Updated `sysio.token` contract to have `sysio` pay for RAM. This requires the contract to be privileged.
* There is likely more changes we will make to the `sysio.token` contract as requirements harden/change.

Once this PR is approved, will update `wire-system-contracts`.


dtaghavi (dismissed):
Are we in some way limiting other tokens from being created on this contract? If the contract / sysio is paying is there anything stopping users from just creating other tokens on this contract?

dtaghavi (approved):

PR #50, opened 2025-09-19
Proposal: Allow RAM move

heifner:
Change will be discussed in a meeting on 9/16/2025.

Allow RAM movement as long as the delta of RAM usage for a user does not increase.

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -100,30 +100,7 @@ void apply_context::exec_one()
                   control.get_wasm_interface().apply( receiver_account->code_hash, receiver_account->vm_type, receiver_account->vm_version, *this );
                } catch( const wasm_exit& ) {}
             }
-
-            if( !privileged ) {
-               const size_t checktime_interval = 10;
-               size_t counter = 0;
-               bool not_in_notify_context = (receiver == act->account);
-               const auto end = _account_ram_deltas.end();
-               for( auto itr = _account_ram_deltas.begin(); itr != end; ++itr, ++counter ) {
-                  // wlog("pending RAM delta: ${account} ${delta}", ("account", itr->account)("delta", itr->delta) );
-                  if( counter == checktime_interval ) {
-                     trx_context.checktime();
-                     counter = 0;
-                  }
-                  if( itr->delta > 0 && itr->account != receiver ) {
-                     SYS_ASSERT( not_in_notify_context, unauthorized_ram_usage_increase,
-                                 "unprivileged contract cannot increase RAM usage of another account within a notify context: ${account}",
-                                 ("account", itr->account)
-                     );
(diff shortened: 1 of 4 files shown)


libraries/chain/apply_context.cpp
```diff
@@ -188,6 +165,55 @@ void apply_context::exec_one()
...
+         SYS_ASSERT( has_authorization( itr->account ), unauthorized_ram_usage_increase,
+                     "unprivileged contract cannot increase RAM usage of another account that has not authorized the action: ${account}",
+                     ("account", itr->account)
+         );
+      }
+      auto& payer = itr->account;
+      auto& ram_delta = itr->delta;
+      if (payer != receiver && ram_delta > 0) {
+         if (receiver == config::system_account_name) {
+         } else if (payer == config::system_account_name && is_privileged()) {
```
dtaghavi, line 193: Are lines 192 / 193 for something we will want in the future?

valthon: might make sense to add a `// no-op`  comment to make sure future readers of the code understand the intent

libraries/chain/apply_context.cpp
```diff
@@ -188,6 +165,55 @@ void apply_context::exec_one()
...
+               if( auth.actor == payer && auth.permission == config::sysio_payer_name ) {
+                  payer_found = true;
+                  break;
+               }
+               if ( auth.actor == config::system_account_name ) {
+                  // If the payer is the system account, we don't enforce explicit payer authorization.
+                  // This avoids a lot of changes for tests and sysio should not be calling untrusted contracts
+                  // or spending user's RAM unexpectedly.
+                  payer_found = true;
+                  break;
```
valthon, line 207: optimization opportunity, no need to do the auth loop for additional deltas.  but maybe the explicit flow is better, so feel free to leave as-is

unittests/protocol_feature_tests.cpp
```diff
@@ -628,14 +628,10 @@ BOOST_AUTO_TEST_CASE(steal_my_ram) {
...
-            ("len1", 0)
-            ("len2", 10)
-            ("payer", alice_account)
-         ),
-         unsatisfied_authorization,
-         fc_exception_message_is("Requested payer alice did not authorize payment")
+      c.push_action(tester1_account, "setdata"_n, bob_account, mutable_variant_object()
+         ("len1", 0)
+         ("len2", 10)
+         ("payer", alice_account)
```
valthon, line 634: I'd love to see an attempt to move from `len1 10 len2 0` to `len1 0 len2 20` fail before we do the successful net-0 transfer.  Also, would be nice to reverse the transfer intentionally losing ram (ie, to `len1 5 len2 10`) and verify that the ram is allocated/refunded appropriately. (if that's done below and I missed it, feel free to disregard)

valthon (dismissed):

valthon (approved):

PR #54, opened 2025-09-22
Fix unit tests - WIP

heifner:
WIRE-33

- Removed `produce_block()` from tester `init_roa()` as there are a number of tests that assume a given number of blocks at the start of the test.
  - Update sroot tests for these missing `produce_block()` calls.
- Update `chain_plugin` to remove: `self_delegated_bandwidth`, `refund_request`, `voter_info`, and `rex_info` from `get_account` info as these are not supported on Wire. This will break Antelope `cleos get_account` as `cleos` expects these fields. Seems like it is worth breaking compatibility here for a better Wire user experience (not showing these empty values).
- Update `chain_plugin` `get_account` to use ROA `reslimit` instead of the old system contract `userres`.
-


tests/get_table_tests.cpp
```diff
@@ -318,147 +318,7 @@ BOOST_FIXTURE_TEST_CASE( get_table_test, validating_tester ) try {
 
 } FC_LOG_AND_RETHROW()
 
-BOOST_FIXTURE_TEST_CASE( get_table_by_seckey_test, validating_tester ) try {
```
heifner, line 321: There are existing secondary key tests that also test this same functionality.

dtaghavi:
Is this as expected?
image

heifner:
> Is this as expected?

No, I'll take a look. Thanks.

PR #64, opened 2025-10-01
Remove unneeded clio system commands

heifner:
WIRE-38

Remove the `clio system` commands related to buyram, delegate bandwidth, and voting.
This PR includes some work toward getting `bios-boot-tutorial.py` working. That work is only partly done as it is likely we will just remove the `bios-boot-tutorial.py` as `wire-install` will take its place.


valthon (approved):
love these PRs where it's mostly just lines removed

PR #65, opened 2025-10-03
Do not allow actions to accounts with no contracts

heifner:
WIRE-17

Native EOSIO/Antelope allows any action as a no-op to accounts with no contract.
- on-notify (require-recipient) requires this functionality.
- Native actions and `onblock` require this functionality.

Assert on input actions to accounts with no contract. 
- Allow `sysio` to allow for `onblock` and native actions
- Allow inline and notify to non-existing contracts.

- Removed `clio push action --force-unique` as `nonce` action to `sysio.null` is not allowed.
   - Options considered.
     - Add a contract to a `sysio.noop` account that requires explicit payer. This would restrict the functionality to explicit payer.. yuck.
     - Rename unused `transaction_header` `delay_sec` to `nonce` and allow it to be any value. This would create a JSON difference in the API of transactions which could cause problems with existing tooling. Also breaks existing `sysio.msig` contract use of `delay_sec`. Could use the `nonce` in `sysio.msig` as a `delay_sec` but that seems like a huge hack. Could also rewrite the `sysio.msig` contract to not use `delay_sec`, but rather pass it as an option to the action. This would break existing MSIG tooling. 
     - Remove `clio push action --force-unique`
       - Require users to make their own transactions unique manually. This is as easy as just indicating a different TAPOS or diff `max_net_usage_words`  or diff `max_cpu_usage_ms` or diff `expiration`.


dtaghavi (approved):
Seems good to me.

PR #80, opened 2025-10-07
Fix `net_plugin_impl` dtor order

heifner:
See https://github.com/AntelopeIO/leap/pull/2300

Destroy thread_pool's executor after users of it.

> `net_plugin_impl` currently destroys its `named_thread_pool` before a number of users of it such as expire_timer. Those users can, and do, access their executor on their destruction which in this case is accessing an already destroyed executor.
> 
> Move the `named_thread_pool` up to the top so it's destroyed last. A nice side effect of this is being able to clean up a number of the `unique_ptr`s to just being a plain member.


dtaghavi (approved):
If the above comment doesn't matter, looks good.

PR #216, opened 2026-02-20
Add missing COMPONENT to all install() rules for correct deb packaging

heifner:
With CPACK_DEB_COMPONENT_INSTALL enabled, install() rules without a COMPONENT go to "Unspecified" which is excluded from both the `base` and `dev` .deb packages. This adds COMPONENT to all 21 install() rules that were missing it:

- `trace_api_util` binary -> `base` (was silently excluded from .deb)
- Licenses directory -> `base` and `dev`
- Headers, static libraries, and cmake configs -> `dev` (`libfc, libfc-lite, sysio_chain, builtins, sysio_testing, wasm-jit, vcpkg includes/libs`, sysio cmake configs)


huangminghuang (changes requested):
COMPONENT dev EXCLUDE_FROM_ALL now belongs only to PUBLIC_HEADER, because install(TARGETS) options after artifact kinds apply only to current artifact group. ARCHIVE stays Unspecified, so default static build still omits libsysio_testing.a from sysio-dev. Fix: put COMPONENT dev EXCLUDE_FROM_ALL in each artifact group, or before artifact-kind groups as generic target options where valid.

PR #246, opened 2026-03-12
Snapshot on-chain attestation

heifner:
## Summary

  - Builds on #245 
   - **On-chain snapshot attestation contract** in `sysio.system`: registered snapshot providers submit BLAKE3 root hashes via `votesnaphash`, attested record created when configurable quorum is reached
   - **Snapshot provider mode** in nodeop: `snapshot-provider-account` auto-schedules snapshots every 25,000 blocks and submits `votesnaphash` transactions; role exclusivity enforced (cannot also be a producer)
   - **Bootstrap verification**: nodes started with `--snapshot` verify hash against on-chain attestation after syncing to head (warn on missing record, fatal on mismatch)
  - See `docs/snapshot-serving-plan.md` for additional info.

   ### Contract actions (`sysio.system`)
   | Action | Auth | Description |
   |--------|------|-------------|
   | `regsnapprov` | producer | Register a snapshot provider account (producer rank ≤ 30) |
   | `delsnapprov` | account | Deregister as snap_account or producer |
   | `votesnaphash` | snap_account | Submit snapshot hash vote for a block |
   | `setsnpcfg` | sysio | Configure min_providers and threshold_pct |
   | `getsnaphash` | read-only | Query attested record by block_num |


plugins/chain_plugin/src/chain_plugin.cpp
```diff
@@ -1351,6 +1367,93 @@ bool chain_plugin::accept_transactions() const {
...
+      }
+
+      // Deserialize the snap_record row:
+      //   uint64_t block_num + checksum256 block_id + checksum256 snapshot_hash + uint32_t attested_at_block
+      // We need the snapshot_hash at offset 8+32 = 40, length 32
+      const auto& row = it->value;
+      constexpr size_t hash_offset = sizeof(uint64_t) + 32; // block_num(8) + block_id(32)
+      constexpr size_t hash_size = 32;
+
+      if (row.size() < hash_offset + hash_size) {
```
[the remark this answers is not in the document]
heifner: https://github.com/Wire-Network/wire-sysio/pull/246/commits/1854a338a13fa80d478fbdb5f326aeab08f7c09b

plugins/producer_plugin/src/producer_plugin.cpp
```diff
@@ -1621,6 +1725,18 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
    _snapshot_scheduler.set_db_path(_snapshots_dir);
```
huangminghuang, line 1725: This loads persisted snapshot schedules, then plugin_startup() always calls schedule_snapshot() again at plugins/producer_plugin/src/producer_plugin.cpp:1801. Since schedule_snapshot() rejects duplicate (block_spacing, start_block_num, end_block_num) requests, a provider node that restarts with the saved auto-schedule still in snapshot-schedule.json will throw duplicate_snapshot_request during startup. Provider mode should reuse the existing auto request or make this scheduling idempotent.

plugins/chain_plugin/src/chain_plugin.cpp
huangminghuang, line 1231: This verifies on the first irreversible block greater than the loaded snapshot block, and verify_snapshot_attestation() clears snapshot_loaded_block_num before reading the table at plugins/chain_plugin/src/chain_plugin.cpp:1473. If the snapshot file was taken at block N, the snaprecords row can only be created by later vote transactions, so the first LIB > N check may run before that row has synced into state. The code then logs “Skipping snapshot verification” at plugins/chain_plugin/src/chain_plugin.cpp:1502 and never retries, leaving an actually attestable snapshot unverified. The integration test avoids this by waiting for the attestation to be irreversible before restart, but a real bootstrap from a fresh snapshot can hit this race.

huangminghuang (changes requested):

plugins/chain_plugin/src/chain_plugin.cpp
```diff
@@ -1351,6 +1367,93 @@ bool chain_plugin::accept_transactions() const {
...
+         app().quit();
+         return;
+      }
+
+      ilog("Snapshot attestation verified successfully for block #{}: hash {} matches on-chain record",
+           snap_block_num, snapshot_loaded_root_hash.str());
+
+   } catch (const fc::exception& e) {
+      elog("Error verifying snapshot attestation for block #{}: {}. "
+           "To continue syncing without snapshot verification, restart without the --snapshot option.",
```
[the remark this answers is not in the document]
heifner: Done in a4d4bbad59. Both the hash-mismatch path and the catch blocks now drop the "restart without --snapshot" suggestion and instead report that the node has stopped and strongly recommend deleting chain state and acquiring a fresh snapshot from a trusted source before restarting.

plugins/producer_plugin/src/producer_plugin.cpp
heifner: Fixed in a4d4bbad59. Added `snapshot_scheduler::find_snapshot_request(block_spacing, start_block_num, end_block_num)`; in provider mode `plugin_startup()` now reuses the persisted request when an identical one already exists instead of calling `schedule_snapshot()` again. The scheduler dup check uses the same helper, so a provider restart with the auto-schedule already in `snapshot-schedule.json` no longer throws `duplicate_snapshot_request`.

plugins/chain_plugin/src/chain_plugin.cpp
heifner: Fixed in a4d4bbad59. `verify_snapshot_attestation()` no longer resets `snapshot_loaded_block_num` up front or gives up after the first finalized block past the snapshot height. It now retries on each subsequent finalized block until either the `snaprecords` row appears (verify) or the node has caught up to the live chain tip (then logs the skip). This handles the fresh-bootstrap case where the vote transactions have not synced into state when LIB first crosses the snapshot block.

plugins/chain_plugin/src/chain_plugin.cpp
```diff
@@ -1451,6 +1471,93 @@ bool chain_plugin::accept_transactions() const {
...
+      p.table       = "snaprecords";
+      p.lower_bound = std::to_string(snap_block_num);
+      p.limit       = 1;
+      p.values_only = true;
+
+      // verify_snapshot_attestation() runs on the main thread from the irreversible-block
+      // handler during sync, so read_table_rows takes its inline main-thread fast path and
+      // never consults the shutdown flag.
+      static const std::atomic<bool> not_shutting_down{false};
+      auto rows = app().get_plugin<chain_plugin>().read_table_rows(
```
huangminghuang: verify_snapshot_attestation() now calls read_table_rows() for snaprecords, but read_table_rows() throws if the table is not in the system ABI. That means loading a snapshot on a chain whose system contract predates snapshot attestation goes through the catch block at line 1544 (line 1544), logs FATAL, and quits. The PR summary/design says missing attestation table/record should warn to support non-attesting chains. This should distinguish “table/ABI absent” from decode/query errors and keep the warning behavior.

plugins/chain_plugin/src/chain_plugin.cpp
```diff
@@ -1451,6 +1471,93 @@ bool chain_plugin::accept_transactions() const {
...
+
+      if (!have_record) {
+         // The snaprecords row is created by votesnaphash transactions that only land a few blocks
+         // after the snapshot height, so on a fresh bootstrap it may not have synced into state by
+         // the time the first finalized block past the snapshot height arrives. Keep retrying on
+         // later finalized blocks; only conclude the height was never attested once the node has
+         // caught up to the live chain tip. snapshot_loaded_block_num is left set so the caller
+         // retries. This is local, non-consensus bookkeeping, so comparing the finalized block time
+         // against wall-clock to detect "caught up" is appropriate.
+         const auto caught_up_tolerance = fc::seconds(30);
```
huangminghuang, line 1563: the retry fix can still permanently skip verification too early. If a node loads a fresh snapshot and catches up to live tip before providers’ votesnaphash transaction has landed, caught_up becomes true, snapshot_loaded_block_num is reset, and a later attestation record is never checked. The previous race is reduced but not eliminated. Consider keeping verification pending for a deterministic block/time grace window after catch-up, or requiring an explicit “unattested snapshot accepted” terminal condition.

contracts/sysio.system/src/snapshot_attest.cpp
```diff
@@ -0,0 +1,176 @@
...
+   }
+
+   // Find or create the vote entry for this block_num + snapshot_hash
+   snap_votes_table votes(get_self());
+   auto by_bn = votes.get_index<"byblocknum"_n>();
+
+   uint64_t vote_id     = 0;
+   uint32_t voter_count = 0;
+   bool     found       = false;
+   for (auto itr = by_bn.lower_bound(static_cast<uint64_t>(block_num));
```
huangminghuang, line 98: votes are grouped by block_num + snapshot_hash, ignoring block_id, even though snap_vote stores block_id and snap_record attests it. If two votes at the same height carry the same snapshot hash but different block ids, they aggregate toward quorum and the resulting record keeps whichever block id was first inserted. The contract should include block_id in the vote match, and when a record already exists it should check both hash and block id.

huangminghuang (commented):
Was snapshot_attest_test.py intended to prove that chain_plugin logs successful verification? It currently verifies the table is readable after sync, but it would not fail if verify_snapshot_attestation() skipped early.

plugins/chain_plugin/src/chain_plugin.cpp
heifner: Fixed in 65eca55d69. One correction on the mechanism: `read_table_rows()` cannot throw out of the scan -- `run_scan` catches everything, logs, and collapses to an empty result -- so the missing table never actually reached the FATAL catch block; it surfaced as a query elog on every finalized block until catch-up, then the generic "no record" warning. That said, the substance of the comment stands: the table/ABI-absent case deserved explicit handling rather than incidental behavior. `verify_snapshot_attestation()` now checks the system contract's ABI for the `snaprecords` table before attempting the read. While syncing, an absent table just retries on the next finalized block (no error spam, and a mid-sync system-contract upgrade that adds attestation support is still picked up); once caught up, it terminally downgrades to a dedicated warning ("this chain does not support snapshot attestation") instead of entering the missing-record grace window. Decode errors on a present record still take the FATAL path.

plugins/chain_plugin/src/chain_plugin.cpp
heifner: Fixed in 65eca55d69. Verification now stays pending for a deterministic window of finalized blocks past the snapshot height (`snapshot_attestation_grace_blocks` = 12,500, half the 25,000-block provider snapshot interval) after the node reaches the live tip. The terminal skip-with-warning only happens once the node is caught up AND the finalized head is at least 12,500 blocks past the snapshot height, so a fresh bootstrap that reaches the tip before the providers' votesnaphash transactions land keeps retrying until the attestation can no longer be expected. This mirrors the grace window already added on PR #248 (ed68df0e9f) for auto-fetched snapshots; #248 will be synced on top of this so the window is common to both paths, with #248's auto-fetched case still escalating to FATAL after the window instead of warning.

contracts/sysio.system/src/snapshot_attest.cpp
heifner: Fixed in 65eca55d69. Votes now aggregate per (block_num, block_id, snapshot_hash) -- two votes with the same hash but different block ids accumulate in separate groups and cannot jointly reach quorum. The existing-record disagreement check also compares both the hash and the block id, so a vote matching the attested hash under a different block id is rejected with snap_hash_disagreement_error (9001). Added contract tests: blockid_mismatch_votes_not_aggregated (same hash under two block ids does not reach quorum; the record created by a proper second vote carries the right block id) and record_blockid_disagreement (post-attestation vote with same hash / different block id rejected with 9001).

docs/snapshot-attestation.md
```diff
@@ -0,0 +1,285 @@
...
+This auto-vote path is specific to `snapshot-provider-account` mode. Snapshots created
+through the producer API (`create_snapshot` / `schedule_snapshot`) are not auto-voted; a
+provider using that path submits `votesnaphash` itself (for example from a monitoring
+script).
+
+## 4. Snapshot verification on load
+
+When a node starts with `--snapshot`, `chain_plugin` records the loaded snapshot's block
+number and BLAKE3 root hash. Once the node syncs and the first irreversible block advances
+past the snapshot block, it runs a one-time check against the on-chain `snaprecords` table
```
huangminghuang: This still describes verification as a “one-time check” and says missing table/record immediately becomes “Warning only.” The implementation now retries on later finalized blocks and waits through the 12,500-block grace window before warning. The doc should be updated so operators understand the actual pending/retry behavior.

heifner: Fixed in c6f4f53da9. Section 4 now describes the per-finalized-block retry loop: the outcome table covers all pending/terminal cases (match, mismatch, still syncing, no table at catch-up, within grace window, grace exhausted, unexpected error), defines "caught up" (finalized timestamp within 30s of wall clock) and the 12,500-block grace window (half the 25,000-block provider interval), and notes the ABI is re-checked on each attempt so a system-contract upgrade that lands mid-sync is picked up.

huangminghuang (approved):

PR #248, opened 2026-03-13
Add snapshot_api_plugin and bootstrap from snapshot endpoint

heifner:
## Summary

- **`snapshot_api_plugin`**: New plugin with public read-only HTTP endpoints (`/v1/snapshot/latest`, `/v1/snapshot/by_block`, `/v1/snapshot/download`) for snapshot discovery and download. Adds `snapshot_ro` API category so operators can expose only snapshot serving publicly via `--http-category-address`. File serving uses Beast's `file_body` for zero-copy sendfile with Range header support for resumable downloads.

- **`--snapshot-endpoint` bootstrap**: New CLI option on `chain_plugin` to bootstrap a node directly from a remote snapshot provider. Downloads snapshot, verifies root hash via footer metadata, loads and syncs, then verifies on-chain attestation with a retry-based check (12,500 block timeout). Strict enforcement for auto-fetched snapshots (fatal on missing attestation); warnings only for manual `--snapshot`.

- **Supporting changes**: multiple snapshot finalized callbacks in `snapshot_scheduler`, `send_file_response`/`get_request_header`/`add_raw_handler` on http_plugin, `post_to_file` on `fc::http_client`, operator guide at `plugins/snapshot_api_plugin/README.md`

- **Integration test**: 9 tests covering API endpoints (metadata, download, Range header, catalog updates) and bootstrap flows (latest + specific block number)

## Key files

| Area | Files |
|------|-------|
| New plugin | `plugins/snapshot_api_plugin/` (plugin, README) |
| HTTP extensions | `beast_http_session.hpp`, `common.hpp`, `http_plugin.hpp/.cpp`, `abstract_conn_fwd.hpp` |
| Bootstrap | `chain_plugin.cpp` (`fetch_snapshot_from_endpoint`, `verify_snapshot_attestation`) |
| Download support | `fc/http_client.hpp/.cpp` (`post_to_file`) |
| Callback refactor | `snapshot_scheduler.hpp/.cpp`, `producer_plugin.hpp/.cpp` |
| Docs | `docs/snapshot-serving-plan.md`, `plugins/snapshot_api_plugin/README.md` |
| Test | `tests/snapshot_api_test.py` |


libraries/libfc/src/network/http/http_client.cpp
```diff
@@ -366,6 +369,74 @@ class http_client_impl {
...
+
+      auto conn_iter = get_connection(dest, deadline);
+      auto eraser = make_scoped_exit([this, &conn_iter](){
+         _connections.erase(conn_iter);
+      });
+
+      error_code ec = std::visit(write_request_visitor(this, req, deadline), conn_iter->second);
+      FC_ASSERT(!ec, "Failed to send POST request: {}", ec.message());
+
+      // Use string_body response — for very large files this buffers in memory.
```
huangminghuang: This uses http::response<http::string_body> for post_to_file(), then writes res.body() to disk at :427. Real snapshot files can be multi-GB, so --snapshot-endpoint can OOM or fail long before it reaches hash verification. This needs a streaming Beast parser/file writer path, not string_body.

libraries/chain/snapshot_scheduler.cpp
```diff
@@ -42,7 +42,19 @@ void snapshot_scheduler::on_irreversible_block(const signed_block_ptr& lib, cons
       auto next = pending->next;
 
       try {
-         next(pending->finalize(block_id, chain));
+         auto si = pending->finalize(block_id, chain);
+         next(si);
```
huangminghuang: This calls next(si), and the next lambda registered by execute_snapshot() invokes _snapshot_finalized_cbs at :165. Immediately after, on_irreversible_block() invokes the same callbacks again at :48. That makes provider mode submit duplicate votesnaphash transactions for every finalized scheduled snapshot and also double-notifies the catalog callback.

plugins/http_plugin/include/sysio/http_plugin/beast_http_session.hpp
```diff
@@ -461,6 +470,79 @@ class beast_http_session : public detail::abstract_conn,
...
+         if (range_start > range_end) {
+            error_results results{416, "Range Not Satisfiable"};
+            send_response(fc::json::to_string(results, fc::time_point::maximum()), 416);
+            return;
+         }
+
+         auto range_len = range_end - range_start + 1;
+         file_res_->result(http::status::partial_content);
+         // Seek to the start of the range
+         file_res_->body().seek(range_start, ec);
```
huangminghuang: This seeks to range_start and sets Content-Length to range_len, but the file_body still has the original file size and will read from the seek position to EOF. The current test only uses bytes=0-1023, which hides this. Resumed downloads like bytes=1048576-2097151 can over-send beyond the requested range and violate HTTP Range semantics.

huangminghuang (changes requested):

libraries/libfc/src/network/http/http_client.cpp
heifner: Fixed in ed68df0e9f. `post_to_file` now reads the response through a `response_parser<file_body>` that streams the body directly to the temp file, with the header read first so a non-OK status is reported from a small in-memory error body instead of being written to disk. No full-body buffering remains.

libraries/chain/snapshot_scheduler.cpp
heifner: Fixed in ed68df0e9f. Callback notification moved into a single `notify_snapshot_finalized()` invoked exactly once per finalized snapshot - from `create_snapshot()` in irreversible read mode or from `on_irreversible_block()` promotion - and removed from `execute_snapshot()`'s completion handler. Added regression tests in `tests/test_snapshot_scheduler.cpp` covering the scheduled, irreversible-mode, and direct-API paths with two registered callbacks.

plugins/http_plugin/include/sysio/http_plugin/beast_http_session.hpp
heifner: Fixed in ed68df0e9f. Replaced the seek-on-`file_body` approach with a `bounded_file_body` type whose writer serializes exactly the requested `[start, start+len)` window. `snapshot_api_test.py` now covers a mid-file range, verifies on a raw socket that the wire body is exactly the advertised length, and checks clamping when the range end is past EOF.

plugins/http_plugin/include/sysio/http_plugin/beast_http_session.hpp
```diff
@@ -461,6 +472,75 @@ class beast_http_session : public detail::abstract_conn,
...
+
+      file_res_->version(11);
+      file_res_->set(http::field::content_type, content_type);
+      file_res_->set(http::field::accept_ranges, "bytes");
+      file_res_->set(http::field::content_disposition, "attachment; filename=\"" + file_path.filename().string() + "\"");
+      if (plugin_state_->server_header.size())
+         file_res_->set(http::field::server, plugin_state_->server_header);
+      file_res_->keep_alive(false); // close after file transfer
+
+      auto tracked_size = byte_range ? (byte_range->second - byte_range->first + 1) : file_size;
```
huangminghuang: send_file_response() bypasses max-bytes-in-flight admission control. Normal JSON responses call verify_max_bytes_in_flight() before sending, but the new raw snapshot download path always increments and starts async_write. A public snapshot_ro endpoint can therefore accept multi-GB downloads even when the configured cap is already exceeded. Also, tracked_size is computed from the original, unclamped range, so Range: bytes=1-18446744073709551615 can inflate bytes_in_flight to an enormous value while only sending a small clamped response. Compute the final clamped response length first, call verify_max_bytes_in_flight(length), and return 503 before opening/writing if over the cap.

libraries/libfc/src/network/http/http_client.cpp
```diff
@@ -365,6 +388,94 @@ class http_client_impl {
...
+      // by the caller's deadline and by available disk space instead.
+      parser.body_limit(boost::none);
+
+      ec = std::visit([&](auto& stream) {
+         return sync_read_header_with_timeout(*stream, buffer, parser, deadline);
+      }, conn_iter->second);
+      FC_ASSERT(!ec, "Failed to read response header: {}", ec.message());
+
+      const auto status = parser.get().result();
+      if (status != http::status::ok && status != http::status::partial_content) {
```
huangminghuang: post_to_file() treats 206 Partial Content as a successful full snapshot download, but fetch_snapshot_from_endpoint() never sends a Range request. If a buggy/malicious endpoint or proxy returns partial content, we rename that partial body into the final bootstrap snapshot path and only fail later during footer/hash/load handling. For this bootstrap path, require 200 OK unless the client explicitly implements ranged/resumable downloads and validates the assembled file.

huangminghuang (changes requested):

plugins/http_plugin/include/sysio/http_plugin/beast_http_session.hpp
heifner: Fixed in 35f9bd25e6. `send_file_response()` now clamps the requested range against the file size first, computes the exact payload length that will be sent, and runs it through `verify_max_bytes_in_flight(payload_len)` before any response state is built -- over the cap it sends the same 503 busy response as the JSON path and returns without writing. bytes_in_flight is incremented with the clamped payload length, so `Range: bytes=1-18446744073709551615` can no longer inflate the in-flight accounting (that expression previously also overflowed uint64 to 0 for bytes=0-2^64-1). The file open does still happen before the check -- the authoritative size is needed to clamp the range -- but on rejection the handle is dropped immediately and nothing is sent. snapshot_api_test.py gains Test 10: with --http-max-bytes-in-flight-mb 1, a full-file download is rejected with 503, while a Range with an absurd raw end (bytes=N-18446744073709551615) on a small tail window still succeeds, proving admission control sees the clamped payload length.

libraries/libfc/src/network/http/http_client.cpp
heifner: Fixed in 35f9bd25e6. `post_to_file()` now requires exactly 200 OK. The client never sends Range requests, so a 206 can only mean a broken or hostile endpoint/proxy returning partial content, and it is treated as an error rather than renamed into the final snapshot path.

huangminghuang (approved):

PR #253, opened 2026-03-16
Add get_permission_lower_bound intrinsic and expandauth action

heifner:
## Summary
- New `get_permission_lower_bound` host function allowing contracts to read account permission authorities
- New `expandauth` system contract action that adds keys and/or account permissions to an existing authority
- 9 unit tests including ETH (EM) key signing verification

## Details

**Host function:** Lower-bound search on `permission_index<by_owner>`, serializes `(perm_name, parent, last_updated, authority)`. Registered across all runtimes (sys-vm, sys-vm-oc, native-module) and the genesis intrinsics whitelist.

**expandauth action:** Requires system contract auth. Reads the current authority via the new intrinsic, deduplicates and sorts added keys/accounts, then sends an inline `updateauth` using the account's own permission (privileged contract pattern).

Companion CDT PR: https://github.com/Wire-Network/wire-cdt/pull/37


jglanz (approved):
AWESOME

PR #255, opened 2026-03-16
Add precompiled header support for fc and sysio_chain

heifner:
## Summary
- Adds precompiled header (PCH) support for the two heaviest build targets (`fc` and `sysio_chain`), controlled via `-DENABLE_PCH=ON` (default ON)
- Includes all frequently used Boost headers (multiprecision, multi_index, asio, beast, iostreams, signals2, etc.), stable STL headers, and widely included fc/chain project headers
- Replaces `boost::range` usage with `std::ranges`/`std::find` in `authority_checker.hpp` and `parallel_markers.hpp` to resolve ADL ambiguity between `boost::range::find` and `std::ranges::find`

## Benchmark (clean build of fc + sysio_chain, 12-core, clang-18, Debug)

| | Wall time | CPU time |
|---|---|---|
| Without PCH | 2:54.55 | 3013.60s |
| With PCH | 1:36.06 | 1128.80s |
| **Speedup** | **45%** | **63%** |

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(GNUInstallDirs)
 include(MASSigning)
 include(AddTestHelpers)
+include(precompiled-headers)
 
 # Options
 option(ENABLE_WERROR "Enable `-Werror` compilation flag." Off)
--- a/cmake/include/sysio-pch.hpp
+++ b/cmake/include/sysio-pch.hpp
@@ -1,0 +1,168 @@
+// sysio-pch.hpp — precompiled header for fc and sysio_chain targets.
+//
+// This file is consumed by target_precompile_headers() in
+// cmake/precompiled-headers.cmake.  Both targets share a single source
+// file; headers that are only available in the chain target's include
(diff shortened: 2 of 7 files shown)


jglanz (changes requested):
I'd prefer if the content was under `cmake/include/sysio-pch.hpp`

I generally don't love PCH's due to staleness and traceability - but given the compilation improvement, as long as its clean, I wont argue too much

huangminghuang (changes requested):
Findings:

- **P1** [cmake/contract-tools.cmake:61](/Users/huangming/opensource/wire-sysio/cmake/contract-tools.cmake:61) / [cmake/contract-tools.cmake:91](/Users/huangming/opensource/wire-sysio/cmake/contract-tools.cmake:91)  
  Source/header globs freeze at parent configure time, but PR intentionally does not put them in `CMAKE_CONFIGURE_DEPENDS`. On branch switch that adds a new `.cpp/.hpp/.h`, file is absent from generated `DEPENDS`, so no reconfigure/clean trigger. On branch switch that removes one, generated build still depends on now-missing file, so Ninja/Make can fail before reaching reconfigure/clean. This leaves same class of stale branch-switch failures for add/remove file cases. Need directory-level sentinel, stamp regenerated by git/index state, or parent configure dependency on glob patterns/tracked file list.

heifner:
Just going to close this one for now.

PR #266, opened 2026-03-19
OPP > Part #1 > Round-trip envelope (WIRE, ETHEREUM, AND SOLANA) delivery works via batch operator with consensus (no slash/challenge yet)

jglanz:
## Summary

Introduces the OPP (Outpost Protocol) subsystem end-to-end: protobuf schema definitions, ZPP code-generation for CDT contracts, four new system contracts, two new plugins, and the build/tooling plumbing to tie it all together. Round-trip envelope delivery through the batch operator with on-chain consensus is functional; slashing and challenge mechanics are deferred to Part #2.

## OPP Protobuf Library (`libraries/opp/`)

- New protobuf schema covering core types (`ChainId`, `TokenAmount`, `ChainAddress`, `WireAccount`), attestation messages (`OppAttestation`, `OppEnvelope`, `BatchOperatorNextGroup`), and the top-level `OppMessage` wrapper.
- CMake integration via `opp_proto` (standard protobuf) and `opp_cdt_models` (ZPP zero-overhead headers for WASM contracts).
- `generate-opp-bundles.fish` tooling to produce Solidity and Solana protobuf bundles from the same `.proto` sources.

## New System Contracts

| Contract | Purpose |
|---|---|
| **sysio.epoch** | Epoch lifecycle management — tracks epoch index, group assignment, and drives epoch advancement with `BatchOperatorNextGroup` attestations. |
| **sysio.msgch** | Message channel — `deliver` stores inbound OPP data per outpost/epoch/operator with sha256 dedup; `evalcons` evaluates consensus (full agreement or majority at epoch boundary); `queueout`/`buildenv` prepare outbound envelopes with merkle roots. |
| **sysio.uwrit** | Underwriter — collateral tracking, fee configuration, greedy knapsack selection for optimal cross-chain swap underwriting. |
| **sysio.chalg** | Challenge — placeholder for future slashing/challenge logic. |

Each contract includes pre-built `.wasm`/`.abi` artifacts and Boost.Test unit tests (`contracts/tests/sysio.*_tests.cpp`).

### CDT Compatibility (`opp_table_types.hpp`)

Serialization (`<<`/`>>`) operators for all ZPP varint types and OPP protobuf structs, enabling seamless use in `multi_index` tables and contract action parameters.

### Batch Operator Plugin (major rework)
- Epoch state parsing with group assignment logic and state-change validation.
- Modularized polling/cranking — consensus evaluation moved from plugin to on-chain `sysio.msgch::evalcons`.
- Accurate ABI resolution and properly signed `push_action` transactions.
- Improved error handling so outbound delivery failures don't block inbound processing.

### Underwriter Plugin (new)
- Cross-chain swap validation for Ethereum and Solana via the existing outpost client plugins.
- Intent submission, collateral updates, and confirmation checks.

## Other Notable Changes

- **Chain plugin** — conditional BLS key auto-registration for producer nodes only; safer signature-provider queries.
- **ABI serializer** — improved enum-to-integer and string conversions for contract enum types.
- **Build system** — `linuxbrew` PATH exclusion in `build-init.cmake`; protobuf vcpkg dependency added; fixed path pollution in CMake.
- **`nodeop`** — registers `batch_operator_plugin` and `underwriter_plugin`.
- **Tooling** — `generate-system-contract-types.py` for TypeScript SDK type generation from contract ABIs.
- **Docs** — `docs/outpost-client-plugins.md`, entity diagrams in `etc/schema/`.

## Test Plan

- [ ] `ctest -j$(nproc) -LE _tests` passes (unit tests)
- [ ] Contracts build via `contracts_project` target
- [ ] Batch operator round-trip: inbound envelope delivery -> `evalcons` consensus -> outbound `buildenv`
- [ ] Verify pre-built `.wasm`/`.abi` match compiled output


.claude/wire-sysio-cicd-perf-session.txt
```diff
@@ -0,0 +1,765 @@
+ ▐▛███▜▌   Claude Code v2.1.80
```
heifner: I don't think we need this committed.

libraries/chain/authorization_manager.cpp
```diff
@@ -292,7 +292,7 @@ namespace sysio { namespace chain {
 
       // Prevents users from updating / adding protected 'ex.*' permissions.
       if(update.permission.prefix() == name("ex")) {
-         SYS_ASSERT( auth.actor == name("sysio"), invalid_permission, "Protected permission namespace. Only 'sysio' can update or add 'ex.*' permissions." );
+         SYS_ASSERT( auth.actor == name("sysio") || auth.actor == name("sysio.authex"), invalid_permission, "Protected permission namespace. Only 'sysio' can update or add 'ex.*' permissions." );
```
heifner, line 324: Should we make it `sysio.*` ?

CLAUDE.md
```diff
@@ -8,7 +8,7 @@ Wire Sysio is a C++ implementation of the AntelopeIO protocol (a fork of Spring)
 
 ## Build Commands
 
-**Note:** The build directory varies by developer (e.g. `cmake-build-debug`, `build`, `build/debug-claude`). Examples below use `$BUILD_DIR` — substitute your actual build path.
+**Note:** The build directory MUST be located under `<wire-sysio>/build/`, examples include `<wire-sysio>/build/claude`, `<wire-sysio>/build/debug-claude`, etc). Examples below use `$BUILD_DIR` — substitute your actual build path.
```
heifner, line 65: Would prefer this not dictate location.

libraries/chain/abi_serializer.cpp
```diff
@@ -713,21 +712,37 @@ namespace sysio::chain {
...
+            if( !found ) {
+               for( const auto& ev : e_itr->second.values ) {
+                  auto pos = ev.name.rfind('_');
+                  if( pos != std::string::npos && ev.name.substr(pos + 1) == name_str ) {
+                     val_to_pack = fc::variant(ev.value);
+                     found = true;
+                     break;
+                  }
+               }
+            }
```
heifner, line 745: Need test coverage for the enum support.

heifner (changes requested):

heifner (commented):
## Review

This PR is still WIP. @jglanz will address most/all of these in part 2/3/etc. They are not blockers for merge.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -0,0 +1,291 @@
...
+   // (deferred — needs per-AttestationEntry parsing first)
+}
+
+// ---------------------------------------------------------------------------
+//  queueout — queue outbound attestation for an outpost
+// ---------------------------------------------------------------------------
+void msgch::queueout(uint64_t outpost_id,
+                     opp::types::AttestationType attest_type,
+                     std::vector<char> data) {
+   auto now_sec = static_cast<uint64_t>(current_time_point().sec_since_epoch());
```
heifner, line 299: **Security: Missing `require_auth` — arbitrary attestation injection**

**Severity:** HIGH

`queueout` has no `require_auth()` call. Any account on the chain can call this action to inject arbitrary outbound attestations with `ATTESTATION_STATUS_READY` into the attestations table.

Contrast with other actions in this contract that _do_ check auth:
- `evalcons` → `require_auth(get_self())`
- `buildenv` → `require_auth(EPOCH_ACCOUNT)`
- `deliver` → `is_batch_operator_active(batch_op_name)`

The README (line 31) also documents that `queueout` should require auth from `sysio.msgch`.

**Exploit scenario:** An attacker calls `queueout` with a crafted `ATTESTATION_TYPE_BATCH_OPERATOR_NEXT_GROUP` payload containing attacker-controlled operator keys. When `buildenv` runs next, these fraudulent attestations are packed into the outbound OPP envelope and delivered to external chains. Forged `ATTESTATION_TYPE_SWAP` attestations could similarly fabricate cross-chain transfer instructions.

**Fix:** Add `require_auth(get_self());` as the first line. Since the intended caller is `sysio.epoch::advance()` via inline action, this ensures only contracts holding `sysio.msgch` authority can invoke it.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -0,0 +1,291 @@
...
+      e.raw_envelope  = packed;
+   });
+}
+
+// ---------------------------------------------------------------------------
+//  cleanup — remove old attestations and envelopes
+// ---------------------------------------------------------------------------
+void msgch::cleanup(uint32_t before_epoch) {
+   attestations_t atts(get_self(), get_self().value);
+   auto epoch_idx = atts.get_index<"byepoch"_n>();
```
heifner, line 398: **Security: Missing `require_auth` — arbitrary data deletion**

**Severity:** MEDIUM

`cleanup` has no `require_auth()` call and its `before_epoch` parameter is unconstrained. The epoch retention guard exists only in the _caller_ (`sysio.epoch::advance()` computes `before_epoch = current - retention_count`), not in `cleanup` itself.

**Exploit scenario:** An attacker calls `cleanup(UINT32_MAX)` repeatedly. Each call erases up to 50 attestation rows and 50 envelope rows. This destroys `ATTESTATION_STATUS_READY` entries before `buildenv` packs them, and deletes inbound envelope records needed for `evalcons` consensus evaluation — effectively halting the OPP cross-chain message channel.

**Fix:** Add `require_auth(EPOCH_ACCOUNT);` (matching `buildenv`'s pattern), or `require_auth(get_self());` if called via inline action.

jglanz: intentional - still in dev

libraries/chain/abi_serializer.cpp
jglanz: I will add it in part2, the `contracts_unit_test` uses enum abi(s) everywhere so it is covered. but i agree big picture

libraries/chain/authorization_manager.cpp
heifner, line 324: Question: `check_updateauth_authorization` now allows `sysio.authex` to create/update `ex.*` permissions, but `check_deleteauth_authorization` has no corresponding guard. Do we want users to be able to delete the `ex.*` permissions set up by `sysio.authex`? If not, the same prefix check should be added to the delete path.

Also minor: the assertion message still says "Only 'sysio' can update or add 'ex.*' permissions" but `sysio.authex` is now allowed too.

CLAUDE.md
jglanz: Claude makes assumptions otherwise.  I can make it specific to my host, but this is a fairly standard approach

libraries/chain/authorization_manager.cpp
jglanz: I'd prefer not to use `.*` - this should be very specific

.claude/wire-sysio-cicd-perf-session.txt
jglanz: removing

libraries/chain/abi_serializer.cpp
```diff
@@ -713,21 +712,37 @@ namespace sysio::chain {
...
+            // Try matching "name_str" as a suffix of each member name after a '_' separator.
+            if( !found ) {
+               for( const auto& ev : e_itr->second.values ) {
+                  auto pos = ev.name.rfind('_');
+                  if( pos != std::string::npos && ev.name.substr(pos + 1) == name_str ) {
+                     val_to_pack = fc::variant(ev.value);
+                     found = true;
+                     break;
+                  }
+               }
```
heifner, line 744: The prefix-stripped matching here is ambiguous: if two enum members share the same suffix after the last `_` (e.g., `chain_kind_ethereum` and `token_kind_ethereum`), the first one in ABI declaration order wins. Reordering members in a future ABI update would silently change the match result.

Also, the comparison is case-sensitive. Protobuf enum names are typically `UPPER_CASE` (`CHAIN_KIND_ETHEREUM`), so `rfind('_')` yields suffix `"ETHEREUM"`, but callers may pass `"ethereum"`. The match would silently fail.

Consider either requiring fully-qualified enum names (removing Pass 2), or building an explicit suffix map at ABI load time that rejects ambiguous suffixes.

heifner:
**Pre-existing issue in `ethereum_abi.cpp` worth fixing while you're in this file:**

`decode_static_value` (line ~437) takes `data` and `offset` but not `data_size`. It reads 32 bytes at `data + offset` with no bounds validation. When `decode_value` dispatches to it (line ~659), the `data_size` parameter is dropped. On malformed ABI data from an external chain, this is an out-of-bounds read.

Fix: Add `size_t data_size` parameter to `decode_static_value` and check `FC_ASSERT(offset + 32 <= data_size)`.

contracts/sysio.msgch/src/sysio.msgch.cpp
jglanz: same as above

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
```diff
@@ -1,22 +1,776 @@
...
+      std::promise<void> done;
+      auto future = done.get_future();
+
+      rw.push_transaction(
+         packed_var.get_object(),
+         [&done, &contract, &action_name](const auto& result) {
+            if (auto* err = std::get_if<fc::exception_ptr>(&result)) {
+               elog("batch_operator: push {}::{} failed — {}", contract, action_name, (*err)->to_string());
+            } else {
+               ilog("batch_operator: pushed {}::{} ok", contract, action_name);
```
heifner, line 566: **Bug: dangling references + promise use-after-free on timeout**

The lambda captures `contract`, `action_name`, and `done` by reference, but `push_transaction` is async. If `future.wait_for` times out (line 683), this function returns and destroys the stack locals. When the callback eventually fires, it accesses freed memory (`contract`, `action_name` for the log message, `done.set_value()` on a destroyed promise).

Same issue exists in `underwriter_plugin.cpp` at the equivalent `push_action`.

Fix: `auto done = std::make_shared<std::promise<void>>();` and capture `contract`, `action_name`, `done` by value.

libraries/chain/abi_serializer.cpp
jglanz: agreed - this is something i'll fix - but given enums are only used by the new code, I figured I'd hold off on a hard nose impl until everything is connected

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
jglanz: still WIP  - i've not gone and done a detailed review of these myself yet

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+   check(active_batch.size() >= cfg.batch_operator_minimum_active,
+         "not enough active batch operators for group assignment");
+
+   // Sort by registration epoch (stable, deterministic)
+   std::sort(active_batch.begin(), active_batch.end(), [&](const name& a, const name& b) {
+      auto ia = ops.find(a.value);
+      auto ib = ops.find(b.value);
+      return ia->registered_epoch < ib->registered_epoch;
+   });
```
heifner: Minor: `std::sort` is unstable, so operators with the same `registered_epoch` get an implementation-defined order. This is deterministic today since all nodes execute the same WASM binary, but recompiling with a different CDT version (different libc++) could silently change group assignment without any source code change.

Adding a tiebreaker (e.g., `return a.value < b.value`) would make this robust across recompiles.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -0,0 +1,257 @@
...
+   ledger.emplace(underwriter, [&](auto& e) {
+      e.id = ledger.available_primary_key();
+      e.underwriter = underwriter;
+      e.message_id = msg_id;
+      e.status = UnderwriteStatus::UNDERWRITE_STATUS_INTENT_SUBMITTED;
+      e.intent_time = now;
+      e.unlock_time = unlock;
+      e.source_sig = source_sig;
+      e.target_sig = target_sig;
+   });
```
heifner, line 68: `source_amount`, `target_amount`, `source_chain`, `target_chain`, and `fee_earned` are never set in this emplace -- they default to zero-amount assets with empty symbols. `expirelock` (line 119-128) later reads these fields to adjust collateral, but subtracting/adding zero is a no-op. Collateral is never actually locked or unlocked.

Either pass the amount/chain info into `submituw` and set them here, or document that `confirmuw` is responsible for populating these fields before collateral accounting matters.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -0,0 +1,291 @@
...
+   }
+
+   // Build OPP Envelope
+   opp::Envelope env;
+   env.epoch_index = zpp::bits::vuint32_t{epoch};
+   env.epoch_timestamp = zpp::bits::vuint64_t{now_sec};
+
+   // Serialize envelope
+   auto [packed, out] = zpp::bits::data_out<char>();
+   (void)out(env);
```
heifner, line 379: The `entries` vector is collected from READY attestations (lines 222-233), but only `epoch_index` and `epoch_timestamp` are set on the `Envelope` struct here. The `entries` are never written into the envelope payload -- `(void)out(env)` serializes an envelope with no attestation data. Outpost chains will receive an empty envelope.

The collected entries need to be assigned to the envelope (e.g., `env.attestations = entries;` or serialized alongside it) before packing.

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -0,0 +1,402 @@
...
+
+      // Serialize with zpp_bits (protobuf wire format)
+      auto [packed, out] = zpp::bits::data_out<char>();
+      (void)out(attest);
+
+      // Queue to sysio.msgch for each registered outpost
+      outposts_t outposts_tbl(get_self(), get_self().value);
+      for (auto it = outposts_tbl.begin(); it != outposts_tbl.end(); ++it) {
+         action(
+            permission_level{"sysio.epoch"_n, "owner"_n},
```
heifner, line 199: These inline actions should use `"active"_n` instead of `"owner"_n`. `owner` is the highest-privilege permission and shouldn't be used for routine inline operations. Same applies to lines 204 and 217.

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -0,0 +1,402 @@
...
+   check(old_it != ops.end(), "old operator not found");
+
+   auto new_it = ops.find(new_op.value);
+   check(new_it != ops.end(), "new operator not found");
+   check(new_it->type == OperatorType::OPERATOR_TYPE_BATCH, "replacement must be a batch operator");
+   check(new_it->status == OperatorStatus::OPERATOR_STATUS_ACTIVE, "replacement must be active");
+   check(!new_it->is_blacklisted, "replacement is blacklisted");
+
+   uint8_t group = old_it->assigned_batch_op_group;
+   check(group < 3, "old operator has no valid group assignment");
```
heifner: Hardcoded `group < 3` but `batch_op_groups` is configurable via `setconfig`. If reconfigured to more than 3 groups, operators in groups 3+ can't be replaced.

Use `cfg.batch_op_groups` or `state.batch_op_groups.size()` instead.

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -0,0 +1,402 @@
...
+         o.is_blacklisted = false;
+      });
+   } else {
+      check(!it->is_blacklisted, "operator is blacklisted");
+      ops.modify(it, same_payer, [&](auto& o) {
+         o.type = type;
+         o.status = OperatorStatus::OPERATOR_STATUS_WARMUP;
+         o.registered_epoch = state.current_epoch_index;
+      });
+   }
```
heifner: Re-registering an existing operator overwrites `type` and resets `status` to WARMUP, but doesn't check if the operator is currently in a `batch_op_group`. An ACTIVE operator assigned to a group could be changed from BATCH to UNDERWRITER while still appearing in `state.batch_op_groups`, breaking the group assignment invariant.

Consider disallowing type changes for operators with `assigned_batch_op_group != 255`, or removing them from their group first.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -0,0 +1,257 @@
...
+   for (auto c_it = uw_idx.lower_bound(it->underwriter.value);
+        c_it != uw_idx.end() && c_it->underwriter == it->underwriter; ++c_it) {
+      if (c_it->chain_kind == it->source_chain) {
+         collateral.modify(*c_it, same_payer, [&](auto& c) {
+            c.locked_amount -= it->source_amount;
+            c.available_amount += it->source_amount;
+         });
+      }
+      if (c_it->chain_kind == it->target_chain) {
+         collateral.modify(*c_it, same_payer, [&](auto& c) {
```
heifner, line 125: These are separate `if` blocks, not `else if`. If `source_chain == target_chain`, both branches execute and the amounts are subtracted/added twice.

Should be `else if`, or guard against the equal case.

heifner:
**Pre-existing issue in `ethereum_abi.hpp` — DCLP race in `EthDataTypeReflector::to_string`**

Around line 229, `data_type_name_map.contains(elem)` is read outside the mutex. If another thread concurrently inserts via `operator[]` (which can rebalance the `std::map` tree), this is a data race / UB.

Fix options: always lock before reading, use a pre-populated `static const` map (the enum is fixed at compile time), or use `std::shared_mutex` for read/write separation.

heifner:
**Pre-existing issue in `ethereum_abi.cpp` — `be_uint_from_decimal` breaks on negative signed integers**

`be_uint_from_decimal` (line ~57) asserts every character is a digit, rejecting negative numbers. But `encode_static_value` calls it for `int128`/`int256` types (line ~97-98). Encoding a negative signed integer (e.g., `-1` as `int256`) will hit the assertion. Ethereum ABI uses two's complement for signed types, which this function doesn't handle.

heifner:
**Pre-existing issue in `ethereum_abi.cpp` — `contract_encode_data` head-size calculation diverges from `encode_tuple`**

`contract_encode_data` (line ~876) computes head size as `heads.size() * 32`, assuming every head is exactly 32 bytes. But `encode_tuple` (line ~308) correctly sums actual head sizes, which handles static tuples that occupy more than 32 bytes (e.g., `(uint256,uint256)` has a 64-byte head). If a top-level input is a static tuple, `contract_encode_data` computes the wrong offset for subsequent dynamic parameters.

Fix: Use the same summing approach as `encode_tuple`.

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
```diff
@@ -1,22 +1,776 @@
...
+
+   // Plugin references
+   chain_plugin*                     chain_plug = nullptr;
+   cron_plugin*                      cron_plug  = nullptr;
+   outpost_ethereum_client_plugin*   eth_plug   = nullptr;
+   outpost_solana_client_plugin*     sol_plug   = nullptr;
+
+   // Cron job handle
+   cron_service::job_id_t            epoch_poll_job_id = 0;
+   bool                              shutting_down = false;
```
heifner, line 139: `shutting_down` is a plain `bool` written from the main thread (`plugin_shutdown`) and read from the cron thread pool. This is a data race.

Should be `std::atomic<bool>`. Same issue in `underwriter_plugin.cpp`.

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
```diff
@@ -1,22 +1,776 @@
...
+      ilog("batch_operator: querying ETH events (addr={}, abis={}, events={}, opp_addr={})",
+           opp_addr, opp_abis.size(), event_abi_count, eth_opp_addr);
+
+      // Query OPPMessage logs via ethereum_client
+      auto events = entry->client->get_events(
+         opp_addr,
+         {"OPPMessage"},
+         opp_abis,
+         eth::block_tag_t{std::string("0x0")},
+         eth::block_tag_t{std::string(eth::block_tag_latest)});
```
heifner: ETH events are queried from block 0 (`"0x0"`) to latest on every poll cycle. As the chain grows, this becomes increasingly expensive and will eventually cause timeouts or rate-limiting from the ETH RPC node.

Track the last processed block number and query only from there. Same issue in `underwriter_plugin.cpp` (`"earliest"` to latest).

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
```diff
@@ -1,22 +1,776 @@
...
+      p.limit = limit;
+      auto deadline = fc::time_point::now() + fc::milliseconds(delivery_timeout_ms);
+      auto result_fn = ro.get_table_rows(p, deadline);
+      auto result = result_fn();
+      if (auto* err = std::get_if<fc::exception_ptr>(&result)) {
+         elog("batch_operator: table read failed {}::{} — {}", code, table, (*err)->to_string());
+         return {};
+      }
+      return std::get<read_only::get_table_rows_result>(result);
+   }
```
heifner, line 316: Table reads are capped at `limit = 100` and the `more` flag on the result is never checked. If any table (outposts, envelopes, attestations, operators) grows beyond 100 rows, data is silently truncated.

Either paginate using `lower_bound` from the last key when `result.more == true`, or at minimum log a warning when rows are truncated. Same pattern in `underwriter_plugin.cpp`.

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
```diff
@@ -1,22 +1,776 @@
...
+
+   void do_poll_epoch_state() {
+      auto [ok, epoch_index] = parse_epoch_state();
+      if (!ok) return;
+
+      // ALL operators call advance regardless of election status.
+      // The contract is idempotent: returns silently if epoch hasn't elapsed.
+      // This ensures the chain can recover from downtime — any operator can
+      // crank the epoch forward, not just the elected group.
+      push_action("sysio.epoch", "advance", operator_account, fc::mutable_variant_object());
```
heifner: `advance` is pushed unconditionally every poll cycle (15s) by every operator, regardless of whether the epoch has elapsed. With 21 operators, that's ~84 no-op transactions/minute (each costs CPU for signing, ABI serialization, network propagation, and on-chain billing).

Consider reading epoch state first and only pushing `advance` if `next_epoch_start` has passed.

libraries/libfc/include/fc/variant.hpp
```diff
@@ -277,6 +277,25 @@ namespace fc
...
+        EnumType as_enum_value() const {
+           if (is_integer() || is_numeric())
+              return static_cast<EnumType>(as_int64());
+           if (is_string()) {
+              try { return static_cast<EnumType>(std::stoll(get_string())); }
+              catch (...) {}
+           }
+           throw std::runtime_error("Cannot convert variant to enum value");
+        }
+
```
heifner, line 298: All other `variant` methods throw FC exceptions (`fc::bad_cast_exception`, etc.). This throws `std::runtime_error`, which won't be caught by FC exception handlers and won't include the standard FC backtrace.

Use `FC_THROW("Cannot convert variant of type {} to enum value", get_type())` for consistency.

libraries/opp/include/sysio/opp/opp.hpp
```diff
@@ -0,0 +1,183 @@
+#pragma once
+#pragma GCC diagnostic ignored "-Wswitch"
```
heifner, line 2: This `#pragma GCC diagnostic ignored "-Wswitch"` is at file scope, so it suppresses warnings for every translation unit that includes this header -- not just OPP code. Could hide real bugs in consumer code.

Scope it with `push`/`pop`:
```cpp
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wswitch"
// ... FC_REFLECT_ENUM macros ...
#pragma GCC diagnostic pop
```

libraries/chain/authorization_manager.cpp
jglanz: kk - it's in my notes as discussed - openclaw == amazing

contracts/sysio.uwrit/src/sysio.uwrit.cpp
jglanz: lol - this is actually a leftover - sigs dont even exist anymore

libraries/libfc/src/network/ethereum/ethereum_client.cpp
```diff
@@ -556,6 +556,7 @@ std::vector<ethereum_event_data> ethereum_client::get_events(const address_compa
    filter("toBlock", to_block_tag(to_block));
    // topics[0] is an OR-array of event signature hashes
    filter("topics", fc::variants{topic_hashes});
+   ilog("Querying events for contract {} with filter: {}", to_hex(addr, true), fc::json::to_pretty_string(filter));
```
heifner, line 559: This `ilog` fires on every `get_events` call, logging the full JSON filter at INFO level. With batch operators polling every ~15 seconds, this creates significant log noise in production.

Should be `dlog` (debug level).

plugins/underwriter_plugin/src/underwriter_plugin.cpp
```diff
@@ -0,0 +1,736 @@
...
+            ("actor", auth_account.to_string())
+            ("permission", "active"))
+      });
+      action_obj("data", data);
+
+      auto& chain = chain_plug->chain();
+      auto head_id = chain.head().id();
+      auto ref_block_num = boost::endian::endian_reverse(head_id._hash[0]);
+      auto ref_block_prefix = head_id._hash[1];
+      auto expiration = chain.head().block_time() + fc::seconds(30);
```
heifner: Manual ref_block construction using `endian_reverse` and `_hash[]` indexing. The batch_operator plugin uses `trx.set_reference_block(head_id)` which does this correctly internally. Use the same approach here for consistency and to avoid divergence bugs.

heifner (approved):
Approved so not blocking. Issues can be addressed in part 2/3/etc

contracts/sysio.msgch/src/sysio.msgch.cpp
jglanz: This is weird because I verified the envelope was delivered?   I'll look into it

contracts/sysio.epoch/src/sysio.epoch.cpp
jglanz: intentional, groups is always 3 - but group sizes can vary - regardless it should not be a literal in the code

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
jglanz: 15s is a debugging value - in the realworld is ~60s but more importantly timed to the epoch, not wall-clock.  That said, I agree as you'll see shortly

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
jglanz: this batchop read_table on the host side plugin never touches messages, attestations, only state tables, but this does apply to the underwriter

plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
jglanz: OPPEnvelope is quickly pruned

PR #275, opened 2026-03-30
Harden snapshot deserialization against crafted snapshots

heifner:
## Summary

- Validates structural invariants in snapshot-loaded state that could crash or hang a node before on-chain attestation is verified
- Covers config validation (zero denominators, invalid WASM limits), singleton/count enforcement, finality core invariants, finalizer policy weights/thresholds, permission authority/cycles, and snapshot reader bounds checking
- All checks are O(1) or O(n) with negligible overhead relative to snapshot load time

## Motivation

Peer-served snapshots are hash-attested, but a node must process blocks from the snapshot state *before* it syncs far enough to verify the on-chain attestation record. A crafted snapshot with structurally valid but semantically poisoned state (e.g. zero denominators, circular permission parents, empty finality links) could crash or hang the node in this pre-attestation window.

Data consistency issues (code_hash mismatches, cross-object referential integrity) are covered by hash attestation and excluded from scope.

## Changes

**Config validation** — `chain_config::validate()`, `wasm_config::validate()` during GPO deserialization; elastic limit parameter max/max_multiplier/denominator checks; usage window and virtual limit positivity checks

**Count enforcement** — singleton assertions for GPO, DGPO, protocol_state, resource_limits config/state; block_summary == 65536; preactivated_protocol_features must be empty

**Finality state** — `finality_core::validate_snapshot()` checking all 9 documented invariants; finalizer policy threshold/weight validation with overflow protection on all policy types; null checks on active_finalizer_policy and active_proposer_policy; incremental_merkle_tree mask/trees consistency; valid_t.validation_mroots size

**Permissions** — authority threshold > 0, key type validation (k1/r1/wa/em/ed), inline circular parent detection via depth-bounded walk (capped at max_auth_depth=6)

**Snapshot reader** — integer overflow fix in section bounds check, num_sections cap at 256, duplicate section name detection, overlapping data range detection, JSON row array bounds check


libraries/chain/include/sysio/chain/database_utils.hpp
```diff
@@ -166,18 +166,20 @@ struct reader {
...
-            if (next == '\0') break;          // 0x00,0x00 = end
-            FC_ASSERT(next == '\x01', "Invalid NUL-escape sequence in BE key string");
+            if (next == '\0') { terminated = true; break; } // 0x00,0x00 = end
+            FC_ASSERT(next == '\x01', "Invalid NUL-escape sequence in BE key string: 0x00,0x{}", next);
             s.push_back('\0');                 // 0x00,0x01 = literal NUL
          } else {
             s.push_back(c);
          }
       }
+      FC_ASSERT(terminated, "Unterminated NUL-escaped string in BE key");
```
brianjohnson5972, line 182: Do we care if 0x00,0x00 is found, but `pos != end`? Seems like having throw away part of string should at least result in something logged.

libraries/chain/include/sysio/chain/finalizer_policy.hpp
```diff
@@ -40,6 +41,25 @@ namespace sysio::chain {
...
+         uint64_t sum = 0;
+         for (const auto& f : finalizers) {
+            SYS_ASSERT(f.weight > 0, snapshot_exception,
+                       "finalizer_policy generation {}: finalizer weight must be positive", generation);
+            SYS_ASSERT(sum <= std::numeric_limits<uint64_t>::max() - f.weight, snapshot_exception,
+                       "finalizer_policy generation {}: sum of weights overflows uint64_t", generation);
+            sum += f.weight;
+         }
+         SYS_ASSERT(threshold <= sum, snapshot_exception,
+                    "finalizer_policy generation {}: threshold ({}) exceeds sum of weights ({})",
```
brianjohnson5972: I think `generation {}: sum of weights ({}) didn't meet the required threshold ({})" would make more sense for a reported error.

libraries/chain/include/sysio/chain/kv_context.hpp
```diff
@@ -69,6 +69,7 @@ class kv_iterator_pool {
       s.is_primary = false;
       s.status = kv_it_stat::iterator_end;
       s.code = code;
+      s.key_format = config::kv_format_raw;
```
brianjohnson5972, line 72: I'm confused as to why allocate_primary (above) has the caller set the key_fmt, but allocate_secondary is always kv_format_raw? Are we guaranteed it will be set correctly later?

libraries/state_history/create_deltas.cpp
```diff
@@ -166,8 +166,8 @@ void pack_deltas(boost::iostreams::filtering_ostreambuf& obuf, const chainbase::
 
    // Count KV tables separately since kv_index may produce 2 tables
    // (contract_row for format=1, contract_row_kv for format=0)
-   auto is_standard_kv = [](const chain::kv_object& o) { return o.key_format == 1 && o.key.size() == chain::kv_key_size; };
-   auto is_raw_kv      = [](const chain::kv_object& o) { return o.key_format != 1 || o.key.size() != chain::kv_key_size; };
+   auto is_standard_kv = [](const chain::kv_object& o) { return o.key_format == chain::config::kv_format_standard && o.key.size() == chain::kv_key_size; };
+   auto is_raw_kv      = [](const chain::kv_object& o) { return o.key_format != chain::config::kv_format_standard || o.key.size() != chain::kv_key_size; };
```
brianjohnson5972, line 170: If this is NOT intending to also catch eventually added other kv formats, I think this should be `o.key_format != chain::config::kv_format_raw`

plugins/producer_plugin/src/trx_priority_db.cpp
```diff
@@ -97,7 +97,9 @@ void trx_priority_db::load_trx_priority_map(const controller& control, trx_prior
       const auto& kv_idx = db.get_index<chain::kv_index, chain::by_code_key>();
       auto itr = kv_idx.lower_bound(boost::make_tuple(config::system_account_name, config::kv_format_standard, prefix.to_string_view()));
 
-      while (itr != kv_idx.end() && itr->code == config::system_account_name) {
+      while (itr != kv_idx.end()
+             && itr->code == config::system_account_name
+             && itr->key_format == config::kv_format_standard) {
```
brianjohnson5972, line 102: Nit: Would prefer not to put logical &&/|| at the start, particularly as one off.

heifner: I actually prefer it to at the end 🤷

libraries/chain/include/sysio/chain/kv_context.hpp
heifner: Secondary index keys are always raw format — they store the secondary key value concatenated with the primary key, not BE-encoded structured keys like primary iterators. So `kv_format_raw` is the correct constant here, not a missing initialization.

libraries/state_history/create_deltas.cpp
heifner: This is changed in PR #288 where key_format is going away.

libraries/chain/include/sysio/chain/finalizer_policy.hpp
heifner: https://github.com/Wire-Network/wire-sysio/pull/275/commits/8044aa023bbbd569203ccc0308066a032ac15615

libraries/chain/include/sysio/chain/database_utils.hpp
heifner: https://github.com/Wire-Network/wire-sysio/pull/275/commits/8044aa023bbbd569203ccc0308066a032ac15615

brianjohnson5972 (approved):

PR #280, opened 2026-04-02
Add beacon chain update plugin and cranker executable

brianjohnson5972:
Summary                                                                                                                                                                            
                                                                                                                                                                                     
  - Introduces beacon_chain_update_plugin, a new nodeop plugin that periodically fetches Ethereum beacon chain state from beaconcha.in and pushes updates into on-chain smart        
  contracts.                                                                                                                                                                         
  - Adds cranker, a lightweight standalone executable that runs only the plugins needed for beacon chain updates (no full node required).                                            
  - Adds a cron expression parser to cron_plugin for human-readable schedule configuration.                                                                                          
  - Fixes big-integer type routing in libfc's JSON parser.                                                                                                                           
  - Adds a bounded-retry identify_block_for_transaction method to ethereum_client.                                                                                                   
                                                                                                                                                                                     
  ---                                                                                                                                                                                
  What the plugin does                                                                                                                                                               
                                                                                                                                                                                     
  beacon_chain_update_plugin runs on a configurable cron schedule and:
                                                                                                                                                                                     
  1. Fetches queue lengths from beaconcha.in/api/v2/ethereum/queues — both entry_queue and exit_queue — and calls DepositManager.setEntryQueue and WithdrawalQueue.setWithdrawDelay  
  with the estimated wait time in seconds.                                                                                                                                           
  2. Fetches current APY from beaconcha.in/api/v1/ethstore/latest and calls DepositManager.updateApyBPS with the value expressed in basis points.                                    
  3. Finalizes epochs on a separate configurable interval by calling OPP.finalizeEpoch.                                                                                              
                                                                                                                                                                                     
  All on-chain calls are submitted via the existing outpost_ethereum_client_plugin. Block confirmation is awaited using a new identify_block_for_transaction future.                 
                                                                                                                                                                                     
  Configuration options (all via nodeop config):                                                                                                                                     
  - beacon-chain-queue-url — queue endpoint (default: beaconcha.in)
  - beacon-chain-apy-url — APY endpoint (default: beaconcha.in)                                                                                                                      
  - beacon-chain-api-key — Bearer token for beaconcha.in API   
  - beacon-chain-contracts-addrs — JSON map of contract name → address                                                                                                               
  - beacon-chain-update-interval — cron schedule for queue/APY updates
  - beacon-chain-finalize-epoch-interval — cron schedule for epoch finalization                                                                                                      
   
  ---                                                                                                                                                                                
  Changes by area                                           
                                                                                                                                                                                     
  plugins/beacon_chain_update_plugin/ (new)                 
  - Plugin implementation with pimpl pattern; pure utility functions extracted to beacon_chain_update_detail.hpp for testability.                                                    
  - Contract client structs (OPP, deposit_manager, withdrawal_queue) wrapping ethereum_contract_client.                                                                              
  - HTTP fetchers using Boost.Beast over TLS for both the queue and APY endpoints.                                                                                                   
  - Loop capture bug fixed: cron job lambdas capture shared_ptr + string by value rather than dangling loop-local references.                                                        
  - Unsigned wrap guard in get_queue_length: returns nullopt if the API returns a past timestamp rather than wrapping to a huge value.                                               
                                                                                                                                                                                     
  plugins/beacon_chain_update_plugin/test/ (new)                                                                                                                                     
  - 15 Boost.Test cases covering get_field_from_object, get_queue_length (including missing fields, wrong types, past timestamps), and apy_fraction_to_bps (including floating-point 
  epsilon robustness).                                                                                                                                                               
                                                            
  programs/cranker/ (new)                                                                                                                                                            
  - Minimal main.cpp that boots signature_provider_manager_plugin, outpost_ethereum_client_plugin, cron_plugin, and beacon_chain_update_plugin — no chain, no p2p.
                                                                                                                                                                                     
  plugins/cron_plugin/ (new files)
  - cron_parser.hpp/cpp: parses standard 5-field and extended 6-field (milliseconds) cron expressions into cron_service::job_schedule objects, with full range/step/list syntax.     
                                                                                                                                                                                     
  libraries/libfc/src/network/ethereum/ethereum_client.cpp
  - identify_block_for_transaction(tx_hash): polls eth_getTransactionReceipt on a detached thread, returning a std::future<uint64_t> for the block number. Bounded to 600 attempts   
  (10 min); captures weak_from_this() to prevent use-after-free if the client is destroyed before the transaction mines.                                                             
  - Fixed ABI call encoding: 0x prefix added to encoded call data; to_hex called with with_prefix=true.                                                                              
                                                                                                                                                                                     
  libraries/libfc/src/io/json.cpp                                                                                                                                                    
  - Fixed big-integer type routing in number_from_stream. The min_len boundary values were all off by one (min_str.size() - 1 instead of min_str.size()), causing 18-digit negative
  numbers ≥ 922... to fall through to int128, 20-digit unsigned values to be routed to uint256 instead of uint128, and the int256/uint256 overflow checks to use inverted direction  
  comparisons. All four boundaries and their comparison operators are corrected.


libraries/libfc/src/io/json.cpp
```diff
@@ -293,14 +330,40 @@ namespace fc
...
+      const std::string& const_s = s;
+      const auto no_neg_start = neg ? 1 : 0;
+      const auto start = s.find_first_not_of('0', no_neg_start);
+      const auto str = (start != std::string::npos) ? std::string_view(const_s).substr(start) : std::string_view(const_s);
+
+      // if the string is empty and we dropped zeros
+      if (str.empty() && no_neg_start < start)
+        return 0;
+      // check for s== ".", "-","-.", since "[-]0*" is checked above
+      if (str == "." || str.empty()) // check the obviously wrong things we could have encountered
```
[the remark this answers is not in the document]
brianjohnson5972: comment is about original 's' parameter, str will have the '-' removed

libraries/libfc/src/network/ethereum/ethereum_client.cpp
```diff
@@ -84,7 +86,7 @@ fc::variant ethereum_client::execute(const std::string& method, const fc::varian
 fc::variant ethereum_client::execute_contract_view_fn(const address& contract_address, const abi::contract& abi,
                                                       const std::string& block_tag,
                                                       const contract_invoke_data_items& params) {
-   auto abi_call_encoded = contract_encode_data(abi, params);
+   auto abi_call_encoded = "0x" + contract_encode_data(abi, params);
```
brianjohnson5972: Here and 1 or 2 other places where contract_encode_data is called, the prefix is added. Wondering if we should add a default parameter to the method so that the underlying to_hex call can add it. Of course there will still be other places with "0x" manually added, so not sure it is worth it, unless we resolve not to do that.

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
```diff
@@ -0,0 +1,637 @@
...
+         return just_once_actions;
+      }
+
+      // This is used to make sure that there is a corresponding cron schedule associated with each collection of actions
+      if(schedules.count(interval_name) == 0) {
+         ilog("Could not find a schedule named {}, using {} interval", interval_name, default_interval_name);
+         interval_name = actual_default_schedule;
+      }
+
+      return intervals[interval_name];
```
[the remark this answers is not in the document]
brianjohnson5972: [8f20ec5](https://github.com/Wire-Network/wire-sysio/pull/280/changes/8f20ec5e08749e40c8624292f635dd4734577b43)

heifner (commented):
Code review of cranker/beacon_chain_update_plugin changes. 16 findings across 5 files.

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -355,6 +358,16 @@ class ethereum_client : public std::enable_shared_from_this<ethereum_client> {
...
 
+   /**
+    * @brief Receives a transaction hash that resolves to the block number once the transaction is included in a block.
+    * @param tx_hash The transaction hash
+    * @return A future<uint64_t> that resolves to the block number of the block
+    *         the transaction was included in. The future is fulfilled by a background
+    *         thread that polls eth_getTransactionReceipt until the receipt is available.
+    * @throws fc::network::json_rpc::json_rpc_exception if the initial RPC call fails.
+    */
+   std::future<uint64_t> identify_block_for_transaction(const std::string& tx_hash);
```
jglanz:
```
template <typename Fn, typename ... Args>
std::expected<decltype(Fn(Args...)), fc::exception> retry(retry_config_t config, Fn fn, Args&& ...) { ... }
```

jglanz (changes requested):
plus all the other stuff we just discussed

libraries/libfc/src/io/json.cpp
```diff
@@ -293,14 +330,40 @@ namespace fc
...
         FC_THROW_EXCEPTION(parse_error_exception, "Can't parse token \"{}\" as a JSON numeric constant", str);
       if( dot )
-        return parser_type == json::parse_type::legacy_parser_with_string_doubles ? variant(str) : variant(to_double(str));
-      if( neg )
-        return to_int64(str);
-      return to_uint64(str);
+        return parser_type == json::parse_type::legacy_parser_with_string_doubles ? variant(s) : variant(to_double(s));
+      if( neg ) {
+        if( str.size() < check_int128.min_len ||
```
qhool: What's the point of doing things this way, rather than converting to uint256 and then downconverting to a smaller type if possible (or even just sticking with uint256 here and doing the conversion in the calling code if needed rather than fiddling with variants). String <= is expensive enough that it's hard to imagine any huge performance gains (and I'm not seeing how performance of this stuff is particularly important at all, really)

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
```diff
@@ -0,0 +1,650 @@
...
+      asio::io_context    ioc;
+      asio::ssl::context  ssl_ctx{asio::ssl::context::tlsv12_client};
+      tcp::resolver       resolver{ioc};
+      auto                dest = resolver.resolve(host, port);
+
+      http::request<http::string_body> req{http::verb::post, path, 11};
+      req.set(http::field::host, host);
+      req.set(http::field::content_type, "application/json");
+      req.set(http::field::authorization, "Bearer " + api_key);
+      req.body() = R"({"chain":"mainnet"})";
```
qhool: I'm not loving the fact that "mainnet" is hard-coded here (or anywhere). It means this can't really be tested against or used for a testnet.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,650 @@
...
+      if(schedules.count(interval_name) == 0) {
+         ilog("Could not find a schedule named {}, using {} interval", interval_name, default_interval_name);
+         interval_name = actual_default_schedule;
+      }
+
+      return intervals[interval_name];
+   }
+
+   template <typename C>
+   std::pair<std::shared_ptr<C>, ethereum_client_ptr> get_contract(const outpost_ethereum_client_plugin& oec_plugin,
```
qhool, line 294: Are you planning to move this function? It is entirely generic AFAICT, and should be part of some shared client code.

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
```diff
@@ -0,0 +1,650 @@
...
+         try {
+            ilog("update Queue");
+            auto queues = get_queues_mainnet(my_.beacon_chain_queue_url, *(my_.beacon_chain_api_key));
+            ilog("queues: {}", fc::json::to_string(queues, fc::time_point::maximum()));
+            constexpr auto exit_queue = "exit_queue";
+
+            const auto exit_queue_len_sec = beacon_chain_detail::get_queue_length(queues, exit_queue);
+
+            constexpr auto nine_days = 9;
+            constexpr auto nine_days_in_sec = 60 * 60 * 24 * nine_days;
```
qhool: The length of the sweep does grow, slowly, over time, so a config value might be worthwhile.  Also, this is just weirdly named 
```
constexpr auto seconds_per_day = 60 * 60 * 24;
constexpr auto withdrawal_sweep_duration = 9 * seconds_per_day
```
Would be more clear

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
```diff
@@ -0,0 +1,650 @@
...
+   static constexpr auto contract_name = "OPP";
+
+   ethereum_contract_tx_fn<fc::variant> finalizeEpoch;
+   OPP(const ethereum_client_ptr& client,
+       const address_compat_type& contract_address_compat,
+       const std::vector<fc::network::ethereum::abi::contract>& contracts)
+      : ethereum_contract_client(client, contract_address_compat, contracts)
+      , finalizeEpoch(create_tx<fc::variant>(get_abi("finalizeEpoch"))) {
+
+      };
```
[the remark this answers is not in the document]
brianjohnson5972: Since I usually forget to add the semicolon after class/structs, until the compiler tells me, I think that is Claude's fault. Either way, I will fix it.

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
```diff
@@ -0,0 +1,650 @@
...
+                    *deposit_queue_len_sec, seconds_per_day, depositQDaysFl);
+
+            ilog("Sending setEntryQueue({} days) transaction to {} contract using address {}",
+                 depositQDaysFl, deposit_manager::contract_name,
+                 fc::to_hex(eth_client->get_address(), true));
+
+            auto res2 = dm_contract->setEntryQueue(depositQDaysFl);
+            const auto tx_hash2 = res2.as_string();
+            ilog("setEntryQueue tx sent, hash: {}", tx_hash2);
+            auto bn2 = eth_client->identify_block_for_transaction(tx_hash2);
```
qhool: This and updateApyBPS below are independent and don't need to be mined in any particular order. You could push both and then call identify_block_for_transaction for each afterwards to cut down on the time that the contracts are in a half-updated state.

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
```diff
@@ -0,0 +1,637 @@
...
+      req.prepare_payload();
+
+      beast::ssl_stream<beast::tcp_stream> stream(ioc, ssl_ctx);
+      if (!SSL_set_tlsext_host_name(stream.native_handle(), host.c_str()))
+         throw beast::system_error(beast::error_code(static_cast<int>(::ERR_get_error()),
+                                                     asio::error::get_ssl_category()));
+
+      beast::get_lowest_layer(stream).connect(dest);
+      stream.handshake(asio::ssl::stream_base::client);
+      http::write(stream, req);
```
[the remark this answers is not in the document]
brianjohnson5972: Handled by Kevin's comment fix

libraries/libfc/src/network/ethereum/ethereum_abi.cpp
```diff
@@ -1072,4 +1077,19 @@ void fc::from_variant(const fc::variant& var, fc::network::ethereum::abi::contra
 
    parse_components(vo.inputs, "inputs");
    parse_components(vo.outputs, "outputs");
+   bool missed = true;
+   if(deferred_name) {
```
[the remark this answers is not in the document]
brianjohnson5972: changed to throwing exception to identify what we may be missing.

libraries/libfc/src/network/ethereum/ethereum_client.cpp
```diff
@@ -476,6 +478,52 @@ std::string ethereum_client::send_raw_transaction(const std::string& raw_tx_data
...
+ *
+ * @param raw_tx_data The signed, RLP-encoded transaction data (hex string with "0x" prefix)
+ * @return A pair of the transaction hash string and a future<uint64_t> that resolves to the
+ *         block number once the transaction is mined
+ */
+std::future<uint64_t> ethereum_client::identify_block_for_transaction(const std::string& tx_hash) {
+   std::promise<uint64_t> promise;
+   std::future<uint64_t> future = promise.get_future();
+
+   constexpr int max_retries = 600; // 10 minutes at 1s/poll
```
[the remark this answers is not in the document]
brianjohnson5972: covered with Jonathan's suggestion

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
brianjohnson5972: missed that in there, fixing

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -23,29 +26,22 @@ using namespace fc::network::json_rpc;
  * Can hold either a string (for block numbers) or string_view (for tags like "latest", "pending")
  */
 using block_tag_t = std::variant<std::string, std::string_view>;
+class block_tag {
+public:
+   enum labeled { latest, pending, earliest, not_valid };
+   block_tag(labeled name);
+   block_tag(uint64_t bn);
```
heifner: nit: `explicit` on both

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -491,10 +494,15 @@ class ethereum_client : public std::enable_shared_from_this<ethereum_client> {
     */
    std::mutex _contracts_map_mutex{};
 
+   /**
+    * @brief Cached nonce for _signature_provider
+    */
+   uint256 _nonce  GUARDED_BY(_contracts_map_mutex) {0u};
```
heifner: I think these `GUARDED_BY` only work with `fc::mutex`.

libraries/libfc/src/io/json.cpp
```diff
@@ -28,6 +30,28 @@ namespace fc
...
+   big_int_as_str<int64_t> check_int64;
+
+   template<>
+   struct big_int_as_str<uint64_t> {
+      static constexpr std::string_view max_str = "18446744073709551615";
+      static constexpr auto max_len = max_str.size();
+   };
+   big_int_as_str<uint64_t> check_uint64;
+}
+
```
heifner, line 45: Could we lean on std::numeric_limits here? Something like:

```cpp
  template<std::integral I>
  constexpr std::size_t max_decimal_len = []{
     std::size_t n = std::is_signed_v<I> ? 1 : 0;
     auto v = std::numeric_limits<I>::max();
     do { ++n; v /= 10; } while (v != 0);
     return n;
  }();
```
A few small wins: works for any integer type (no new specializations), can't drift from the real max, and the sign slot is included so callers don't have to remember +1 for -.

plugins/cron_plugin/src/services/cron_parser.cpp
```diff
@@ -0,0 +1,242 @@
...
+using exact_value = cron_service::job_schedule::exact_value;
+using step_value = cron_service::job_schedule::step_value;
+using range_value = cron_service::job_schedule::range_value;
+
+// Trim whitespace from both ends
+std::string_view trim(std::string_view s) {
+   auto start = s.find_first_not_of(" \t\r\n");
+   if (start == std::string_view::npos) return "";
+   auto end = s.find_last_not_of(" \t\r\n");
+   return s.substr(start, end - start + 1);
```
heifner: Consider replacing this with 
`boost::algorithm::trim_copy_if(s, boost::is_any_of(" \t\r\n"))` from `<boost/algorithm/string/trim.hpp>`. It returns the same range type (so string_view in, string_view out — no allocation) and avoids rolling our own. Use `boost::is_space()` instead if you want locale-aware whitespace.

plugins/cron_plugin/src/services/cron_parser.cpp
```diff
@@ -0,0 +1,242 @@
...
+   size_t end = s.find(delim);
+
+   while (end != std::string_view::npos) {
+      result.push_back(s.substr(start, end - start));
+      start = end + 1;
+      end = s.find(delim, start);
+   }
+   result.push_back(s.substr(start));
+   return result;
+}
```
heifner, line 32: boost::algorithm::split from <boost/algorithm/string/split.hpp> handles this directly and preserves string_view semantics (no allocation per piece). The predicate form also makes it trivial to extend to multi-char delimiters later via boost::is_any_of(",;|").

Note: matches current behavior including the trailing empty piece on a trailing delimiter (e.g. "a,b," → ["a","b",""]). If you'd rather drop empties, pass boost::token_compress_on as the 4th arg.

std alternative (C++20): std::views::split(s, delim) is lazy and allocation-free, but materializing into vector<string_view> requires a transform + ranges::to (C++23) or a manual loop — not a clean one-liner yet, so Boost is the cleaner pick here.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
brianjohnson5972: I moved most of it out, but would like to leave the rest till I have another script pulled into WIRE so that reuse APIs will hopefully be clearer

heifner (changes requested):

libraries/libfc/src/io/json.cpp
brianjohnson5972: Changed it to just use a 256 int if too big for 64 int, and throws if 256 would be exceeded. Lots of unit tests fail if we just used a 256 int and skipped 64 int altogether.

plugins/beacon_chain_update_plugin/src/beacon_chain_update_plugin.cpp
brianjohnson5972: With the change to cache the _nonce value for ethereum_client, any delaying is just left till the reporting at the end of the script, which likely would just be for the last transaction that we are getting the block number for.

libraries/libfc/src/io/json.cpp
brianjohnson5972: we need the string as well, and the unit tests protect against '+1'

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -12,40 +13,28 @@
...
+   explicit block_tag(labeled name);
+   explicit block_tag(uint64_t bn);
+   std::string to_string() const;
 
-/**
- * @brief Block tag constant representing pending transactions
- */
-constexpr std::string_view block_tag_pending = "pending";
+   labeled  kind;
+   uint64_t number;
```
heifner: Nit: if you never have both, could be a `std::variant`

libraries/libfc/src/io/json.cpp
```diff
@@ -350,14 +363,56 @@ namespace fc
...
+      }
+
+      // using the string with no leading 0s, to avoid the string being assumed to be in octal,
+      // since a leading 0 with only digits between 0 and 7 are assumed to be octal
+      fc::uint256 val256(str);
+      if( val256 <= std::numeric_limits<uint64_t>::max() ) {
+         return static_cast<uint64_t>(val256);
+      }
+
+      return val256;
```
heifner: All the returns should return a `fc::variant` for NVRO. This was pre-existing, but might as well improve it.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,613 @@
...
+               const auto sec_sleep_str = field.value();
+               auto [ptr, ec] = std::from_chars(sec_sleep_str.data(), sec_sleep_str.data() + sec_sleep_str.size(), sec_sleep);
+               if (ec == std::errc() && ptr == sec_sleep_str.data() + sec_sleep_str.size()) {
+                  // identified a valid reason to retry
+                  valid = true;
+                  std::this_thread::sleep_for(std::chrono::milliseconds(sec_sleep * 1000));
+                  break;
+               }
+            }
+         }
```
heifner: `field.name_string()` in Beast returns the header name *without* a trailing colon, so `== "Retry-After:"` never matches. The whole block is dead: the Retry-After sleep never fires, `valid = true` never runs, and the `retry > 0` cap means the second iteration returns `{}` unconditionally regardless of what the server sent.

Suggestion replaces the loop with a typed lookup; a few other cleanups rolled in:
- `http::field::retry_after` is the typed enum — can't rot if the spelling changes.
- `std::chrono::seconds(sec_sleep)` is clearer than `milliseconds(sec_sleep * 1000)` and has no overflow path.
- Inner `auto [ptr, ec]` shadowed the outer `beast::error_code ec` — renamed to `perr`.

```suggestion
         const auto retry_after = res.base()[http::field::retry_after];
         if (!retry_after.empty()) {
            uint64_t sec_sleep = 0;
            auto [ptr, perr] = std::from_chars(retry_after.data(),
                                               retry_after.data() + retry_after.size(), sec_sleep);
            if (perr == std::errc() && ptr == retry_after.data() + retry_after.size()) {
               std::this_thread::sleep_for(std::chrono::seconds(sec_sleep));
            }
         }
```

After accepting, line 165's `uint64_t sec_sleep = 0;` becomes dead — safe to delete in a follow-up commit.

plugins/cron_plugin/include/sysio/services/cron_service.hpp
```diff
@@ -187,10 +194,86 @@ class cron_service {
...
+               signal_done();
+         } catch (const fc::exception& e) {
+            error = e.dynamic_copy_exception();
+            signal_done();
+         }
+      };
+
+      auto scheduled_id = this->add(opts.retry_schedule, retry_fn);
+      done_future.wait();
+      this->cancel(scheduled_id);
```
heifner: `retry_fn` captures `ret` by reference, then gets handed to the cron service and invoked on worker threads. That opens three failure modes.

**1. Concurrent writes to `ret` (data race).** `this->add(opts.retry_schedule, retry_fn)` passes no metadata, so the default is `one_at_a_time = false`. The scheduler fires on every tick regardless of whether a previous worker is still running. If `fn` takes longer than `retry_schedule` (e.g. a 40ms HTTP call with a 25ms retry schedule), two cron workers execute `retry_fn` concurrently — both doing `ret = fn(...)` on the same `std::expected`. Two threads writing to one object without synchronization is UB. Tests don't hit this today because their `fn` lambdas return instantly, so workers never overlap.

**2. Lost success after `signal_done`.** Even if execution is serialized:

- Worker A: `ret = fn(...)` succeeds → `signal_done()` (promise fires).
- Main thread returns from `done_future.wait()` and calls `cancel(scheduled_id)`.
- Worker B was already dispatched on the next tick: `ret = fn(...)` empty → clobbers A's success.
- Main thread reads `ret.has_value()` → false → returns `on_exhaustion()`.

The success value isn't pinned — `ret` is a mutable slot that everyone writes into, and the read happens separately from the signal. The once_flag gates only the promise, not the result.

**3. Stack-lifetime if `cancel` is async.** `cron_service::cancel(id)` appears non-blocking — it marks the job cancelled but doesn't wait for in-flight executions to return. After `done_future.wait()` + `cancel()` + the `if (ret.has_value())` check + return, the main thread unwinds and `ret` / `error` / `done_promise` are destroyed. Any worker still mid-`retry_fn` then reads/writes dead stack memory — use-after-scope.

**Suggested structure:**

```cpp
using ResultT = typename std::invoke_result_t<Fn, Args...>::value_type;
std::promise<std::expected<ResultT, fc::exception>> promise;
auto future = promise.get_future();
std::once_flag fired;

auto retry_fn = [&, attempt = 0]() mutable {
   try {
      auto r = fn(std::forward<Args>(args)...);   // local, per-invocation
      if (r.has_value())
         std::call_once(fired, [&]{ promise.set_value(std::move(*r)); });
      else if (++attempt >= opts.max_retries)
         std::call_once(fired, [&]{ promise.set_value(std::unexpected(opts.on_exhaustion())); });
   } catch (const fc::exception& e) {
      auto err = e.dynamic_copy_exception();
      std::call_once(fired, [&, err]{ promise.set_value(std::unexpected(std::move(*err))); });
   }
};

auto id = this->add(opts.retry_schedule, retry_fn,
                    job_metadata_t{.one_at_a_time = true});
auto result = future.get();
this->cancel(id);
return result;
```

- `r` is local to each invocation — no shared mutable slot.
- The result goes into the promise under `call_once`; `get`/`set_value` synchronizes the read.
- `one_at_a_time = true` prevents accidental pileup if `fn` ever runs long.

Problem 3 (stack lifetime) isn't fully fixed by the above — it also needs either `cron_service::cancel` to block on in-flight invocations, or `blocking_retry` to keep the captured state alive until the scheduler confirms no worker is still holding it.

libraries/libfc/src/io/json.cpp
```diff
@@ -350,14 +363,56 @@ namespace fc
...
+
+      const auto no_neg_start = neg ? 1 : 0;
+      const auto start = s.find_first_not_of('0', no_neg_start);
+      const auto str = (start != std::string::npos)
+         ? std::string_view(s).substr(start)
+         : std::string_view{};
+
+      // if the string is empty and we dropped zeros
+      if (str.empty() && no_neg_start < start)
+         return 0u;
```
heifner: The leading-zero shortcut fires on a bare `-` also.

Trace for input `-` followed by a non-digit (e.g. `[-,1]` or `{"x":-}`):
- `s = "-"`, `neg = true`, loop exits on non-digit.
- `no_neg_start = 1`, `find_first_not_of('0', 1)` on a length-1 string returns `npos` (starting-pos past end).
- `str = ""` (empty), then: `str.empty()` true && `1 < size_t_max` true → `return 0u`.

So malformed JSON like `[-,1]` silently parses as `[0,1]` instead of throwing `parse_error_exception`. The pre-PR code caught `"-"` explicitly; this refactor kept the `"."` / empty throw but dropped `"-"`. Regression.

The distinguishing signal is "did we actually strip any digits": `s.size() > no_neg_start` is true iff there were characters after the sign (which must all be `'0'`s if `str` ended up empty). For bare `"-"`, `s.size() == 1 == no_neg_start` and the shortcut doesn't fire — falls into the `str.empty()` throw below.

```suggestion
      // if the string is empty and we had actual digits after the sign
      if (str.empty() && s.size() > no_neg_start)
         return 0u;
```

Worth adding regression tests in `libraries/libfc/test/io/test_json_variant.cpp` too:

```cpp
BOOST_AUTO_TEST_CASE(number_from_stream_bare_minus_throws) {
   BOOST_CHECK_THROW(json::from_string("-"), fc::parse_error_exception);
}
BOOST_AUTO_TEST_CASE(number_from_stream_bare_minus_in_array_throws) {
   BOOST_CHECK_THROW(json::from_string("[-,1]"), fc::parse_error_exception);
}
```

libraries/libfc/src/network/ethereum/ethereum_client.cpp
```diff
@@ -135,15 +160,27 @@ fc::variant ethereum_client::execute_contract_tx_fn(const eip1559_tx& source_tx,
...
+fc::uint256 ethereum_client::get_signer_nonce() {
+   const auto count = raw_get_transaction_count(get_signer_address(), block_tag_latest);
+   fc::lock_guard lock(_contracts_map_mutex);
+   if (_nonce < count) {
+      _nonce = count;
+   } else {
+      ++_nonce;
+   }
+   return _nonce;
+}
```
heifner: Silent behavior change: the nonce source went from `pending` to `latest` (plus caching).

Previously, `create_default_tx` called `get_transaction_count(get_signer_address(), "pending")` — always asks the node what the next usable nonce is, including txs sitting in the mempool.

Now `raw_get_transaction_count(..., block_tag_latest)` returns only the count of *mined* txs, and the cached `_nonce` tracks what we've broadcast since.

**Failure mode — cold restart with an outstanding mempool tx:**

Causes: a prior instance crashed mid-broadcast, a low-gas tx is stuck pending, a second signer is active on the same address, or the RPC dropped a reorg. Sequence:

- `_nonce = 0` at construction.
- First call: `raw_get_transaction_count(latest)` returns `N` (the last mined count, which does NOT include the stuck pending tx at nonce `N`).
- `_nonce` seeded to `N`, returns `N`.
- We broadcast a tx with nonce `N` → collides with the stuck one. Depending on gas, either the old tx gets replaced-by-fee, or the new one gets rejected, or the node happily keeps both and only one ever mines.

The pre-PR `pending` form would have returned `N+1` and avoided the collision.

The *intent* behind caching (commit `890746866d` — "ensure the client uses the most recent next number, rather than relying on that a transaction is in a block already") is right for the warm case: we don't want to wait for one tx to mine before sending the next. But seeding from `latest` inherits the cold-start collision risk without the `pending` form's self-correction.

**Suggestion:** seed the cache from `pending` (optionally `max(pending, latest)` for paranoia), then keep the cache-increment logic for warm operation:

```suggestion
fc::uint256 ethereum_client::get_signer_nonce() {
   const auto on_chain = raw_get_transaction_count(get_signer_address(), block_tag_pending);
   fc::lock_guard lock(_contracts_map_mutex);
   if (_nonce < on_chain) {
      _nonce = on_chain;
   } else {
      ++_nonce;
   }
   return _nonce;
}
```

Separately, reusing `_contracts_map_mutex` for the nonce is semantically unrelated.  It makes contract-cache reads serialize with nonce reads for no reason. I think that is fine, but a comment would be good. You could consider a dedicated `std::mutex _nonce_mutex`, but contention is minimal so probably better as is.

libraries/libfc/src/io/json.cpp
heifner: JSON number parsing now routes every numeric token through `fc::int256` / `fc::uint256` (backed by `boost::multiprecision::int256_t`) before downcasting to 64-bit. The pre-PR path went directly to 64-bit via `to_int64` / `to_uint64`.

Every ABI JSON, SHiP record, RPC response, trace_api payload, and config file hits `number_from_stream`. The vast majority of tokens are small ints that fit in 64-bit. Constructing a 256-bit big-int from a decimal string, multiplying by -1, range-comparing, and downcasting is substantially more work than the direct 64-bit path that was there before.

Suggest fast-pathing the 64-bit case using the same length-compare idiom already used for the 256-bit limits, and only falling through to the big-int path when the 64-bit fast path would overflow. Sketch for the negative branch:

```cpp
constexpr std::string_view int64_min_magnitude_str = "9223372036854775808"; // |INT64_MIN|

if (neg) {
   // Common case: fits in int64
   if (str.length() < int64_min_magnitude_str.length() ||
       (str.length() == int64_min_magnitude_str.length() && str <= int64_min_magnitude_str)) {
      uint64_t mag = 0;
      std::from_chars(str.data(), str.data() + str.size(), mag);
      // magnitude fits in uint64; negate via wider arithmetic to handle INT64_MIN correctly
      return static_cast<int64_t>(-static_cast<int64_t>(mag));
   }

   // int256 fallback for magnitudes beyond int64 range
   // ... existing int256 path here
}
```

Same structure for the positive branch: compare against `uint64_max_str` first, parse directly into `uint64_t` if it fits, fall through to `uint256` only when needed.

Preserves all the new semantics (leading-zero handling, int256 support, range validation) but keeps the common path on the fast lane.

libraries/libfc/src/io/json.cpp
```diff
@@ -350,14 +363,56 @@ namespace fc
...
+
+         // using the string with no leading 0s, to avoid the string being assumed to be in octal,
+         // since a leading 0 with only digits between 0 and 7 are assumed to be octal
+         fc::int256 val256(str);
+         val256 *= -1;
+         if( val256 >= std::numeric_limits<int64_t>::min() ) {
+            return static_cast<int64_t>(val256);
+         }
+
+         return val256;
```
heifner: `fc::int256` is `boost::multiprecision::int256_t` — range `[-2^255, 2^255 - 1]`. `int256_max_str` is `|INT256_MIN|` = 2^255, which is 1 larger than `INT256_MAX`.

For the edge case where `str == int256_max_str`:

```cpp
fc::int256 val256(str);   // 2^255 -- out of range for signed int256
val256 *= -1;
```

`int256(2^255)` wraps via two's-complement to `INT256_MIN`. Then `INT256_MIN *= -1` would be 2^255, which wraps back to `INT256_MIN`. Double-wrap gives the correct result by coincidence, but it relies on boost.multiprecision's implementation-defined behavior for out-of-range fixed-width signed ints.

Cleaner: special-case the `|INT256_MIN|` magnitude first, then guarantee the `int256` construction is in-range. Also `val256 = -val256;` is clearer than `val256 *= -1;` and doesn't involve a multiply.

```suggestion
         // str is already validated to fit int256 magnitude.
         if (str == int256_max_str) {
            return std::numeric_limits<fc::int256>::min();
         }

         // using the string with no leading 0s, to avoid the string being assumed to be in octal,
         // since a leading 0 with only digits between 0 and 7 are assumed to be octal
         fc::int256 val256(str);
         val256 = -val256;
         if (val256 >= std::numeric_limits<int64_t>::min()) {
            return static_cast<int64_t>(val256);
         }

         return val256;
```

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,613 @@
...
+               action();
+            }
+            catch (const std::exception& e) {
+               elog("Error executing beacon chain update for the just once actions: {}", e.what());
+            }
+         }
+      },
+      cron_service::job_metadata_t{
+         .one_at_a_time = true, .tags = {"ethereum", "gas"}, .label = "beacon_chain_startup"
+      });
```
heifner, line 492: The just-once actions are scheduled with cron pattern `*/1 * * * *` and self-cancel on first fire. First fire happens at the top of the next minute — up to 60 seconds of latency between startup and execution. The README says:

> a `once` interval is always available -- it runs immediately on startup and does not repeat

The "immediately" part isn't accurate. In a crash-restart loop the 60s delay is cumulative with whatever restart latency nodeop/cranker has.

A few options:
- Extend `cron_service` with a `post_one_shot(fn)` (or `post(fn)`) that enqueues directly onto the thread pool, and use that here instead of a self-cancelling cron job. Probably the right shape longer-term.
- Schedule with a millisecond-granularity 6-field pattern (e.g. `0 * * * * *`, ms=0) if the scheduler respects sub-minute resolution.
- Minimum: update the README to read "runs at the next minute boundary after startup, then not again", so operators know what to expect.

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -393,12 +389,17 @@ class ethereum_client : public std::enable_shared_from_this<ethereum_client> {
...
    /**
-    * @brief Retrieves the transaction count (nonce) for an address.
+    * @brief Retrieves the raw transaction count (nonce) for an address via RPC.
+    *
+    * This is an uncached lookup. For obtaining the next nonce to use for this
+    * client's own signer, use the internal cached path via create_default_tx.
+    *
     * @param address The address for which to fetch the transaction count.
-    * @param block_tag
+    * @param block_tag Block tag at which to query.
```
heifner: Parameter is named `tag`, not `block_tag`. Doxygen mismatch.

```suggestion
    * @param tag Block tag at which to query.
```

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,613 @@
...
+         SYS_ASSERT(signed_epa >= 0, sysio::chain::plugin_config_exception,
+                    "queues[{}][{}] is negative: {}", queue_branch, epa_field, signed_epa);
+      }
+
+      const auto now_sec = fc::time_point::now().sec_since_epoch();
+      const auto epa = epa_var->as_uint64();
+      if (epa <= now_sec) {
+         wlog("queue {} epa={} is in the past (now={}), returning nullopt", queue_branch, epa, now_sec);
+         return std::nullopt;
+      }
```
heifner, line 242: This `wlog` fires on every update cycle when the queue is empty / EPA is in the past — which is a normal operating condition, not a warning. With the default `0 * * * *` hourly interval, that's 24 warning-log lines per day per queue during a quiet period. If the interval gets tightened to `*/5 * * * *` (every 5 min), it becomes 288/day/queue.

Suggest either downgrading to `dlog`, or keeping `wlog` but only logging on state transitions (first past-EPA after a valid one).

```suggestion
      if (epa <= now_sec) {
         dlog("queue {} epa={} is in the past (now={}), returning nullopt", queue_branch, epa, now_sec);
         return std::nullopt;
      }
```

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,613 @@
...
+         }
+      }, exit_buffer_days));
+      ilog("There are {} actions currently registered.", actions.size());
+   }
+   else {
+      SYS_ASSERT(!!opp_contract, sysio::chain::plugin_config_exception,
+                 "Nothing is configured to run in wire_eth_maintenance_plugin");
+   }
+
+   auto res = curl_global_init(CURL_GLOBAL_DEFAULT);
```
heifner: `curl_global_init` is documented not-thread-safe and `curl_global_cleanup` behavior is UB if any curl handle is still in use. If a second plugin ever calls `curl_global_init` in the same process (likely, given how nodeop's plugin set has grown, and `programs/clio/do_http_post_libcurl.cpp` already has its own pre-existing init call), state corrupts silently.

Suggest adding `fc::ensure_libcurl_initialized()` to libfc -- a `std::call_once`-guarded wrapper -- and calling that here instead. And dropping the `curl_global_cleanup()` call from `plugin_shutdown` entirely (curl's docs recommend against cleanup when handles might still be in flight; process exit reclaims the state).

Sketch:

```cpp
// libraries/libfc/include/fc/network/curl_init.hpp
namespace fc {
   // Idempotent, thread-safe one-time init of libcurl's global state.
   // Safe to call from any plugin or thread; actual init happens exactly
   // once per process. Intentionally does not register a cleanup handler --
   // curl's global state is reclaimed at process exit.
   void ensure_libcurl_initialized();
}

// libraries/libfc/src/network/curl_init.cpp
#include <fc/network/curl_init.hpp>
#include <fc/exception/exception.hpp>
#include <curl/curl.h>
#include <mutex>
namespace fc {
   void ensure_libcurl_initialized() {
      static std::once_flag flag;
      std::call_once(flag, []() {
         const auto rc = curl_global_init(CURL_GLOBAL_DEFAULT);
         FC_ASSERT(rc == CURLE_OK, "curl_global_init failed: {}", curl_easy_strerror(rc));
      });
   }
}
```

Then in this plugin:
- `plugin_initialize`: replace the current `curl_global_init` + assert with `fc::ensure_libcurl_initialized();`.
- `plugin_shutdown`: drop the `curl_global_cleanup()` call entirely.

While you're in there, also update `programs/clio/do_http_post_libcurl.cpp` to use the same helper.

libraries/libfc/src/io/json.cpp
```diff
@@ -350,14 +363,56 @@ namespace fc
...
+         ? std::string_view(s).substr(start)
+         : std::string_view{};
+
+      // if the string is empty and we dropped zeros
+      if (str.empty() && no_neg_start < start)
+         return 0u;
+
+      // check for s== ".", "-","-.", since "[-]0*" is checked above
+      if (str == "." || str.empty()) // check the obviously wrong things we could have encountered
+         FC_THROW_EXCEPTION(parse_error_exception, "Can't parse token \"{}\" as a JSON numeric constant", str);
```
heifner, line 385: The inline comment here claims `"."`, `"-"`, and `"-."` are the "obviously wrong things we could have encountered", but the test suite doesn't actually exercise `"-."`. `"-"` is addressed in a separate comment on this PR; `"-."` could regress silently too.

Add a companion regression test in `libraries/libfc/test/io/test_json_variant.cpp`:

```cpp
BOOST_AUTO_TEST_CASE(number_from_stream_negative_dot_throws) {
   BOOST_CHECK_THROW(json::from_string("-."), fc::parse_error_exception);
}
```

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,613 @@
...
+            auto& cron_svc = app_ref.get_plugin<sysio::cron_plugin>().cron_service();
+            auto make_retry_opts = []() -> cron_service::retry_options {
+               return cron_service::retry_options{
+                  .retry_schedule = job_schedule{.milliseconds = {job_schedule::step_value{5000}}},
+                  .max_retries = 600,
+                  .on_exhaustion = []() -> fc::exception {
+                     return fc::ethereum_abi_decode_exception(
+                        FC_LOG_MESSAGE(error, "transaction not mined within retry timeout"),
+                        fc::ethereum_abi_decode_exception_code,
+                        "ethereum_abi_decode_exception",
```
heifner: Reusing `fc::ethereum_abi_decode_exception` for a transaction-confirmation timeout is a type mismatch -- this isn't an ABI decoding failure. A developer searching "where do we throw ABI decode exceptions?" lands here and gets misdirected, and anyone pattern-matching on this exception type elsewhere would catch an unrelated condition.

Options:
- Use `sysio::chain::plugin_config_exception` -- closest existing type, since this is an operational timeout surfaced as a hard failure.
- Define a purpose-built `tx_confirmation_timeout_exception` if the plugin (or future plugins) might want to catch this specifically.

Minimum fix is swapping the type; the message can stay the same.

heifner (changes requested):

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
qhool: 'plan to move' is fine by me.

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
brianjohnson5972: I think with how simple this class is, I'd rather keep it as is. I don't think there are any efficiency, size wise or speed wise, in using the variant, and I think the code will not be as easy to read with a variant as it is now.

libraries/libfc/src/io/json.cpp
brianjohnson5972: FYI: @qhool wrote the original request to simplify this, which, since @heifner didn't respond to, I figured he was agreeing. I agree with Kevin here, but at least through out also checking for int128. Hopefully this won't continue to be dueling comments.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
brianjohnson5972: changed it to run in the first millisecond and then cancel the job

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -487,14 +493,23 @@ class ethereum_client : public std::enable_shared_from_this<ethereum_client> {
...
+    * Note: mutex is used for both _contracts_map and _nonce because there will
+    * be little contention on this mutex between them, so there is not really a
+    * need to have _nonce's own mutex
     */
    std::mutex _contracts_map_mutex{};
 
+   /**
+    * @brief Cached nonce for _signature_provider
+    */
+   uint256 _nonce  GUARDED_BY(_contracts_map_mutex) {0u};
```
heifner: The `GUARDED_BY` annotations on `_nonce` (and `_contracts_map`) only fire when the mutex is annotated with `CAPABILITY("mutex")` -- i.e. an `fc::mutex`, not a plain `std::mutex`. With `_contracts_map_mutex` declared as `std::mutex`, clang's thread-safety analysis silently treats both annotations as no-ops. Either switch to `fc::mutex` (the surrounding code already uses `fc::lock_guard`) or drop the annotations so they don't suggest enforcement that isn't happening.

libraries/libfc/include/fc/network/curl_init.hpp
```diff
@@ -0,0 +1,22 @@
+// libraries/libfc/include/fc/network/curl_init.hpp
+namespace fc {
+   // Idempotent, thread-safe one-time init of libcurl's global state.
+   // Safe to call from any plugin or thread; actual init happens exactly
+   // once per process. Intentionally does not register a cleanup handler --
+   // curl's global state is reclaimed at process exit.
+   void ensure_libcurl_initialized();
```
heifner: This header mixes the declaration and a non-`inline` definition of `ensure_libcurl_initialized()` in a single file, with a `// libraries/libfc/src/network/curl_init.cpp` marker comment in the middle that suggests the bottom half was meant to be a separate translation unit. As-is it works because exactly one `.cpp` includes it -- once a second TU pulls it in you'd get a multiple-definition link error.

A few small items:
- Add `#pragma once` (or include guard).
- Either mark the function `inline` if you want a header-only helper, or split into a real `curl_init.cpp` and keep just the declaration here.
- Add a trailing newline.

plugins/cron_plugin/include/sysio/services/cron_service.hpp
```diff
@@ -187,10 +194,78 @@ class cron_service {
...
+         } catch (const fc::exception& e) {
+            auto err = e.dynamic_copy_exception();
+            std::call_once(fired, [&, err]{ promise.set_value(std::unexpected(std::move(*err))); });
+         }
+      };
+
+      auto scheduled_id = this->add(opts.retry_schedule, retry_fn,
+         job_metadata_t{
+            .one_at_a_time = true, .tags = {"ethereum", "gas"}, .label = "beacon_chain_startup"
+         });
```
heifner: The `job_metadata_t` here hardcodes caller-specific values (`tags = {"ethereum", "gas"}`, `label = "beacon_chain_startup"`) inside a generic library helper. Every future caller of `blocking_retry()` -- regardless of what they're retrying -- will appear in cron telemetry as a beacon-chain startup job. Better to expose tags/label through `retry_options` (or fall back to a neutral default like `"blocking_retry"`).

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -12,40 +13,27 @@
...
+   explicit block_tag(labeled name);
+   explicit block_tag(uint64_t bn);
+   std::string to_string() const;
 
-/**
- * @brief Block tag constant representing pending transactions
- */
-constexpr std::string_view block_tag_pending = "pending";
+   const labeled  kind;
+   const uint64_t number;
```
heifner: nit: making `kind` and `number` `const` data members silently disables the implicit copy/move assignment operators. The global `block_tag_latest` / `_pending` / `_earliest` instances are already declared `const` at variable scope, which is what enforces immutability for those. Dropping the `const` on the members keeps the type assignable in case anyone ever wants to reassign a local or store one in a container.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+              "wire_eth_maintenance_plugin uses cron_service::blocking_retry for tx confirmation;"
+              " --cron-threads must be >= 2");
+   auto& oec_plugin = app().get_plugin<outpost_ethereum_client_plugin>();
+   const auto clients = oec_plugin.get_clients();
+   SYS_ASSERT(clients.size() > 0, sysio::chain::plugin_config_exception,
+              "At least one ethereum client must be configured for beacon chain update plugin");
+   const auto eth_client = clients.front()->client;
+
+   ilog("Scheduling {} to execute right after startup", just_once_interval_name);
+   job_schedule jo_schedule = services::parse_cron_schedule_or_throw("*/1 * * * * *");
```
heifner: The 6-field `*/1 * * * * *` parses as "every 1 millisecond in the milliseconds field" -- the cron scheduler will compute and fire ~1000 triggers per second until the lambda's `cancel_job` lands. With `one_at_a_time = true` only one runs at a time, but the scheduler still wakes 1000x/sec for the duration of the first invocation. For a fire-once-asap intent, `"0 0 * * * *"` (ms=0, sec=0, every minute) or a dedicated immediate-post mechanism would be much cheaper.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,607 @@
...
+   constexpr auto beacon_chain_api_key                  = "beacon-chain-api-key";
+   constexpr auto beacon_chain_contracts_addrs          = "beacon-chain-contracts-addrs";
+   constexpr auto beacon_chain_update_interval          = "beacon-chain-update-interval";
+   constexpr auto beacon_chain_interval                 = "beacon-chain-interval";
+   constexpr auto beacon_chain_finalize_epoch_interval  = "beacon-chain-finalize-epoch-interval";
+   constexpr auto beacon_chain_network                  = "beacon-chain-network";
+   constexpr auto beacon_chain_exit_buffer_days         = "beacon-chain-exit-buffer-days";
+
+   constexpr auto client_target_chain                   = fc::crypto::chain_kind_t::chain_kind_ethereum;
+   constexpr auto default_interval_schedule             = "* */1 * * *"; // every hour
```
heifner: `"* */1 * * *"` parses as minute=`*`, hour=`*/1` (which is equivalent to `*`), so this fires every minute, not every hour. For "every hour at minute 0" use `"0 * * * *"` (or `"0 */1 * * *"`).

plugins/cron_plugin/test/test_cron_parser.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+BOOST_AUTO_TEST_CASE(parse_dow_sunday_zero_accepted) try {
+   auto sched_opt = parse_cron_schedule("* * * * 0");
+   BOOST_REQUIRE(sched_opt.has_value());
+   BOOST_CHECK_EQUAL(sched_opt->day_of_week.size(), 1);
+} FC_LOG_AND_RETHROW();
+
+BOOST_AUTO_TEST_CASE(parse_dow_sunday_seven_alias) try {
+   // Documents current behavior for DOW=7 (the crontab Sunday-alias convention).
+   // If parser later adds the alias this test should change to check acceptance.
```
heifner: The lead-in comment is now stale: "If parser later adds the alias this test should change to check acceptance" -- the parser does accept DOW=7 now, and the test was updated to check acceptance. Worth updating the comment so it reflects current behavior.

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
brianjohnson5972: removed

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
brianjohnson5972: I think waiting till the next minute is too long to wait, but I think `*/60 * * * * *` will work

heifner (approved):

jglanz (changes requested):
I took 2 secs to look and saw some glaring issues.  Remove the curl dep - rebase on to feature/opp-part2 - this should have near 0 impact - remove changes from sig prov plugin, etc.  the 2 modifications to existing code that we discussed was adding receive and enhancing the cron parsing, but i even question the need for that

* curl should not be a dep for `fc`,i know its used for clio, but thats legacy

brianjohnson5972:
@jglanz here is the description of all the changes to existing files:

#### cmake/dependencies.boost.cmake

Adding url to boost dependencies for http api support


#### libraries/custom_appbase/include/sysio/chain/app.hpp

Preventing the app plumbing from revealing the api key.


#### libraries/libfc/network/ethereum/ethereum_abi.cpp

Added `receive` for "payable" action. And added support to fc::from_variant for `receive` contract methods with no name. Added default false to_hex parameter to contract_encode_data to pass to the to_hex function.


#### libraries/libfc/network/ethereum/ethereum_client.cpp

Added method get_block_for_transaction as a helper function to retrieve the transaction receipt, if it exists, and return the block number that the transaction is part of.


#### libraries/libfc/src/io/json.cpp

The json processing is (incorrectly) dropping the fact that integer values that are wrapped by quotes, were originally quoted, and is processed just as a bare integer. Fixing this caused a ripple effect in many unit tests and determined that providing support for integers greater than int64 as in256 would be the solution at this time and writing up a Jira task to deal with the complete fix when we are past the critical timeline. Also added returning a variant variable on all paths in that method to ensure ellision is allowed.


#### plugins/cron_plugin/include/sysio/services/cron_service.hpp

Added blocking_retry


#### plugins/cron_plugin/src/services/cron_parser.cpp

Added support for cron schedule format: "* * * * *" (and claude through in a millisecond version: "* * * * * *") to allow configuring a cron schedule for a job taking advantage of the expansive interface for the cron_service.


#### plugins/outpost_ethereum_client_plugin/src/outpost_ethereum_client_plugin.cpp

Change three existing implicitly const methods to have the explicit "const" so they could be called using a const&.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,590 @@
...
+/**
+ * Thread-safety contract: invoked from the signal-catching thread via the
+ * stop_executor_cb registered in main(), concurrently with cron pool workers
+ * that are running https_request. Must only touch thread-safe state —
+ * interruption_handle takes care of that via its atomic flag and
+ * mutex-guarded io_context set.
+ */
+void wire_eth_maintenance_plugin::interrupt() {
+   ilog("interrupt");
+   app().executor().stop();
```
heifner, line 587: This is done for you by custom_appbase. No need to call it here.

programs/cranker/src/main.cpp
```diff
@@ -0,0 +1,39 @@
...
+   chain::application exe{application_config{.enable_resource_monitor = false, .log_on_exit = false}};
+
+   auto r = exe.init<signature_provider_manager_plugin, outpost_ethereum_client_plugin, cron_plugin, wire_eth_maintenance_plugin>(argc, argv);
+   if (r != exit_code::SUCCESS)
+      return r == exit_code::NODE_MANAGEMENT_SUCCESS ? exit_code::SUCCESS : r;
+
+   wire_eth_maintenance_plugin& wire_plug = app().get_plugin<wire_eth_maintenance_plugin>();
+   exe.set_stop_executor_cb([&wire_plug]() {
+      ilog("Exiting cranker");
+      wire_plug.interrupt();
```
heifner, line 22: Not needed as `wire_plug.interrupt()` only does what the default does anyway.

plugins/wire_eth_maintenance_plugin/src/wire_eth_maintenance_plugin.cpp
```diff
@@ -0,0 +1,590 @@
...
+
+void wire_eth_maintenance_plugin::plugin_initialize(const variables_map& options) {
+   ilog("initializing beacon chain plugin");
+
+   if( options.contains(beacon_chain_contracts_addrs) ) {
+      auto client_specs    = options.at(beacon_chain_contracts_addrs).as<std::vector<std::string>>();
+      for(const auto& client_spec : client_specs) {
+         ilog("found beacon chain outpost addresses: {}", client_spec);
+         fc::variant addrs = fc::json::from_file<fc::variant>(client_spec);
+         const auto addrs_obj = addrs.get_object();
```
heifner, line 323: Can be `const auto&`

heifner (changes requested):

PR #281, opened 2026-04-02
Emissions: configurable cadence + epoch-frequency-agnostic gate

heifner:
## Summary

T5 treasury emissions, triggered inline from `sysio.epoch::advance` via a cross-contract readiness gate. Three composable pieces:

1. **Readiness gate.** `sysio.epoch::check_emissions_ready` reads `sysio.system::emitcfg`, `sysio.system::t5state`, and `sysio.token`'s WIRE balance to compute the next epoch's emission. If emissions cannot pay, the gate writes a `blocklog` row on `sysio.epoch` and emits an `EmissionsBlocked` OPP attestation per registered outpost. The chain stays at the prior epoch index until conditions allow a successful pass; the wall-clock duration of the current epoch effectively extends.
2. **Epoch-frequency-agnostic config.** Decay and emission caps are expressed per *year* on `emit_cfg`; per-epoch values are derived inside `compute_epoch_emission` from the canonical `epoch_duration_sec` on `sysio.epoch::epochcfg`. Governance can retune the epoch frequency without distorting the wall-clock emission curve.
3. **Configurable pay cadence.** `emit_cfg.pay_cadence_epochs` controls how many epochs accumulate before payepoch fires. At a 6-min epoch the per-epoch payepoch was firing ~31 inline `send_wire_transfer` calls (21 active + 7 standbys at default `standby_end_rank=28` + 3 batch ops) every 6 min, ~310/hr -- the frequency was the cost driver, not the math. Recommended production value is 100 (≈10 h between pays at 6-min epochs); `setemitcfg` rejects 0.

## Cadence-aware lifecycle

Every successful `advance` queues two emission-side inline actions in FIFO order:

- `accrueepoch(epoch_index, batch_group_index, per_epoch_emission)` -- always. Owns per-epoch state advancement (`last_epoch_index`, `last_epoch_time`, `last_epoch_emission` for decay continuity) plus the period accumulators (`pending_emission_amount`, `batch_group_epochs[]`).
- `payepoch(epoch_index, batch_op_groups, period_emission)` -- pay-epochs only. Drains `pending_emission_amount` and `batch_group_epochs`, distributes the period total across producer / batch / capital / capex / governance pools, resets the accumulator, and advances `period_start_epoch`.

Pay-epoch decision: `target_epoch >= period_start_epoch + pay_cadence_epochs - 1`. With `pay_cadence_epochs=1` this collapses to "every advance is a pay-epoch" -- the legacy per-epoch behavior, which existing tests cover bit-for-bit.

`TREASURY_EXHAUSTED` gates every epoch (pay or non-pay) so the chain cannot silently roll forward into a depleted treasury. The balance-coverage check fires only on pay-epochs against the period total (pending + this-epoch's per-epoch share), so a non-pay epoch never reads `sysio.token::accounts`.

## Producer / batch pool math under cadence > 1

- **Producers (rank 1..21).** Pay is proportional to `eligible_rounds / expected_rounds_period`, where `expected_rounds_period = epoch_duration_sec * pay_cadence_epochs * 2 / TOTAL_BLOCKS_PER_ROUND`. Counters accumulate across non-pay epochs (`accrueepoch` does not reset them); only `payepoch` clears them. Below the truncate-to-zero floor (~126 s effective period duration, e.g. 60 s epoch with cadence=1), `expected_rounds` clamps to 1 and pay collapses to "elig_rounds==0 ? 0 : full_share" -- same fallback as before.
- **Standbys (rank 22..`cfg.standby_end_rank`).** Paid every pay-epoch unconditionally by rank-decreasing weight; no `eligible_rounds` requirement. Default `standby_end_rank=28` gives 7 standbys; safety cap is 100.
- **Batch ops.** `batch_pool` divides into per-group slices weighted by each group's active-epoch count over the period: `group_pool = batch_pool * batch_group_epochs[g] / pay_cadence_epochs`. Members not registered as ACTIVE in `sysio.opreg` (slashed / terminated / unknown) are skipped; their slice stays in the treasury. Groups that were active in zero epochs (only possible when `pay_cadence_epochs < batch_op_groups.size()`) are skipped the same way.
- **Audit log.** `epochlog` gains one row per pay-epoch; non-pay epochs add nothing. Retention pruning is unchanged (head-first, capped at `cfg.epoch_log_retention_count`).

## emission_config (governance-tunable via `setemitcfg`)

**Node-owner allocations**
- `t1_allocation`, `t2_allocation`, `t3_allocation` (int64, WIRE subunits) -- total vesting amount per tier registration.
- `t1_duration`, `t2_duration`, `t3_duration` (uint32, seconds) -- linear vesting period per tier.
- `min_claimable` (int64) -- minimum claim amount; smaller claims are deferred until vested or threshold is met.

**T5 treasury (annual semantics)**
- `t5_distributable` (int64) -- total amount available for T5 emissions across the schedule.
- `t5_floor` (int64) -- treasury reserve. Emissions stop once `t5_distributable - t5_floor - total_distributed <= 0`.
- `target_annual_decay_bps` (uint16, `(0, 10000]`) -- surviving fraction per year; `compute_per_epoch_decay` derives the per-epoch factor from this and `epoch_duration_sec`.
- `annual_initial_emission` (int64) -- E_0 expressed per year.
- `annual_max_emission`, `annual_min_emission` (int64) -- per-year ceiling / floor; per-epoch values derive via `scale_annual_to_epoch`.

**Pay cadence**
- `pay_cadence_epochs` (uint16, `> 0`) -- how many epochs accumulate before payepoch fires. Default helper sets 1; recommended production value 100; soak cluster bootstrap uses 2.

**Audit-log retention**
- `epoch_log_retention_count` (uint32) -- caps `epochlog` rows. payepoch prunes head-first up to two rows per call to drain after a retention-cap shrink.

**Category splits (basis points; must sum to 10000)**
- `compute_bps`, `capital_bps`, `capex_bps`, `governance_bps`.

**Compute sub-split (basis points; must sum to 10000)**
- `producer_bps`, `batch_op_bps`.

**Producer config**
- `standby_end_rank` (uint32, `[22, 100]`) -- last standby rank that receives pay.

## Hardcoded economic constants (in `emissions.hpp`)

- `T1_MAX_NODE_OWNERS = 21`, `T2_MAX_NODE_OWNERS = 84`, `T3_MAX_NODE_OWNERS = 1000` -- per-tier caps; same constants used by `sysio.roa::activateroa` for pool seeding.
- `ACTIVE_PRODUCER_COUNT = 21`, `STANDBY_START_RANK = 22`, `MAX_STANDBY_END_RANK = 100`.
- `BPS_DENOMINATOR = 10000`.
- `TOTAL_BLOCKS_PER_ROUND = 21 * 12 = 252` (per-rotation block count).

## New / changed actions

- `payepoch(epoch_index, batch_op_groups, period_emission)` (sysio.system, `require_auth(sysio.epoch)`) -- pays one pay period; called inline from advance. Signature changed from the per-epoch shape: takes the full `batch_op_groups` vector (so it can pay any group that was active in the period) and the period total.
- `accrueepoch(epoch_index, batch_group_index, per_epoch_emission)` (sysio.system, `require_auth(sysio.epoch)`) -- new. Per-epoch state advancement + accumulator update; runs every successful advance.
- `setemitcfg`, `setinittime`, `addnodeowner`, `claimnodedis`, `viewnodedist`, `initt5`, `viewepoch`, `viewemitcfg` -- emissions configuration and queries.
- `sysio.epoch::advance` -- runs the readiness gate before any state mutation; on block, writes `blocklog` row and queues `EmissionsBlocked` attestations. On gate-pass, queues `accrueepoch` (always) plus `payepoch` (pay-epochs only).
- `sysio.roa::finalizereg` -- gains `require_auth(get_self())` (was missing in master, allowing any caller to bypass registration); corrects status validation so confirmation is reachable.

## New protobuf types

- `ATTESTATION_TYPE_EMISSIONS_BLOCKED = 60953`.
- `EmissionsBlocked` message (`epoch_index`, `reason`, `attempted_emission`, `treasury_remaining`, `sysio_balance`, `first_blocked_at`).
- `EmissionsBlockReason` enum (UNSPECIFIED, CONFIG_MISSING, STATE_UNINITIALIZED, TREASURY_EXHAUSTED, BALANCE_INSUFFICIENT).

## Removed

- `processepoch` action (cranker entry; no caller exists in this design).
- `prunesnap` action and `batchsnap` table on sysio.epoch.
- Cranker idempotency guard and one-epoch-per-call catch-up logic.
- Readonly-mirror header files for sysio.token, sysio.system, sysio.opreg, sysio.epoch (cross-contract reads now use canonical headers directly; the `[[sysio::contract(\"...\")]]` attribute on each table prevents ABI pollution).


contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -42,6 +48,197 @@ using links_t = sysio::kv::table<"links"_n, links_key, links_row,
 
 } // namespace authex_readonly
 
+// ---------------------------------------------------------------------------
+// Emissions readiness gate. Runs in advance() before any state mutation:
+// reads sysio.system::emitcfg + t5state and sysio.token::accounts to compute
+// whether emissions can pay this epoch. If not, advance emits an
+// EmissionsBlocked attestation per outpost (deduped by the local blocklog
```
jglanz, line 24: why would an emissions crank fail?

contracts/sysio.roa/sysio.roa.cpp
```diff
@@ -522,35 +525,28 @@ namespace sysio {
         });
     };
 
-    void roa::finalizereg(const name& owner,const uint8_t& status) {
+    void roa::finalizereg(const name& owner, const uint8_t& status) {
```
jglanz, line 528: status should be an enum?

contracts/sysio.system/include/sysio.system/emissions.hpp
```diff
@@ -0,0 +1,315 @@
...
+
+struct [[sysio::table("epochlog"), sysio::contract("sysio.system")]] epoch_log {
+   uint32_t               sysio_epoch_index = 0;  // mirrors t5_state.last_epoch_index / sysio.epoch::current_epoch_index
+   uint64_t               epoch_count       = 0;  // internal counter of payepoch invocations
+   sysio::time_point_sec  timestamp;
+   int64_t                total_emission    = 0;
+   int64_t                compute_amount    = 0;
+   int64_t                capital_amount    = 0;
+   int64_t                capex_amount      = 0;
+   int64_t                governance_amount = 0;
```
jglanz, line 287: is there any value in using `std::vector<TokenAmount>` to potentially support other token emissions considering our place in the market?

jglanz (changes requested):
I left a few questions - but I'm happy with the PR once you go thru them

contracts/sysio.roa/sysio.roa.cpp
heifner: This was created before enum support. We could change it but I'm not really sure how many clients that would affect.

contracts/sysio.epoch/src/sysio.epoch.cpp
heifner: The gate returns a non-ready result for four reasons, enumerated by `EmissionsBlockReason` and visible cross-chain via the `EmissionsBlocked` attestation:

- `CONFIG_MISSING` -- `sysio.system::emitcfg` has never been set.
- `STATE_UNINITIALIZED` -- `sysio.system::t5state` has not been initialized (`setinittime` never ran).
- `TREASURY_EXHAUSTED` -- `t5_distributable - t5_floor - total_distributed <= 0`. Computed every advance (pay or non-pay) so we never roll forward into a depleted treasury.
- `BALANCE_INSUFFICIENT` -- on a pay-epoch only, `sysio.token` WIRE balance can't cover `pending + this-epoch's share`. Non-pay epochs don't transfer, so no balance check fires.

All four should not fail in practice -- the gate is defense-in-depth so we report a clean cross-chain signal and hold the epoch index, rather than silently corrupting state.

contracts/sysio.system/include/sysio.system/emissions.hpp
heifner: I'm not sure I'm following -- the four fields here (compute/capital/capex/governance_amount) are category splits of a single WIRE emission per the bps fields on `emit_cfg`, not different tokens. Did you mean the treasury could emit in tokens other than WIRE (BTC/USDC/etc.), or something else?

contracts/sysio.authex/CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ if(BUILD_SYSTEM_CONTRACTS)
             $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/../../libraries/opp/generated-cdt>
             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.opp.common/include>
             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.system/include>
+            # cdt-cpp ignores -isystem; add vcpkg root explicitly as -I so magic_enum resolves.
+            PRIVATE
+            ${CDT_CONTRACT_INCLUDE_PATH}
```
jglanz: I had automated the magic_enum issue in `contract-tools.cmake`? what changed

contracts/sysio.chalg/CMakeLists.txt
```diff
@@ -34,6 +34,9 @@ if(BUILD_SYSTEM_CONTRACTS)
                 $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.opp.common/include>
                 $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.epoch/include>
                 $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.system/include>
+                # cdt-cpp ignores -isystem; add vcpkg root explicitly as -I so magic_enum resolves.
+                PRIVATE
```
jglanz: same as above, appplies to all the CMake contract changes

contracts/sysio.epoch/include/sysio.epoch/sysio.epoch.hpp
```diff
@@ -89,6 +89,35 @@ namespace sysio {
...
+         uint32_t                              first_blocked_at   = 0; // unix seconds
+         uint32_t                              last_retry_at      = 0; // unix seconds
+         uint32_t                              retry_count        = 0;
+
+         SYSLIB_SERIALIZE(blocklog_entry,
+            (epoch_index)(reason)(attempted_emission)(treasury_remaining)
+            (sysio_balance)(first_blocked_at)(last_retry_at)(retry_count))
+      };
+
+      using blocklog_t = sysio::kv::table<"blocklog"_n, blocklog_key, blocklog_entry>;
```
jglanz, line 119: is the table auto-pruned?

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -2,6 +2,11 @@
...
 #include <sysio.msgch/sysio.msgch.hpp>
 #include <sysio.authex/sysio.authex.hpp>
+#include <sysio.token/sysio.token.hpp>
+// Canonical sysio.system emissions types + compute_epoch_emission. The
+// [[sysio::contract("sysio.system")]] attribute on emission_config / t5_state
+// pins them to sysio.system's ABI; no readonly mirror needed here.
+#include <sysio.system/emissions.hpp>
 #include <sysio.chains/sysio.chains.hpp>
 #include <sysio/opp/attestations/attestations.pb.hpp>
 #include <zpp_bits.h>
```
jglanz: i think the zpp include comes with the actual proto includes (i could be wrong)

contracts/sysio.epoch/src/sysio.epoch.cpp
jglanz: Personally i'd have an emissions config thats a oneshot call when everything is bootstrapped which would remove the first 2 reasons?

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -33,7 +227,10 @@ void epoch::setconfig(uint32_t epoch_duration_sec,
...
    require_auth(get_self());
 
-   check(epoch_duration_sec > 0, "epoch_duration_sec must be positive");
+   check(epoch_duration_sec >= MIN_EPOCH_DURATION_SEC,
+         "epoch_duration_sec must be >= MIN_EPOCH_DURATION_SEC");
+   check(epoch_duration_sec <= MAX_EPOCH_DURATION_SEC,
+         "epoch_duration_sec exceeds 30-day ceiling");
    check(operators_per_epoch > 0, "operators_per_epoch must be positive");
    check(batch_op_groups > 0, "batch_op_groups must be positive");
    check(batch_operator_minimum_active == operators_per_epoch * batch_op_groups,
```
jglanz, line 235: this particular attestation could cause mega problems in the case that somehow 1 operator hits a performance termination and the replacement has not been activated yet.  All the others look appropriate/correct

contracts/sysio.epoch/CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ if(BUILD_SYSTEM_CONTRACTS)
                 $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.authex/include>
                 $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.chains/include>
                 $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../sysio.opp.common/include>
+                # sysio.token has a flat header layout; expose contracts/ so <sysio.token/sysio.token.hpp> resolves.
+                $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>
+                # cdt-cpp ignores -isystem; add vcpkg root explicitly as -I so magic_enum resolves.
```
jglanz: fix sysio.token layour instead of customizing the CMake config

contracts/sysio.epoch/CMakeLists.txt
```diff
@@ -37,6 +37,11 @@ if(BUILD_SYSTEM_CONTRACTS)
...
+                PRIVATE
+                ${CDT_CONTRACT_INCLUDE_PATH}
```
jglanz: same as above

jglanz (changes requested):
Generally looks good, just a few CMake questions/concerns and a couple of code/contract call outs

contracts/sysio.authex/CMakeLists.txt
heifner: You did automate it -- `CDTWasmToolchain.cmake` already adds `${CDT_CONTRACT_INCLUDE_PATH}` globally via `include_directories(BEFORE ...)`, so the WASM build resolves magic_enum without any per-contract `-I`. These additions were a redundant workaround and are reverted in 52e89afa11. The real gap was the native-module build -- `add_native_contract` sets up the CDT includes but never linked magic_enum; fixed in Wire-Network/wire-cdt#65.

contracts/sysio.chalg/CMakeLists.txt
heifner: Same -- the per-contract magic_enum `-I` additions are reverted across all the contract CMakeLists in 52e89afa11. See the authex thread for details.

contracts/sysio.epoch/CMakeLists.txt
heifner: Done in 52e89afa11 -- `sysio.token` moved from its flat layout to the standard `contracts/sysio.token/include/sysio.token/` layout. epoch and sysio.system now point at `sysio.token/include` like every other contract, so the `contracts/`-on-path customization is gone.

contracts/sysio.epoch/CMakeLists.txt
heifner: This line was the magic_enum `${CDT_CONTRACT_INCLUDE_PATH}` workaround -- reverted in 52e89afa11, same as the authex/chalg CMakeLists.

contracts/sysio.epoch/src/sysio.epoch.cpp
heifner: Confirmed -- `attestations.pb.hpp` includes `<zpp_bits.h>` directly, so the explicit include was redundant. Removed in c151993f5a.

contracts/sysio.epoch/src/sysio.epoch.cpp
heifner: Fixed in c151993f5a. `advance()` no longer aborts when the new tail group can't be filled -- it pushes the (possibly empty) group so the sliding window keeps its N slots and advancement always completes. The shortage is reported via `print()` and stays visible cross-chain in the `BatchOperatorGroups` attestation, so it's observable without halting OPP. Recovery is operator-roster repair (activating replacements).

contracts/sysio.epoch/include/sysio.epoch/sysio.epoch.hpp
heifner: Yes. `blocklog` only holds rows for epochs currently failing the emissions gate: `record_gate_block` inserts on a block, `clear_gate_block` erases on the next successful gate pass. `advance()` only ever targets `current_epoch_index + 1`, so at most the single in-flight blocked epoch is recorded and the table is empty in steady state -- no separate sweep needed.

jglanz (approved):
please double check the history, its says you are "creating" the chains contract and reserv contract, which makes me think the history may be out of whack....double check its a clean rebase

PR #283, opened 2026-04-03
Chainbase internal optimizations: session allocs, spinlock, snapshot preallocate

heifner:
## Summary

- **Eliminate per-transaction heap allocations in undo sessions** — Replace `vector<unique_ptr<abstract_session>>` (18 heap allocations per transaction) with a lightweight `database* + bool` pair. The `abstract_session` / `session_impl` virtual dispatch layer was redundant since `database::undo()` and `database::squash()` already iterate `_index_list` with the same dispatch through `abstract_index`.

- **Remove null terminator from shared_cow_string** — `shared_blob` stores binary data (KV keys, values, ABI blobs) where null termination is unnecessary. Saves 1 byte per allocation; with 8-byte slab bucket rounding this saves 8 bytes per allocation that crosses a bucket boundary (e.g., 24-byte keys: 33 -> 32 bytes).

- **Replace std::mutex with spinlock in small_size_allocator** — Reduces per-bucket overhead from ~40 bytes (pthread_mutex_t) to 1 byte (atomic_flag), saving ~5KB across 128 buckets in shared memory. Uncontended spinlock is ~7-10ns vs ~25ns for mutex, saving ~15ns per alloc+dealloc cycle.

- **Preallocate chainbase node storage during snapshot loading** — Expose per-section row count from snapshot readers, then batch-allocate node storage upfront before the row creation loop. Avoids repeated `get_some()` calls to the segment manager during row-by-row insertion. Covers all index loading paths (controller, KV, authorization, resource limits).

### DIFF
--- a/libraries/chaindb/src/chainbase.cpp
+++ b/libraries/chaindb/src/chainbase.cpp
@@ -68,12 +68,9 @@ namespace chainbase {
       if ( _read_only_mode )
          BOOST_THROW_EXCEPTION( std::logic_error( "attempting to start_undo_session in read-only mode" ) );
       if( enabled ) {
-         vector< std::unique_ptr<abstract_session> > _sub_sessions;
-         _sub_sessions.reserve( _index_list.size() );
-         for( auto& item : _index_list ) {
-            _sub_sessions.push_back( item->start_undo_session( enabled ) );
-         }
-         return session( std::move( _sub_sessions ) );
+         for( auto& item : _index_list )
+            item->add_undo_session();
+         return session( *this );
       } else {
          return session();
       }
--- a/libraries/chaindb/src/chainbase.cpp
+++ b/libraries/chaindb/src/chainbase.cpp
@@ -27,20 +27,26 @@ namespace chainbase {
    {
       if ( _read_only_mode )
          BOOST_THROW_EXCEPTION( std::logic_error( "attempting to undo in read-only mode" ) );
-      for( auto& item : _index_list )
-      {
-         item->undo();
-      }
+      undo_from_session();
    }
 
    void database::squash()
    {
       if ( _read_only_mode )
          BOOST_THROW_EXCEPTION( std::logic_error( "attempting to squash in read-only mode" ) );
+      squash_from_session();
+   }
+
+   void database::undo_from_session()
+   {
       for( auto& item : _index_list )
-      {
+         item->undo();
+   }
+
+   void database::squash_from_session()
+   {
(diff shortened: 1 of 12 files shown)


libraries/chaindb/src/chainbase.cpp
```diff
@@ -68,12 +74,9 @@ namespace chainbase {
       if ( _read_only_mode )
          BOOST_THROW_EXCEPTION( std::logic_error( "attempting to start_undo_session in read-only mode" ) );
       if( enabled ) {
-         vector< std::unique_ptr<abstract_session> > _sub_sessions;
-         _sub_sessions.reserve( _index_list.size() );
-         for( auto& item : _index_list ) {
-            _sub_sessions.push_back( item->start_undo_session( enabled ) );
-         }
-         return session( std::move( _sub_sessions ) );
+         for( auto& item : _index_list )
```
huangminghuang: database::start_undo_session(true) now calls add_undo_session() across indexes directly, but if a later index allocation throws, earlier indexes keep their newly-pushed undo session and no database::session is returned to clean them up. The old vector-of-RAII-sessions path unwound partial sessions automatically. This can leave indexes with different undo-stack depths/revisions after an allocation failure, corrupting later undo/squash/commit behavior. I’d restore whole-database exception safety by rolling back already-added sessions in a catch block, or by using a small stack/local guard that calls undo() for indexes added so far until ownership is transferred to the returned session.

heifner: Good catch -- fixed in `dda1f15b23`.

`start_undo_session` now tracks how many indexes received a session and, on a throw, rolls back exactly those before rethrowing:

```cpp
size_t added = 0;
try {
   for( ; added < _index_list.size(); ++added )
      _index_list[added]->add_undo_session();
} catch( ... ) {
   while( added > 0 )
      _index_list[--added]->undo();
   throw;
}
return session( *this );
```

One thing worth calling out on why this rolls back only the `added` indexes rather than constructing the `session` early and letting its destructor clean up: `session::undo()` undoes *every* index via `undo_from_session()`, but the indexes typically already hold a parent session during nested block/transaction processing. Undoing all of them would pop the enclosing session from the indexes that never received a new one. So the rollback is scoped to exactly the indexes pushed so far. `undo_index::undo()` is `noexcept` and cleanly pops the just-added (empty) session, and `add_session()` is strongly exception-safe, so the failing index is left untouched and the database returns to its pre-call state.

huangminghuang (approved):

PR #284, opened 2026-04-04
Fix secondary index scope isolation and harden chain_plugin

heifner:
## Summary

- **Bug:** Secondary index iteration in `get_table_rows` leaked entries across scopes because `sec_key` had no scope discriminator.
- **Fix (CDT-side encoding, no intrinsic signature changes):**
  - `sec_key`: prepend `[scope:8B BE]` prefix
  - `pri_key`: shrink from `[scope:8B][pk:8B]` to `[pk:8B]`
  - Drop unused `by_code_table_idx_prikey` chainbase index (consensus change)
  - `kv_index_object` billing: 152 → 120 bytes per entry (-21%)

- **Chain plugin hardening:**
  - `collect_next` validates scope prefix and minimum size before pointer arithmetic (prevents `size_t` underflow and cross-scope `next_key` leak on deadline timeout)
  - `pri_key` size guard (`kv_pri_key_size`) in both forward and reverse iteration prevents OOB read
  - New constants: `kv_scope_prefix_size`, `kv_pri_key_size`

- **Tests:** 7 cross-scope tests (iteration, find, erase, upper_bound, double secondary, scope=0, reverse)
- Recompiled production contracts and regenerated snapshot reference data
- Updated RAM billing docs with EOS mainnet estimates from actual snapshot data

**Depends on:** Wire-Network/wire-cdt#47 (CDT encoding changes)


plugins/chain_plugin/src/chain_plugin.cpp
```diff
@@ -1832,31 +1848,38 @@ read_only::get_table_rows_by_seckey( const read_only::get_table_rows_params& p,
    };
 
    auto collect_next = [&](const chain::kv_index_object& obj) {
+      // Validate sec_key has scope prefix and matches our scope
+      if (obj.sec_key.size() < chain::kv_scope_prefix_size ||
+          memcmp(obj.sec_key.data(), scope_be, chain::kv_scope_prefix_size) != 0)
```
brianjohnson5972: I think we should either have a lambda to do this checking of `sec_key==scope_be`
or possibly make a struct for scope_be, with a method for this, that also returns sec_data, and possibly other scope_be stuff, like a constructor above and decode_sec_key being a method. I'll let you decide between which makes sense. Also used at lines 1873, and 1919.

brianjohnson5972 (approved):
Some comments, otherwise good

PR #288, opened 2026-04-09
KV table_id namespace isolation

heifner:
## Summary

Replace `key_format` (uint8, 0/1) with `table_id` (uint16, DJB2 hash of table name % 65536). Each table and secondary index gets a unique `table_id` computed at compile time by CDT, passed as the first parameter to all KV intrinsics. Keys no longer embed the table name, saving 8 bytes per row.

**Companion PR:** Wire-Network/wire-cdt#49 (CDT table_id support)

### Chain changes

- **kv_table_objects.hpp** — `kv_object` and `kv_index_object` use `table_id` (uint16) replacing `key_format` (uint8) and `table`/`index_id` fields. Composite index `by_code_key` sorts by `(code, table_id, key)`
- **types.hpp** — DJB2 hash functions (`kv_djbh_hash`, `compute_table_id`, `compute_sec_table_id`) matching CDT implementation
- **kv_database.cpp** — `checked_table_id()` validates table_id <= UINT16_MAX at WASM boundary, replacing silent truncation
- **apply_context.cpp** — All KV intrinsic implementations updated for table_id parameter
- **kv_context.hpp** — KV context interface uses table_id throughout
- **interface.hpp** — WASM intrinsic declarations updated (first param: `uint32_t table_id`)
- **abi_def.hpp** — `table_def` gains `table_id` field and `secondary_indexes` array with per-index `name`, `key_type`, `table_id`; table name changed from `name` to `string` for long names
- **abi_serializer.cpp** — Handle new `table_def` format with secondary indexes
- **deep_mind.cpp** — KV_OP logs include table_id
- **config.hpp** — Remove legacy KV config constants replaced by table_id

### SHiP (State History)

- `contract_row_kv_v0` emits `table_id` (uint16) instead of legacy fields
- `contract_index_kv_v0` emits `table_id` replacing `table` (uint64) + `index_id` (uint8)
- `create_deltas` updated for new object layouts

### chain_plugin API

- **get_kv_rows** — New endpoint: query by primary key or secondary index name, with ABI-decoded JSON keys/values and pagination via `next_key`
- **get_table_by_scope** — Resolve table names from ABI reverse map via `tid_to_name` (fixes Spring#615 pagination loop). Response `table` field changed from `name` to `string` to support long table names
- **get_table_rows** — Compute `table_id` from table name for KV lookups
- **get_currency_balance / get_currency_stats** — Updated for table_id-based key lookups

### Contracts

- All production contracts recompiled with scope-aware CDT (new ABIs with `table_id` and `secondary_indexes`)
- All test contracts recompiled

### Tests

- **kv_table_id_isolation** — Chainbase-level test: two rows with same `(code, key)` but different `table_id` coexist independently
- **table_id_overflow_rejected** — WASM intrinsic rejects `table_id` 65536 (> UINT16_MAX)
- **SHiP delta tests** — Strengthened with `compute_table_id` value assertions and negative `contract_row` check
- **get_table_by_scope** — Unfiltered queries return ABI-resolved table names; pagination token format verified
- Regenerated snapshots, deep-mind log, consensus blockchain reference data
- Updated stale format=0/key_format comments


libraries/chain/include/sysio/chain/abi_def.hpp
```diff
@@ -57,17 +57,25 @@ struct action_def {
...
+   table_def(const string& name, const type_name& index_type, const vector<field_name>& key_names, const vector<type_name>& key_types, const type_name& type, uint16_t table_id = 0)
+   :name(name), index_type(index_type), key_names(key_names), key_types(key_types), type(type), table_id(table_id)
    {}
 
-   table_name         name;        // the name of the table
-   type_name          index_type;  // the kind of index, i64, i128i128, etc
-   vector<field_name> key_names;   // names for the keys defined by key_types
-   vector<type_name>  key_types;   // the type of key parameters
-   type_name          type;        // type of binary data stored in this table
+   string              name;              // table name (free-form string)
```
brianjohnson5972, line 72: Is the size of name itself limited, or does it just get limited by the ultimate size of key it creates?

libraries/chain/include/sysio/chain/types.hpp
```diff
@@ -79,9 +79,56 @@ namespace sysio::chain {
    template<typename T>
    using deque = boost::container::deque< T, void, block_1024_option_t >;
 
+   // ── KV table_id computation ─────────────────────────────────────────────────
+   // DJB2 hash truncated to uint16_t. Must match CDT's compute_table_id
+   // in sysiolib/contracts/sysio/kv_constants.hpp.
+
+   /// DJB2-hash a string.
+   inline constexpr uint64_t kv_djbh_hash(std::string_view s) {
+      uint64_t hash = 5381;
```
brianjohnson5972: should have a `constexpr uint64_t djbh_start_value = 5381;` or something like that

libraries/chain/include/sysio/chain/types.hpp
```diff
@@ -79,9 +79,56 @@ namespace sysio::chain {
...
+      for (char c : s)
+         hash = ((hash << 5) + hash) + static_cast<uint8_t>(c);
```
brianjohnson5972, line 94: add a comment like:
```
            // hash * 33 (s^5 + 1)
hash = ((hash << 5) + hash) + static_cast<uint8_t>(c);
```
and similarly in the uint64 hash function below

libraries/chain/include/sysio/chain/types.hpp
```diff
@@ -79,9 +79,56 @@ namespace sysio::chain {
...
+   inline constexpr uint64_t kv_djbh_hash_raw(uint64_t raw) {
+      uint64_t hash = 5381;
+      for (int i = 0; i < 8; ++i)
+         hash = ((hash << 5) + hash) + static_cast<uint8_t>(raw >> (56 - i * 8));
+      return hash;
+   }
+
+   /// Compute table_id from a string name (for test/tooling convenience).
+   inline constexpr uint16_t compute_table_id(std::string_view table_name) {
+      return static_cast<uint16_t>(kv_djbh_hash(table_name) % 65536);
```
brianjohnson5972: Is there a problem with just doing:
`return static_cast<uint16_t>(kv_djbh_hash(table_name));`
because I think it would be clearer. Same in other areas.

If I'm wrong about that being the same thing, I think this would be clearer (and have fewer instructions) if it was replaced with `& 0xffff` or else it would be clearer in hex, 0x10000
and similarly for methods below

libraries/chain/include/sysio/chain/types.hpp
```diff
@@ -79,9 +79,56 @@ namespace sysio::chain {
...
+   /// Compute table_id from a raw uint64_t template parameter (name::raw or hash_id::raw).
+   /// This is the canonical form used by CDT — matches CDT's compute_table_id(uint64_t).
+   inline constexpr uint16_t compute_table_id(uint64_t raw) {
+      return static_cast<uint16_t>(kv_djbh_hash_raw(raw) % 65536);
+   }
+
+   /// Compute secondary index table_id from table + index raw uint64_t values.
+   inline constexpr uint16_t compute_sec_table_id(uint64_t table_raw, uint64_t index_raw) {
+      uint64_t hash = kv_djbh_hash_raw(table_raw);
+      for (int i = 0; i < 8; ++i)
```
brianjohnson5972: Can we make a private method that just does the loop from line 118 to 121? And then call that here and call it inside kv_djbh_hash_raw(...) after it initializes `uint64_t hash = djbh_start_value;// 5381`
alternatively, we could change kv_djbh_hash_raw to have the signature:
```
inline constexpr uint64_t kv_djbh_hash_raw(uint64_t raw, uint64_t hash = djbh_start_value)  {
```
and then this code would just be
```
uint64_t hash = kv_djbh_hash_raw(table_raw);
return static_cast<uint16_t>(kv_djbh_hash_raw(table_raw, hash) % 65536);
```

libraries/chain/webassembly/kv_database.cpp
```diff
@@ -3,26 +3,32 @@
 
 namespace sysio { namespace chain { namespace webassembly {
 
+   static uint16_t checked_table_id(uint32_t table_id) {
+      SYS_ASSERT(table_id <= std::numeric_limits<uint16_t>::max(), kv_key_too_large,
```
brianjohnson5972: I think having a `constexpr uint32_t ui32_max = std::numeric_limits<uint16_t>::max;` will improve readability

libraries/chain/include/sysio/chain/types.hpp
```diff
@@ -95,14 +142,35 @@ namespace sysio::chain {
...
+   struct kv_scoped_key_t {
+      static constexpr size_t size = kv_scoped_key_size;
+      char data[size];
+      std::string_view to_string_view() const { return {data, size}; }
+   };
+
+   inline kv_scoped_key_t make_kv_scoped_key(uint64_t scope, uint64_t pk) {
+      kv_scoped_key_t key;
+      kv_encode_be64(key.data,     scope);
+      kv_encode_be64(key.data + 8, pk);
```
brianjohnson5972: I think it would be easier to review later code changes if we make it `kv_encode_be64(key.data + kv_scope_prefix_size, pk);`
or else could add another constexpr of `kv_scope_offset`, but would prefer just re-using the existing `kv_scope_prefix_size`

libraries/chain/apply_context.cpp
```diff
@@ -1052,24 +1048,24 @@ int32_t apply_context::kv_idx_find_secondary(name code, name table, uint8_t inde
...
+   auto itr = idx.lower_bound(boost::make_tuple(code, table_id, sv_sec));
 
-   if (itr == idx.end() || itr->code != code || itr->table != table || itr->index_id != index_id) {
-      auto first = idx.lower_bound(boost::make_tuple(code, table, index_id));
-      if (first != idx.end() && first->code == code && first->table == table && first->index_id == index_id) {
-         uint32_t handle = kv_iterators.allocate_secondary(code, table, index_id);
+   if (itr == idx.end() || itr->code != code || itr->table_id != table_id) {
+      auto first = idx.lower_bound(boost::make_tuple(code, table_id));
+      if (first != idx.end() && first->code == code && first->table_id == table_id) {
+         uint32_t handle = kv_iterators.allocate_secondary(code, table_id);
```
brianjohnson5972: nit: make const

libraries/chain/apply_context.cpp
```diff
@@ -1052,24 +1048,24 @@ int32_t apply_context::kv_idx_find_secondary(name code, name table, uint8_t inde
...
          auto& slot = kv_iterators.get(handle);
          slot.status = kv_it_stat::iterator_end;
          return static_cast<int32_t>(handle);
       }
       return -1;
    }
 
-   uint32_t handle = kv_iterators.allocate_secondary(code, table, index_id);
+   uint32_t handle = kv_iterators.allocate_secondary(code, table_id);
```
brianjohnson5972: nit: make const

unittests/test-contracts/bench_kv_db/bench_kv_db.cpp
```diff
@@ -94,7 +93,7 @@ class [[sysio::contract("bench_kv_db")]] bench_kv_db : public contract {
    [[sysio::action]]
    void iterall() {
       uint64_t code = get_self().value;
-      uint32_t handle = kv_it_create(key_format, code, nullptr, 0);
+      uint32_t handle = kv_it_create(bench_table_id, code, nullptr, 0);
```
brianjohnson5972: nit: make const

libraries/chain/include/sysio/chain/webassembly/interface.hpp
```diff
@@ -782,8 +782,7 @@ namespace webassembly {
           * write to another contract's namespace. Keys are arbitrary byte sequences
           * (max 256 bytes), values up to 256 KiB.
           *
-          * @param key_format - encoding hint: 0 = raw bytes, 1 = standard 24-byte
-          *   `[table:8B][scope:8B][pk:8B]` layout used by `multi_index`
+          * @param table_id - table namespace identifier (DJB2 hash of table name % 65536)
```
brianjohnson5972: I think "lower 16 bits of  the DJB2 hash of table name" would be a more straight forward description
and similar comment edits throughout this file

brianjohnson5972 (approved):

brianjohnson5972:
Approved, with some minor improvements that don't have to be made

libraries/chain/include/sysio/chain/abi_def.hpp
heifner: The `name` field has no explicit length limit, consistent with all other strings in `abi_def` (`struct_def.name`, `field_def.name`, `action_def.ricardian_contract`, `clause_pair.id/body`, etc. — none are length-limited).

Practical bounds:
- The whole ABI is capped by `max_transaction_size` (524288 bytes) since `setabi` is an action.
- RAM cost for the stored ABI is billed to the setabi caller.
- The `table_id` field is what the chain actually uses on-chain; the `name` is only used for ABI lookups by RPC tooling (`get_kv_table_def` at chain_plugin.cpp:1637).

A 510K name wouldn't break anything but would slow down RPC table lookups (linear `name == table_name` scans). The chain itself never re-hashes the name — `compute_table_id` is only used by CDT codegen and tests, so hash collision risk is a CDT/build-time concern, not runtime. Capping all abi_def strings would be a worthwhile separate hardening PR.

PR #295, opened 2026-04-14
trace_api: history-solution upgrade - auto ABIs, query APIs, indexes

heifner:
## Summary

Upgrades `trace_api_plugin` to serve as a complete history solution for exchanges and indexers. Five themes:

1. **Auto-captured ABIs.** `--trace-rpc-abi` and `--trace-no-abis` are removed. Plugin captures ABIs from chain directly (setabi traces in real time, lazy-fetch on first encounter), versioned per `global_sequence` so actions decode with the exact ABI that was in effect when they ran. Capture only runs for block-context executions (`producer_block_id` set), so speculative execution can never poison the version history. Append-only `abi_log` replaces the earlier O(N^2) `abi_store` rewrite; a failed append truncates its torn tail so recorded blob offsets always stay aligned.

2. **Continuity enforced at startup.** Gaps between existing trace data and chain head shut the node down with operator-facing guidance (load a snapshot covering the gap, copy slices from another node, or delete the trace dir). Holes in the *middle* of the recorded range (deleted or partially copied slices) are detected the same way via filename contiguity. Overlap from a snapshot restore is tolerated (re-applied blocks overwrite existing slice entries). The LIB and last-recorded-block watermarks are re-seeded from disk at startup, so irreversible blocks keep reporting `irreversible` across restarts.

3. **O(1) lookups.** Per-slice `trace_trx_idx_<range>.log` (mmap'd open-addressing hash table) replaces the linear scan in `get_transaction_trace`; per-slice `trace_blk_idx_<range>.log` (sparse `uint64_t` array) replaces the linear scan in `get_block`. Index hits are confirmed against the block's recorded trace (one offset-based read) to rule out 64-bit prefix collisions. Both sidecars are built alongside the existing metadata log and fall back to the scan if a sidecar is missing or corrupt; sidecar *write* failures are advisory too (the node keeps running and queries fall back to the scan).

4. **New query endpoints** over the existing trace data.

5. **Per-slice bloom for `get_actions` slice-skip.** New `trace_recv_bloom_<range>.log` sidecar carries two `boost::bloom` filters (K=7, FPR=0.01): one over action `receiver`, one over packed `(receiver, action)` pairs. `get_actions` probes the bloom once per slice in the requested block range; on a negative probe the entire slice is skipped with no `get_block` call. Turns "receiver never appears in this slice" from a full-slice scan into a single file read. Missing or CRC-corrupt sidecar falls back to the scan (fail-safe - a false negative would silently drop matching actions from the response). Built by the maintenance thread once the slice is fully irreversible, so forks can never reach back into a written sidecar - and a data log that does not parse cleanly end-to-end never produces a (partial) bloom at all. Cleaned up alongside the other per-slice files during retention pruning.

### `POST /v1/trace_api/get_actions`

Paginated search over action traces across a block range with optional filters. The handler clamps the block window to `[block_num_start, block_num_start + trace-max-block-range - 1]` *and* to the last block recorded by the node, and reports the actual window scanned on the response - so clients can resume pagination by advancing `block_num_start` to `block_num_end + 1` without ever skipping blocks that had not been produced (or recorded) at request time.

Request fields:

| Field | Type | Default | Description |
|---|---|---|---|
| `block_num_start` | uint32 | `0` | First block to scan (inclusive). |
| `block_num_end` | uint32 | `UINT32_MAX` | Last block to scan; clamped server-side to the range cap and the last recorded block. |
| `receiver` | string | any | Filter on `act.receiver`. |
| `account` | string | any | Filter on `act.account` (contract code). |
| `action` | string | any | Filter on action name. |
| `include_notifications` | bool | `false` | When `false`, specifying exactly one of `receiver`/`account` implicitly constrains the other to the same value (canonical execution only, no notification copies). When `true`, only the specified filters apply. |

Response rows carry the full action-trace execution-tree context (`action_ordinal`, `creator_action_ordinal`, `recv_sequence`, `auth_sequence`, `code_sequence`, `abi_sequence`, `account_ram_deltas`, optional per-action `cpu_usage_us` / `net_usage`), the raw hex `data` / `return_value` (always present), the ABI-decoded `params` and `return_data` when decoding succeeds (or a `decode_error` field when it fails), and the enclosing block / transaction context (`trx_id`, `block_num`, `block_time`, `producer_block_id`, `block_status`, `trx_cpu_usage_us`, `trx_net_usage_words`). `block_status` is `"irreversible"` or `"pending"`, sourced from the same data-log tuple that powers `get_block`'s `status` field and seeded from disk at startup so it survives restarts; operators wanting only-irreversible responses can run nodeop with `read-mode = irreversible`. Within a transaction actions are ordered by `global_sequence` (execution order). `trx_cpu_usage_us` and `trx_net_usage_words` are the parent transaction's totals; they differ from the action-level `cpu_usage_us` / `net_usage` in scope (whole trx vs single action) and units (action `net_usage` is bytes; trx `net_usage_words` is `ceil(net_usage / 8)`).

### `POST /v1/trace_api/get_token_transfers`

Convenience preset of `get_actions` with `receiver = account = <token_contract>` and `action = transfer`. Defaults `token_contract` to `sysio.token`. Because it constrains `receiver == account`, each transfer appears exactly once — the canonical execution, not the inline notifications to sender and recipient.

Returns a slim subset of `get_actions` fields optimized for exchange/indexer workflows (omits execution-tree ordinals, receipt sequences, RAM deltas, and CPU/NET usage); `block_status` is retained because exchanges crediting transfers need finality. Use `get_actions` directly if you need the full row shape.

## Breaking changes

- Removed CLI options: `--trace-rpc-abi`, `--trace-no-abis`. Operators no longer supply `abi.json` files.
- `action_trace_v0` gains `action_ordinal`, `creator_action_ordinal`, `closest_unnotified_ancestor_action_ordinal`, `recv_sequence`, `auth_sequence`, `code_sequence`, `abi_sequence`, `account_ram_deltas`, optional `cpu_usage_us` / `net_usage`. Persisted trace format is incompatible with earlier builds; operators must delete the trace directory on upgrade.
- Failed actions no longer persisted; `status` removed from `transaction_trace_v0`.
- JSON field renames: `action` -> `name` on actions, `account` -> `actor` on authorizations.

## New config

- `--trace-max-block-range` (default `1000`, clamped `[1, 10000]`) — silently caps `block_num_end - block_num_start + 1` for the range-scanning query endpoints.

## On-disk layout (additions)

| File | Scope | Purpose |
|---|---|---|
| `trace_trx_idx_<range>.log` | per slice | mmap'd hash table: `trx_id_prefix64 -> block_num` |
| `trace_blk_idx_<range>.log` | per slice | sparse `uint64_t[stride]` block_num -> trace offset |
| `trace_recv_bloom_<range>.log` | per slice | `boost::bloom` filters over action `receiver` and packed `(receiver, action)` pairs; probed by `get_actions` for O(1) slice-skip |
| `abi_log.log` | global | append-only records `(account, global_seq, blob)` + CRC32 |

## Docs

Full endpoint reference, configuration table, on-disk layout, ABI versioning semantics, pagination guide, and exchange/indexer integration notes live in [`plugins/trace_api_plugin/trace_api_plugin.md`](plugins/trace_api_plugin/trace_api_plugin.md).


plugins/trace_api_plugin/include/sysio/trace_api/chain_extraction.hpp
```diff
@@ -54,6 +72,90 @@ class chain_extraction_impl_type {
...
+
+      // ABI capture only from block-context executions (production, validation, replay), i.e.
+      // producer_block_id is set.  Speculative executions carry global_sequences that may never
+      // match canonical history, and a speculative-only setabi (or a lazy fetch against
+      // speculative state) could record an ABI version that never lands on-chain, shadowing the
+      // correct version for lookups near that sequence.  Block (re-)execution emits this signal
+      // again with canonical sequences, so skipping the speculative pass loses nothing.  The
+      // trace caching above stays unconditional - on_accepted_block filters it against the
+      // block's actual receipts, which is the long-standing fork/speculation guard for traces.
+      if( !trace->producer_block_id ) return;
```
huangminghuang: This captures lazy ABIs and setabi records during every block-context applied_transaction, and appends them permanently at lines 129-151. Trace/block/trx sidecars handle forks by “last write wins” per block, but abi_log has no equivalent rollback or tombstone. If a forked-out block contains setabi(X) and the canonical replacement at that height does not, the forked ABI remains in (X, global_seq) history and later canonical actions on X can decode with a schema that never existed on the canonical chain. This can occur during ordinary fork switching, not just speculative execution. ABI capture needs to be tied to canonical block storage with fork replacement semantics, or include per-block undo/overwrite metadata so forked-out ABI versions stop participating in lookup.

huangminghuang (approved):

plugins/trace_api_plugin/include/sysio/trace_api/chain_extraction.hpp
heifner: Addressed in 0255b964. The abi log now mirrors how the other trace files treat LIB: the on-disk file only ever contains records for irreversible blocks (so a written record can never be invalidated by a fork), while records from accepted-but-not-final blocks sit in an in-memory reversible overlay tagged with their block number. A `block_start` at height H discards overlay records with block >= H (fork replacement, before the replacing block's transactions execute, so the lazy-fetch first-encounter check also sees canonical state), and `append_lib` flushes records at/below LIB to the file. On restart the overlay is rebuilt from the already-recorded setabi action traces in the (LIB, last_recorded] window. The forked-setabi scenario you describe is covered by new tests at the abi_log, extraction, and store_provider levels (`forked_out_abi_never_reaches_disk`, `fork_switch_replaces_committed_abi_records`, `test_reversible_abi_rebuilt_from_traces_on_restart`).

plugins/trace_api_plugin/src/store_provider.cpp
```diff
@@ -304,14 +449,402 @@ namespace sysio::trace_api {
...
+            get_block_t bt = get_block(block_num);
+            if (!bt)
+               continue; // hole in the reversible window (e.g. fork overwrite gap) - nothing to rebuild
+            const auto& block = std::get<block_trace_v0>(std::get<0>(*bt));
+            for (const auto& trx : block.transactions) {
+               for (const auto& at : trx.actions) {
+                  // Same detection as chain_extraction's live capture: a setabi action on the
+                  // system account.  The recorded action trace carries the exact
+                  // global_sequence and the raw action payload (target account + ABI bytes).
+                  if (at.account != chain::config::system_account_name || at.action != setabi_action_name)
```
huangminghuang: This only rebuilds reversible ABI overlay entries from recorded setabi actions, while lazy global_sequence 0 captures collected in chain_extraction.hpp (line 135) also live only in the reversible overlay until LIB. After a clean restart before LIB, those lazy records are gone. If the account does not appear again before/after finalization, append_lib() has nothing to flush, so already-recorded historical actions remain undecodable. Worse, if the same reversible window contains a later setabi, the rebuilt setabi makes has_abi_entry() true and prevents future lazy global_sequence 0 recapture, so pre-setabi actions that decoded before restart degrade to raw permanently. The rebuild path needs to preserve lazy captures too, likely via a fork-aware reversible ABI journal/trace metadata rather than relying on future encounters.

huangminghuang (changes requested):

plugins/trace_api_plugin/src/store_provider.cpp
heifner: Good catch — confirmed and fixed in a6a052d621.

You're right that the trace-scan rebuild structurally can't recover lazy `global_seq=0` records: they're read from chain state at first-encounter and leave no action trace, so scanning recorded traces never sees them. And the Case-B interaction is real — a rebuilt setabi flips `has_entry()` true and suppresses the lazy re-capture, so pre-setabi actions degrade to raw permanently. Since the pre-setabi ABI is gone from chain state (state holds only the latest version), it can't be re-derived after the fact — the only fix is to persist the overlay *before* the restart.

So I replaced the trace-scan rebuild with a durable append-only journal folded into `abi_log` (`abi_log.journal`):

- `append_reversible` → `PUT`, `rollback_reversible` → `ROLLBACK` (only when it discards something), `flush_irreversible` writes nothing (the record is already in the main log).
- On startup the journal replays in order into the overlay, drops any `(account, global_seq)` already on disk, then compacts. This restores **both** lazy and setabi reversible records exactly, so `rebuild_reversible_abis()` is deleted.
- Fork-safe by construction: the `ROLLBACK` is journaled and replayed in order, so a forked-out record never resurrects across a restart. Torn tails truncate like the main log; a write failure is advisory.

Bounded for long-uptime nodes: `flush_irreversible` truncates the journal to its header when the overlay drains, and compacts it down to the live overlay once it passes a size threshold.

Tests: lazy-survives-restart, lazy+setabi both survive with the pre-setabi seq still decoding (your Case B), forked-out-not-resurrected, rollback+reappend-same-height and partial-window rollback across restart, flushed-compacted-out, drain reset, live-compaction bound, torn-journal-tail recovery, plus a single-writer/4-reader concurrency stress. 154/154 in `test_trace_api_plugin`; `abi_log_tests` clean under UBSan.

plugins/trace_api_plugin/src/abi_log.cpp
```diff
@@ -0,0 +1,783 @@
...
+      for (const auto& [key, entry] : _reversible) {
+         if (entry.block_num >= block_num) { any = true; break; }
+      }
+   }
+   if (!any)
+      return;
+
+   // Journal the rollback before applying it: persisting the intent first means a crash in between
+   // cannot resurrect the forked-out records on the next start (replay applies the ROLLBACK in order).
+   journal_append(encode_rollback_body(block_num));
```
huangminghuang: This calls journal_append(encode_rollback_body(...)) but ignores its return value and still erases the in-memory overlay. If that journal append fails, or journaling was already disabled, the old PUT remains in abi_log.journal without the matching ROLLBACK. A later restart replays that stale PUT and brings back an ABI from a forked-out block, which is exactly the wrong-decode case this journal is meant to prevent. Rollback persistence needs fail-closed behavior: either record the rollback durably before erasing, or invalidate/reset/remove the stale journal so restart degrades to missing ABI rather than resurrecting forked data.

heifner: Good catch — fixed in 959f40e113.

Right on both the bug and the asymmetry: a lost PUT/flush is fail-safe (decodes to raw hex), but a lost ROLLBACK is fail-open — the orphaned PUT replays on restart and resurrects a forked-out ABI.

A trace_api node exists to record history, so I made **all** abi_log write failures fatal rather than recover-and-continue — uniformly across the journal *and* the main log (a failed flush previously just logged and retried, leaving the node up but not recording). They route to a clean `controller_emit_signal_exception` shutdown via the existing extraction handlers (`store_block_trace` for captures, `store_lib` for flushes); `on_block_start` now wraps the rollback the same way `check_continuity` does.

- No data is lost on the storage error: an unflushed record stays in the reversible journal, so a restart with storage free restores it and re-persists it on the next LIB advance.
- A failed rollback first rewrites the journal to the post-rollback overlay (`compact_journal`) — dropping only the orphaned forked records, keeping the canonical ones — so a restart neither resurrects forked data nor loses the still-reversible window. If even that rewrite fails (disk full), it truncates as a safe last resort.

Tests (test-only journal-write / compaction / main-log failpoints): append-, rollback- and flush-write-failure all abort; the rollback case preserves canonical records across a restart, the flush case recovers losslessly, and the truncate fallback fires when the rewrite also fails. 159/159.

huangminghuang (approved):

PR #296, opened 2026-04-15
Preserve y-parity when parsing compressed EM public keys

heifner:
## Summary
- `libfc`'s `to_em_public_key`/`trim_public_key` was silently normalizing compressed EM public keys with a `0x03` (odd y) prefix to `0x02` (even y) during hex parsing, flipping the key to a different curve point. Any address derived from the corrupted key no longer matched the real signer, breaking authex `createlink`, OPERATORS attestations, and any downstream address check for roughly half of all randomly-generated keys.
- Fix: preserve the SEC1 prefix byte for 33-byte compressed input; only strip the fixed `0x04` from 130-char uncompressed input.
- Also rejects bare 32-byte x-coordinate hex (64 chars): y-parity is not recoverable from x alone, and the previous "default to 0x02" behavior was a silent footgun. Callers must now supply 66-char compressed, 128-char x||y, or 130-char SEC1 uncompressed.

### DIFF
--- a/libraries/libfc/include/fc/crypto/ethereum/ethereum_utils.hpp
+++ b/libraries/libfc/include/fc/crypto/ethereum/ethereum_utils.hpp
@@ -19,9 +19,15 @@ namespace fc::em {
 namespace fc::crypto::ethereum {
 
 /**
- * Public key lengths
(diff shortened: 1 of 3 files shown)


jglanz (approved):
ROCKSTAR

PR #297, opened 2026-04-15
Add clio convert name subcommand for sysio::name <-> uint64_t

heifner:
## Summary
Adds `clio convert name <input>` which prints both interpretations of
the input: as a `sysio::name` (showing the uint64_t value) and as a
uint64_t (showing the decoded `sysio::name`). Inputs valid under both
forms show both; invalid interpretations are silently skipped.

## Examples

String input (valid name, not a uint64_t):
```
$ clio convert name sysio
As sysio::name : "sysio" -> uint64_t: 14389258095169634304
```

Numeric uint64_t (too long to parse as a sysio::name):
```
$ clio convert name 14389258095169634304
As uint64_t    : 14389258095169634304 -> sysio::name: "sysio"
```

Input valid under both interpretations:
```
$ clio convert name 12345
As sysio::name : "12345" -> uint64_t: 614251516705898496
As uint64_t    : 12345 -> sysio::name: "..........s3d"
```

Hex uint64_t (not a valid name):
```
$ clio convert name 0xdeadbeef
As uint64_t    : 3735928559 -> sysio::name: "......aypqzij"
```

Neither interpretation valid:
```
$ clio convert name INVALID
ERROR: Input is neither a valid sysio::name nor a uint64_t
```

### DIFF
--- a/programs/clio/main.cpp
+++ b/programs/clio/main.cpp
@@ -2095,7 +2095,7 @@ int main( int argc, char** argv ) {
          std::cout << localized("As sysio::name : \"${name}\" -> uint64_t: ${value}",
                                 ("name", n.to_string())("value", n.to_uint64_t())) << std::endl;
          any_success = true;
-      } catch (const std::exception&) {}
+      } catch (const fc::exception&) {}
       try {
          size_t pos = 0;
          uint64_t v = std::stoull(name_input, &pos, 0);
@@ -2105,9 +2105,11 @@ int main( int argc, char** argv ) {
          std::cout << localized("As uint64_t    : ${value} -> sysio::name: \"${name}\"",
                                 ("value", v)("name", n.to_string())) << std::endl;
          any_success = true;
-      } catch (const std::exception&) {}
+      } catch (const std::invalid_argument&) {
+      } catch (const std::out_of_range&) {}
       if (!any_success) {
-         std::cerr << "ERROR: Input is neither a valid sysio::name nor a uint64_t" << std::endl;
+         std::cerr << localized("ERROR: Input is neither a valid sysio::name nor a uint64_t") << std::endl;
+         FC_THROW_EXCEPTION(explained_exception, "invalid name input");
       }
    });
 
--- a/programs/clio/main.cpp
+++ b/programs/clio/main.cpp
@@ -2085,6 +2085,32 @@ int main( int argc, char** argv ) {
       std::cout << localized("Public key: ${key}", ("key", pubk.to_string({}, true) ) ) << std::endl;
    });
 
+   string name_input;
+   auto name_cmd = convert_cmd->add_subcommand("name", localized("Convert between sysio::name and uint64_t, printing both interpretations"));
+   name_cmd->add_option("input", name_input, localized("A sysio name or uint64_t value (decimal, or 0x-prefixed hex)"))->required();
+   name_cmd->callback([&name_input] {
+      bool any_success = false;
+      try {
+         name n{name_input};
+         std::cout << localized("As sysio::name : \"${name}\" -> uint64_t: ${value}",
+                                ("name", n.to_string())("value", n.to_uint64_t())) << std::endl;
+         any_success = true;
+      } catch (const std::exception&) {}
+      try {
+         size_t pos = 0;
+         uint64_t v = std::stoull(name_input, &pos, 0);
+         if (pos != name_input.size())
+            throw std::invalid_argument("trailing characters after number");
+         name n{v};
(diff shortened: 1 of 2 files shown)


dtaghavi (commented):
## Code Review

### Overview
This PR adds a `clio convert name <input>` subcommand that bidirectionally converts between `sysio::name` and `uint64_t`. The dual-interpretation approach (try both, show whichever succeed) is a sensible UX choice that avoids forcing users to know which direction they need. The implementation is concise at 26 lines.

---

### Code Quality & Style

**Positives:**
- Consistent use of `localized()` for user-facing output strings
- Nice alignment of output labels (`"As sysio::name : "` vs `"As uint64_t    : "`) for readability
- Correct use of `std::stoull` with the `pos` check to reject trailing characters (e.g., `"123abc"` correctly fails)
- `base = 0` in `stoull` properly handles both decimal and `0x`-prefixed hex inputs

---

### Issues & Suggestions

**1. Error message is not localized**
```cpp
// Current
std::cerr << "ERROR: Input is neither a valid sysio::name nor a uint64_t" << std::endl;

// Suggested
std::cerr << localized("ERROR: Input is neither a valid sysio::name nor a uint64_t") << std::endl;
```
All other output in this callback goes through `localized()`. This one inconsistently bypasses it.

**2. No non-zero exit code on failure**
When neither interpretation is valid, the error is printed to stderr but the process exits with `0`. Other CLI tools (and most `clio` error paths) exit non-zero on user error, which matters for scripting.
```cpp
if (!any_success) {
    std::cerr << localized("ERROR: Input is neither a valid sysio::name nor a uint64_t") << std::endl;
    return 1; // or throw an appropriate sysio exception
}
```

**3. Overly broad silent exception swallowing**
Both `catch (const std::exception&) {}` blocks swallow all standard exceptions silently, including things like `std::bad_alloc`. Consider catching only the specific parse/conversion exceptions expected here (e.g., `fc::exception` or `std::invalid_argument` / `std::out_of_range`), and re-throwing anything unexpected.

**4. No tests**
No unit or integration tests are included. Given this is a pure conversion utility with well-defined input/output cases (the PR description even lists them), these examples would make excellent test cases. At minimum, the four cases from the PR description (string name, numeric uint64, ambiguous input, invalid input) should be covered.

---

### Minor Nits
- The output for the `uint64_t` path outputs the *decoded* uint64 value (`v`) while the `name` path outputs the string representation — the asymmetry is intentional and correct, but a brief inline comment might help future readers.
- No issues with the lambda capture by reference (`[&name_input]`) — `name_input` outlives the lambda.

---

### Summary

| | |
|---|---|
| Correctness | Mostly correct; exit code on error is missing |
| Style | Good, one localization inconsistency |
| Tests | None added |
| Risk | Low — additive only, no existing behavior changed |

Overall this is a clean, useful addition. The main ask before merging: fix the exit code on failure, localize the error string, and add tests for the four cases documented in the PR description.

dtaghavi (commented):
## Follow-up Review (after updates)

All four items from the previous review have been addressed. Nice turnaround.

---

### What Changed

| Issue | Status |
|---|---|
| Error message not localized | ✅ Fixed — `localized()` added |
| No non-zero exit on failure | ✅ Fixed — `FC_THROW_EXCEPTION(explained_exception, ...)` used |
| Overly broad exception swallowing | ✅ Fixed — narrowed to `fc::exception` (name path) and `std::invalid_argument` / `std::out_of_range` (stoull path) |
| No tests | ✅ Fixed — `clio_convert_name_test()` covers all four cases |

---

### One Remaining Nit

**Inaccurate comment in the test:**
```python
# Valid sysio::name only ("sysio" is 5 lowercase chars; decimal value is 20 digits, > uint64_t max)
```
The comment says the uint64_t value `14389258095169634304` is `> uint64_t max` — but it isn't: `14389258095169634304 < 18446744073709551615`. The real reason `"sysio"` doesn't match the `uint64_t` path is simply that `std::stoull("sysio")` throws `std::invalid_argument` — the string isn't a number at all. The comment should say something like:

```python
# Valid sysio::name only ("sysio" contains letters, so stoull throws std::invalid_argument)
```

---

### Everything Else Looks Good

- `FC_THROW_EXCEPTION(explained_exception, ...)` is the right idiom here — it integrates with `clio`'s top-level exception handling so the non-zero exit is consistent with the rest of the tool.
- The exception narrowing is correct: `name{v}` (construction from `uint64_t`) is a no-throw operation, so the second catch block only needs to handle `stoull` failures.
- Test assertions are appropriately precise and cover the documented examples exactly.

This is ready to merge once the comment is corrected.

dtaghavi (approved):

PR #298, opened 2026-04-16
PR #111 review: follow-up fixes and cleanups

heifner:
## Summary

Follow-up cleanups from a post-merge review of PR #111 "Bill each action for CPU/NET". Five commits grouped by concern:

1. **`Fix subjective_billing overlap and add coverage tests`** — real bug where an account in both `accounts_billing` and `auth_cpu` maps had its per-trx entry overwritten (not summed), leaking value into the cache on removal. Also fixes underflow guards, `_reset()` missing a field, dead overload removal, and adds 8 new test cases covering the previously-uncovered public API.
2. **`Fix SHiP net_usage_words emission`** — PR #111 emitted bytes×8 instead of words; external SHiP consumers would see 8x inflated values. The test had been updated to lock in the bug.
3. **`Include data varint prefix in action billable size`** — `get_billable_size` under-billed by 1-5 bytes per action; fixed by using `pack_size(*this)`. Test constants updated and reference data regenerated.
4. **`Harden transaction billing correctness`** — uint32 overflow guards on the cpu_usage_us vector sum, payer-at-position-0 defense-in-depth in `validate_referenced_accounts`, gate implicit trxs out of subjective billing, remove dead `leeway_trx_net_limit`, delegate duplicated `get_cpu_limit`, golden-value test for receipt digest format, misc stale-comment fixes.
5. **`PR #111 review cleanups`** — non-behavioral: dead-include removal, stale comments, whitespace, `is_onblock` dedupe, unused `eosio_system_tester.hpp` deletion, CFD discount values in genesis/launcher corrected to match code defaults, producer plugin thread-assert + retry cap restored.


unittests/subjective_billing_tests.cpp
```diff
@@ -263,6 +263,269 @@ BOOST_AUTO_TEST_CASE( subjective_bill_multiple_accounts_test ) {
...
+      BOOST_CHECK_EQUAL( 10 + 5, sub_bill.get_subjective_bill(a, now).count() );
+
+      // Simulate id1 landing in a block.
+      sub_bill.remove_subjective_billing( id1, 0 );
+
+      // Full contribution must be cleared; with the overlap bug only 5 would be
+      // subtracted, leaving 10 stuck in the cache.
+      BOOST_CHECK_EQUAL( 0, sub_bill.get_subjective_bill(a, now).count() );
+   }
+   {  // Same check, but with auth_cpu listed first — verify order-independence.
```
huangminghuang: I don't see the difference of this block and the previous one, it doesn't fit the description in the comment. You might need to adjust the test or remove this block entirely.

huangminghuang (changes requested):

unittests/subjective_billing_tests.cpp
heifner: You are correct. Removed the second one. Also added a few more tests.

huangminghuang (approved):

huangminghuang (approved):

PR #299, opened 2026-04-17
Harden block validation and snapshot load paths

heifner:
## Summary
- Close block-2 validation gap: remove early-return in `verify_basic_proper_block_invariants`, use `core.latest_qc_claim()` as authoritative QC-claim reference
- Factor `finalizer_policy::validate()` / `proposer_policy::validate()` as shared structural-validation methods reused by intrinsics and snapshot loading
- Extract `validate_s_root_extensions_match()` helper from inline 39-line nested-lambda loop in `apply_block`
- Add duplicate-detection in `check_protocol_features` for within-block activation list
- Reorder `qc_t::verify_basic` so pending `verify_vote_format` runs before `verify_dual_finalizers_votes`
- Snapshot null-pointer guards run before member-init-list dereferences
- 40+ new unit tests (finalizer/producer intrinsic negatives, merkle async-determinism, s_root_extension matching, signature-recovery dedup, snapshot invariant checks)


libraries/chain/block_state.cpp
```diff
@@ -184,29 +184,84 @@ block_state::block_state(snapshot_detail::snapshot_block_state_v1&& sbs)
...
+   if (pending_finalizer_policy) {
+      SYS_ASSERT(pending_finalizer_policy->second, snapshot_exception,
+                 "pending_finalizer_policy must not contain a null policy");
+   }
+   for (const auto& [_, pol] : proposed_finalizer_policies) {
+      SYS_ASSERT(pol, snapshot_exception, "proposed_finalizer_policies entry must not be null");
+   }
+
+   // Now safe to compute state that depends on active_finalizer_policy.
+   strong_digest  = compute_finality_digest();
```
huangminghuang, line 245: This computes strong_digest from snapshot state before validating core at block_state.cpp (line 224), and before checking activated_protocol_features. compute_finality_digest() can reach core.latest_qc_claim() and compute_base_digest(), which asserts then dereferences activated_protocol_features at block_header_state.cpp (line 32). A malformed snapshot with a null activated_protocol_features pointer or invalid/empty finality core can still assert/segfault before the intended snapshot_exception. Move core.validate_snapshot() and an explicit activated_protocol_features null check before digest computation, and add regression coverage for both cases.

heifner: Fixed: https://github.com/Wire-Network/wire-sysio/pull/299/commits/da3daa7525a4b0bcac39f497973dd287eb801628

libraries/chain/block_state.cpp
```diff
@@ -184,29 +184,89 @@ block_state::block_state(snapshot_detail::snapshot_block_state_v1&& sbs)
...
+   }
+   validate_prop_pol(*active_proposer_policy, "active");
+   if (latest_proposed_proposer_policy) {
+      validate_prop_pol(*latest_proposed_proposer_policy, "latest_proposed");
+   }
+   if (latest_pending_proposer_policy) {
+      validate_prop_pol(*latest_pending_proposer_policy, "latest_pending");
    }
 
    // Snapshot hardening: validate valid_t
```
huangminghuang: This only validates valid_t when valid is present, but it never rejects valid == std::nullopt for a non-genesis snapshot. The new comment in block_state.cpp (line 365) says null valid is legitimate only for the Savanna genesis core, yet the snapshot constructor accepts a tampered non-genesis snapshot with sbs.valid.reset(). In release builds, get_validation_mroot() then returns an empty digest after an assert(core.is_genesis_core()), so the malformed snapshot is detected too late or only indirectly. Add a constructor check like SYS_ASSERT(valid || core.is_genesis_core(), snapshot_exception, ...), plus a regression test.

heifner: Fixed: https://github.com/Wire-Network/wire-sysio/pull/299/commits/b8c9cbe7365392ccfe12d3b4a58f2317fc56368a

libraries/chain/block_state.cpp
```diff
@@ -184,39 +184,102 @@ block_state::block_state(snapshot_detail::snapshot_block_state_v1&& sbs)
...
+      pending_finalizer_policy ? pending_finalizer_policy->second : finalizer_policy_ptr{}
+   };
+
+   header_exts = header.validate_and_extract_header_extensions();
+
+   // core.latest_qc_claim() is the authoritative reference for downstream QC-claim checks
+   // (see verify_basic_proper_block_invariants). For every non-genesis block it equals
+   // header.qc_claim by construction (finality_core::next sets latest_qc_claim from the
+   // header's claim). Genesis intentionally differs: core is {1, false}, header is {0, false}.
+   SYS_ASSERT(core.is_genesis_core() || core.latest_qc_claim() == header.qc_claim, snapshot_exception,
```
huangminghuang: This lets any refs.empty() core bypass the core.latest_qc_claim() == header.qc_claim check. A tampered non-genesis snapshot can replace the head’s core with a one-link “genesis core” for that block number, trim valid.validation_mroots to the expected size, and avoid the new QC/header consistency check. The exemption should be tied to the actual genesis block, not just core.is_genesis_core().

libraries/chain/block_state.cpp
huangminghuang, line 172: This still trusts sbs.block_id without checking it equals header.calculate_id(). id() and fork-db indexing then use that stored value as authoritative, so a malformed snapshot can load a block_state whose identity does not match its header. The constructor should reject block_id != header.calculate_id() before exposing the state as chain_head.

heifner: Fixed: https://github.com/Wire-Network/wire-sysio/pull/299/commits/324435f612

libraries/chain/block_state.cpp
heifner: Fixed: https://github.com/Wire-Network/wire-sysio/pull/299/commits/324435f612 -- dropped the `is_genesis_core()` exemption entirely and added a guard that rejects `block_num() == 1`. Snapshots of the genesis block aren't a legitimate input (genesis state is reconstructed from `genesis_state` via `initialize_blockchain_state`), so the underlying refs.empty() bypass surface is gone rather than narrowed.

libraries/chain/block_state.cpp
```diff
@@ -184,39 +184,126 @@ block_state::block_state(snapshot_detail::snapshot_block_state_v1&& sbs)
...
    for (const auto& [_, pol] : proposed_finalizer_policies) {
-      pol->validate_snapshot();
+      SYS_ASSERT(pol, snapshot_exception, "proposed_finalizer_policies entry must not be null");
+   }
+
+   // Snapshot hardening: validate finality_core invariants BEFORE computing digests.
+   // compute_finality_digest() calls core.latest_qc_claim() (asserts !links.empty()) and
+   // core.get_block_reference() (asserts bounds on refs); an empty or malformed core
+   // would assert/segfault inside the digest path before we could throw snapshot_exception.
+   core.validate_snapshot();
```
huangminghuang, line 242: This still accepts a snapshot whose finality_core belongs to a different block number than the snapshot header. The new checks validate the core’s internal invariants, block_id == header.calculate_id(), and core.latest_qc_claim() == header.qc_claim, but nothing asserts core.current_block_num() == block_num(). A tampered snapshot can therefore pair a block-N header/id with a structurally valid core for block M and a matching valid_t size. Later code assumes that relationship, e.g. finality_core::next() documents current_block.block_num() == this->current_block_num() and only enforces it with assert at

heifner: Fixed: https://github.com/Wire-Network/wire-sysio/pull/299/commits/81062a081b

huangminghuang (approved):

PR #302, opened 2026-04-19
PerformanceHarness: use trace_api/get_actions for post-test trx extraction

heifner:
## Summary

Migrates the PerformanceHarness post-test block/trx extraction from `trace_api/get_block` to `trace_api/get_actions` + `chain/get_block_info`, exercising the new endpoint introduced in this PR and removing the base58-signature serialization that dominated harness CPU.

## Why

`queryBlockTrxData` iterates every block in the test range. The old call path - `trace_api/get_block` - serialized the entire block including every signature (base58 of each signature does BigNum divide in OpenSSL; heaptrack showed ~5.7M `OPENSSL_malloc` calls on a 5k TPS 30s run). The harness only reads: `trx_id`, `block_num`, `block_time`, `cpu_usage_us`, `net_usage_words`, and the first action's name (to skip onblock). Signatures are never consulted.

`get_actions` returns exactly the fields needed, filtered server-side by action name so onblock trxs are dropped without any base58 work. `chain/get_block_info` covers the block-header fields (id, producer, timestamp).

## Changes

- `tests/PerformanceHarness/performance_test_basic.py`: `queryBlockTrxData` now issues two lightweight calls per block (`chain/get_block_info` + `trace_api/get_actions`) instead of the heavy `trace_api/get_block`. The action-name filter is derived from `userTrxDataDict.actions[0].actionName`, so all test variants (transfer, cpu, ram, net, newaccount, doit, doitslow) work automatically. When no user data is configured, falls back to an explicit onblock skip on the client side. Block finality (`status` column in `blockData.txt`) is read from the per-action `block_status` field added in PR #295; when the action filter drops every entry in a block (typically the leading/trailing ramp window), falls back to `trace_api/get_block` for that block's status -- still single-source-of-truth from trace_api's data log rather than mixing in `chain/get_info` LIB.

## Measurement

On a 50k-TPS 90s test (identical config, same combined-perf branch):

| Harness                        | Total runtime | Post-test overhead |
|--------------------------------|---------------|---------------------|
| old `trace_api/get_block`      | 552 s         | 462 s               |
| new `trace_api/get_actions`    | ~366 s (90s extrapolated) | ~276 s |

~40% less time spent in post-test extraction+analysis.

Targets `feature/trace-api-history` so it rolls into PR #295.


tests/PerformanceHarness/performance_test_basic.py
```diff
@@ -305,24 +305,50 @@ def isOnBlockTransaction(self, transaction):
...
+                seen = set()  # dedup if a trx somehow has multiple matching actions
+                for action in actionsResp['payload']['actions']:
+                    # If no action filter configured, still skip onblock explicitly.
+                    if not actionFilter and action.get('account') == 'sysio' and action.get('name') == 'onblock':
+                        continue
+                    trxId = action['trx_id']
+                    if trxId in seen:
+                        continue
+                    seen.add(trxId)
+                    cpu = action.get('cpu_usage_us', 0) or 0
```
huangminghuang: This now records per-action cpu_usage_us / net_usage, but the old code recorded per-transaction cpu_usage_us / net_usage_words from trace_api/get_block. That changes both scope and units: net_usage is bytes on the action trace, while the previous net_usage_words was ceil(transaction.net_usage / 8). The harness’ block size and transaction net stats will no longer be comparable, and multi-action transactions will undercount CPU/net by taking only the first matching action. I’d either add transaction-level resource fields to get_actions, or keep a lightweight transaction summary endpoint for this harness path.

heifner: Good catch. Went with option 1 - added `trx_cpu_usage_us` and `trx_net_usage_words` to each `get_actions` action variant in PR #295 (`feature/trace-api-history`) at 61906f88e9, then gated them to the full shape only at 22880d28c0 so `get_token_transfers` (slim) keeps its "no resource fields" intent. This branch consumes the new fields at c12471d559, so multi-action trxs are correctly attributed and net is back in words to match the previous `trace_api/get_block` shape.

tests/PerformanceHarness/performance_test_basic.py
```diff
@@ -305,24 +305,54 @@ def isOnBlockTransaction(self, transaction):
...
+                    blockCpuTotal += cpu
+                    blockNetTotal += net
+                    blockTransactionTotal += 1
+            block_data = blockData(blockId=blockInfo["payload"]["id"],
+                                   blockNum=blockInfo['payload']['block_num'],
                                    transactions=blockTransactionTotal, net=blockNetTotal, cpu=blockCpuTotal,
-                                   producer=block["payload"]["producer"], status=block["payload"]["status"],
-                                   _timestamp=block["payload"]["timestamp"])
+                                   producer=blockInfo["payload"]["producer"],
+                                   status="executed",
```
huangminghuang: This line now hardcodes every block status as "executed". The previous trace_api/get_block path wrote the block status from trace API ("irreversible" or "pending"), and blockData.txt still has a status column. This silently changes the artifact semantics, and "executed" is a transaction status, not a block finality status. If downstream harness/report tooling reads that field, it loses finality information. I’d either preserve the old meaning by deriving pending/irreversible from chain/get_info/LIB, or make the artifact schema change explicit.

heifner: Good catch -- went with option 1's intent but kept the source consistent with trace_api rather than crossing the correlation boundary to chain/get_info.

Server side (feature/trace-api-history @ 200ec2dfe4): get_actions and get_token_transfers now emit `block_status` ("irreversible" / "pending") on every action, sourced from the same data-log tuple that get_block uses for its `status` field. Slim shape carries it too. Doc notes that operators wanting only-irreversible responses can run nodeop with `read-mode = irreversible`.

Harness (this branch @ 09b31cefcb): reads `block_status` from the first action of each block (all actions in a block share it). When the action filter drops every entry in a block (typically the leading/trailing ramp window), falls back to trace_api/get_block for that block's status -- still a single trace_api source of truth. blockData.txt schema unchanged, every row has a real finality value.

huangminghuang (changes requested):
performance_test_basic.py (line 367) now passes chain/get_block_info’s timestamp into blockData, but log_reader.py (line 146) always strips the last character because the old trace_api/get_block timestamp ended in Z. get_block_info serializes block_timestamp_type via fc::time_point::to_iso_string() without Z, so blockData.txt will now write timestamps like .50 instead of .500. The harness’ latency math likely survives for 0/500ms blocks, but the artifact format loses precision/shape and can surprise downstream parsers. Normalize by appending Z, or make blockData.timestamp strip only a trailing Z.

tests/PerformanceHarness/performance_test_basic.py
```diff
@@ -305,24 +305,71 @@ def isOnBlockTransaction(self, transaction):
...
+        # Block finality (irreversible/pending) comes from block_status on each action -- trace_api
+        # is the single source of truth so we never mix in a chain/get_info LIB read that could
+        # disagree with what the trace data reflects. For blocks where the action filter dropped
+        # everything (typically the leading/trailing ramp window), fall back to trace_api/get_block
+        # for that one block: empty blocks carry no harness-action trxs so the payload is small.
+        actionFilter = None
+        if getattr(self, 'userTrxDataDict', None):
+            cfgActions = self.userTrxDataDict.get('actions') or []
+            if cfgActions:
+                actionFilter = cfgActions[0].get('actionName')
```
huangminghuang: This only queries trace_api/get_actions for cfgActions[0]['actionName']. userTrxDataDict['actions'] is a list and the whole list is passed to the transaction generator, so a config with multiple action types can successfully send transactions for later actions while post-test extraction never retrieves them. Those transactions then stay out of data.trxDict and get reported as missing/dropped or undercounted in CPU/net/TPS. The extraction should query each distinct configured action name and union/dedup by trx_id, or avoid action filtering when multiple action names are configured.

heifner: Good catch — fixed in `bc449eb994`. The extraction no longer assumes the first configured action appears in every transaction.

`get_actions` filters by a single action name server-side, so I push the filter down only when exactly one distinct `actionName` is configured (keeps the slim single-action payload, onblock dropped server-side). With multiple distinct names I drop the server-side filter and skip just onblock client-side — one request per block, and every trx is captured regardless of which configured action it carries, deduped by `trx_id` (a multi-action trx is still recorded once with its parent-transaction cpu/net totals).

I went with "avoid filtering when multiple action names are configured" rather than per-name union queries because `trx_generator::generate_actions` puts every configured action into a single transaction, so per-name queries would issue N near-identical round-trips per block — against the point of this change. The no-filter path is the same one the default (no-user-data) transfer test already exercises.

Added `performance_test_basic_query_test` — a fast, cluster-free `add_p_test` that drives `queryBlockTrxData` against a fake node emulating server-side action filtering. It covers the multi-action, single-action, dedup-by-trx-id, default, and status-fallback paths; the multi-action case fails against the old `cfgActions[0]` logic.

huangminghuang (approved):

PR #304, opened 2026-04-20
kv: secondary rows reference primary by chainbase id

heifner:
## Summary

Secondary index rows (`kv_index_object`) now store an 8-byte `primary_id` (chainbase id of the referenced `kv_object`) instead of a copy of the primary-key bytes. For contracts with long primary keys (up to 256 B, possibly larger with composite keys), this saves `pri_key.size() + ~16 B` per secondary row in both chainbase and RAM billing. For short (8-byte) primary keys it's neutral.

Requires coordinated wire-cdt change: Wire-Network/wire-cdt#50

## Rollout

Wire is pre-launch: no live `kv_multi_index` sec-index data exists on any chain. The composite-key tiebreaker change (pri_key byte-lex → primary_id insertion order) becomes the baseline behavior for launch and does **not** require a protocol feature gate. No migration or state replay required.

## Behavior changes

- **Secondary-index composite tiebreaker** moves from `pri_key` byte-lex order to `primary_id` (chainbase insertion order). Within duplicate `sec_keys`, iteration order follows the order rows were inserted rather than primary-key lexicographic order.
- **`kv_set` / `kv_erase` return value** is now the primary row's chainbase id (was: billable RAM delta, which CDT didn't use).
- **`kv_idx_store` / `kv_idx_remove` / `kv_idx_update` signature** takes `int64_t primary_id` in place of `(pri_key, pri_key_size)`. Callers thread the id from the preceding `kv_set` / `kv_erase`.
- **`kv_it_value`** accepts secondary iterator handles; resolves the value via the slot's cached `primary_id` (a by_id lookup), faster than the old by_code_key path for sec-iter value reads.
- **`kv_idx_primary_key`** materializes `pri_key` bytes lazily from the primary `kv_object` on demand, caching in the iterator slot until the next position change.
- **`kv_idx_lower_bound`** no longer does a double-scan on miss — it peeks backward from the returned iterator to detect table-non-empty.

## Snapshot / chainbase

- New `chainbase::undo_index::emplace_with_id(id, ctor)` constructs an object at a caller-supplied id and advances `_next_id` past it. Used exclusively by the snapshot loader so primary_id references stay valid across save+load. Asserts `_undo_stack.empty()` at call time (developer-error guard).
- `snapshot_kv_object` / `snapshot_kv_index_object` now carry `id`; the loader reinstates each row at its original id, so cross-row references don't need any post-load resolution pass.
- `_next_id` increment guards against signed overflow UB at the `INT64_MAX` boundary.

## Plugins / SHiP / tools

- `chain_plugin::get_table_rows` secondary scans resolve primary rows via `db.find<kv_object>(sec->primary_id)`. Asserts the reference exists (surfaces a state-inconsistency to the RPC caller rather than silently returning an empty row).
- SHiP continues to emit `pri_key` bytes in its `kv_index_object` stream (wire format unchanged for external consumers); the bytes are materialized from the primary at serialize time.

## Tests

- New `kv_comprehensive_tests/kv_secondary_tiebreak_is_primary_id_insertion_order` pins the composite-key tiebreaker (inserts pri_keys in non-lex order and asserts insertion-order iteration).
- New `kv_api_tests/idx_value_on_secondary_handle` + `test_kv_api/testsecvalue` pin the raw-capi contract that `kv_it_value` accepts sec handles.
- `kv_tests`, `kv_comprehensive_tests`, `kv_api_tests` updated for primary_id semantics throughout.
- RAM-sensitive `protocol_feature_tests` retuned (ROA bump 0.0986 → 0.0968 SYS) to match the smaller recompiled `ram_restrictions_test` WASM.
- Reference data regenerated: `deep-mind.log`, `unittests/snapshots/*`, `unittests/test-data/consensus_blockchain/*`.
- Production and pre-built test contract WASMs recompiled against the new CDT.

## Docs

- `docs/kv-ram-billing.md` updated for the new `kv_index_object` layout and per-row cost tables.
- `docs/kv-security-audit.md` describes the new erase ordering (`kv_erase` first → `primary_id` → `kv_idx_remove` for each sec).


libraries/state_history/include/sysio/state_history/serialization.hpp
```diff
@@ -255,9 +255,15 @@ datastream<ST>& operator<<(datastream<ST>& ds, const history_serial_wrapper<sysi
...
    if (obj.obj.sec_key.size() > 0)
       ds.write(obj.obj.sec_key.data(), obj.obj.sec_key.size());
-   history_pack_varuint64(ds, obj.obj.pri_key.size());
-   if (obj.obj.pri_key.size() > 0)
-      ds.write(obj.obj.pri_key.data(), obj.obj.pri_key.size());
+   // Resolve primary_id -> primary kv_object and emit its key bytes so SHiP
+   // consumers continue to see pri_key bytes in the stream. The wire format
+   // for secondary rows is unchanged (struct_version 0 still ends with
+   // a length-prefixed pri_key byte string).
+   const auto* primary = obj.db.find<sysio::chain::kv_object>(obj.obj.primary_id);
```
huangminghuang, line 270: This line can no longer serialize removed KV secondary rows correctly. The PR says SHiP keeps emitting pri_key bytes by resolving primary_id -> kv_object, but KV erase flow now deletes the primary first (apply_context.cpp:713-715) and then removes secondary rows (apply_context.cpp:1019-1021). create_deltas.cpp:95-97 serializes removed rows after both removals, so the primary lookup returns null and SHiP emits pri_key_size = 0. That breaks the unchanged SHiP wire contract for secondary-row deletes and makes indexers unable to identify the removed primary key.

libraries/chain/include/sysio/chain/kv_table_objects.hpp
```diff
@@ -214,5 +223,5 @@ namespace config {
 CHAINBASE_SET_INDEX_TYPE(sysio::chain::kv_object, sysio::chain::kv_index)
 CHAINBASE_SET_INDEX_TYPE(sysio::chain::kv_index_object, sysio::chain::kv_index_index)
 
-FC_REFLECT(sysio::chain::snapshot_kv_object, (code)(payer)(key)(value)(table_id))
-FC_REFLECT(sysio::chain::snapshot_kv_index_object, (code)(payer)(sec_key)(pri_key)(table_id))
+FC_REFLECT(sysio::chain::snapshot_kv_object, (id)(code)(payer)(key)(value)(table_id))
+FC_REFLECT(sysio::chain::snapshot_kv_index_object, (id)(code)(payer)(sec_key)(primary_id)(table_id))
```
huangminghuang, line 227: This changes the binary snapshot DTO layout by prepending id fields, and controller.cpp:1583-1585 unconditionally reads those ids. There is no legacy reader/version gate for snapshots written with the old (code,payer,key,...) and (code,payer,sec_key,pri_key,...) layouts, so older snapshots will be misdecoded rather than migrated. If old snapshots are intentionally unsupported, this needs an explicit snapshot format/version bump and tests that assert the failure mode.

heifner: No existing snapshots to support. Chain has not launched.

libraries/state_history/include/sysio/state_history/serialization.hpp
heifner: Good catch, confirmed. Fixed in 26478eaf61 by falling back to a scan of the kv_index undo session's removed_values when the live find() misses. SHiP packs deltas at block scope and pack_deltas iterates kv_index removed_values too, so the primary's row (with its key bytes) is guaranteed to be there for the same block. Wire format unchanged.

Added test_deltas_contract_index_kv_secondary_delete_pri_key (state_history_tests.cpp) which deploys get_table_test, addnumobj + erase in separate blocks, then asserts every secondary delete delta carries non-empty pri_key matching the captured pre-delete bytes.

Linear scan is O(N) per fallback (N = primaries removed in this block), only fires on the cold path (live primary missing). For typical erase patterns (a few rows per block) the scan is in the noise; if bulk-delete workloads ever push on this, can pre-build a `flat_hash_map<id, blob*>` in create_deltas as a follow-up.

libraries/chaindb/include/chainbase/undo_index.hpp
```diff
@@ -375,6 +375,42 @@ namespace chainbase {
...
+         auto guard1 = scope_exit{[&]{ alloc_traits::destroy(_allocator, &*p); }};
+         if(!insert_impl<1>(p->_item))
+            BOOST_THROW_EXCEPTION( std::logic_error{ "could not insert object, most likely a duplicate id was supplied" } );
+         std::get<0>(_indices).push_back(p->_item);
+         on_create(p->_item);
+         // Advance the id counter. Guard against signed overflow (UB in C++)
+         // at the saturation boundary; reaching INT64_MAX in practice would
+         // require ~585 years at 1B emplaces/sec, but the explicit check
+         // keeps the arithmetic well-defined for every compiler.
+         if (_next_id <= id && id._id < std::numeric_limits<int64_t>::max())
```
brianjohnson5972: Can we not do `std::numeric_limits<id_type>::max()`? Also, it will overflow in the next emplace call? This seems like more of an assert check before calling `++_next_id`

PR #310, opened 2026-04-23
Test: add adversarial non-kv host intrinsic probes

heifner:
## Background

PR #308 (feature/kv-intrinsic-probe-tests) added adversarial probes for the 22 kv_* host intrinsics by driving the raw host ABI with inputs CDT's kv_multi_index / kv::table wrappers would never emit. This PR is the companion suite covering the remaining non-kv host intrinsics: the crypto family (sha1/sha256/sha512/ripemd160 + assert_*, recover_key/assert_recover_key), the 128-bit integer and float128 compiler_builtins, preactivate_feature, the privileged_check-gated resource / blockchain-parameters ops, and the console / action-data legacy_span readers.

Motivation is the forthcoming legacy_span/legacy_ptr -> span cleanup. That cleanup touches every intrinsic in the list above, so the invariants the current implementation relies on -- alignment-proxy copy-in/out paths for legacy_ptr<fc::sha*> / legacy_ptr<int64_t, 8> / legacy_ptr<__int128>, zero-length-span acceptance, "query required size with buffer=0" semantics, SYS_ASSERT-throws-vs-returns-sentinel contracts -- need to be pinned before the refactor lands.

## Layout

`unittests/test-contracts/intrinsic_probe/intrinsic_probe.cpp` declares every host intrinsic it probes via `extern "C" + __attribute__((sysio_wasm_import))`. The few that conflict with `sysio::internal_use_do_not_use` wrappers CDT already declares (prints, sysio_assert, send_inline, get_action, read_transaction, etc.) are reached through a `raw::` alias. One `[[sysio::action]]` per probe -- 98 probes total across sha family (16), assert_sha family (16), recover_key family (13), preactivate_feature (3), compiler_builtins int128 (12), compiler_builtins float128 (9), resource/auth/producer/blockchain (11), console/IO (18).

`unittests/intrinsic_probe_tests.cpp` deploys the contract to both a non-privileged account (`intprobe`) and a privileged account (`intprobe2`); each `BOOST_FIXTURE_TEST_CASE` routes to the appropriate account based on whether the intrinsic under test is `privileged_check`-gated. Shared `validating_tester` singleton avoids re-paying bios setup per case. Setup policy is `preactivate_feature_and_new_bios` so `set_privileged()` can route through `sysio.bios::setpriv` and `reserved_first_protocol_feature` remains unactivated for the `preactok` probe.

## Coverage

**Crypto** -- four probes per hash family (golden, empty input, 1KB input, unaligned out-ptr forcing `argument_proxy<fc::sha*, 8>` copy-out path), four per `assert_*` family (correct match, zero-hash mismatch rejection, empty input with empty hash golden, unaligned const-hash ptr forcing copy-in path).

**recover_key / assert_recover_key** -- each of the distinct failure paths is isolated: empty sig (fc::raw::unpack runs dry), short sig (truncation mid-shim), bad variant tag (fc::raw variant unpack out-of-range), bad recovery byte (elliptic_secp256k1.cpp `FC_THROW_EXCEPTION` "unable to reconstruct public key"), bad r/s (math either succeeds with a different pub or fails -- both acceptable, probe and driver handle both), K1 small-pub buffer (fc::datastream `FC_ASSERT` for fixed-size key types), unaligned digest (argument_proxy copy-in path). `assert_recover_key` adds the type-mismatch and recovered-pub-mismatch `SYS_ASSERT` paths separately.

**compiler_builtins** -- each intrinsic probed with golden, unaligned out-ptr (`argument_proxy<__int128*, 16>` / `argument_proxy<float128_t*, 16>` copy-out path), and edge values (U64 carry in `__multi3`, INT128 shift >= 128 in `__ashlti3`, divide-by-zero in `__div[u]ti3`, NaN in `__multf3`, 1.0/0.0 in `__divtf3`, 2^127 overflow in `__fixtfti`).

**Privileged gates** -- three probes per priv-gated family: the accepted priv path, the non-priv rejection (`unaccessible_api`), and either a body-level rejection (`preactivate_feature` with bogus digest) or an idempotent in-priv probe (`setreslim` reads current limits, bumps, restores so shared-tester state is unaffected).

**P2** -- `get_resource_limits` with aligned and unaligned int64_t out-ptr trios (the aligned-proxy copy-back path is the part most likely to regress under the cleanup), `get_blockchain_parameters_packed` with size=0 and with a buffer sized to the size query, `get_active_producers` with a sufficient buffer and with size=0, `set_*` probes for each family's non-priv rejection.

**P3** -- `prints` (null_terminated_ptr) and `prints_l` (legacy_span) both covered; `sysio_assert` vs `sysio_assert_message` both covered in firing and non-firing variants; `sysio_assert_message` with empty span pinned (not crashed); `read_action_data` with zero size pinned (returns 0, doesn't overflow); `read_transaction` and `get_blockchain_parameters_packed` both pin the "size=0 returns required bytes" contract; `get_action` with invalid type pinned to throw `action_not_found_exception` rather than return -1 (its return-(-1) path is only for valid-type / out-of-range-index); `get_context_free_data` pinned to throw `unaccessible_api` from a non-context-free apply (it is `REGISTER_LEGACY_CF_ONLY_HOST_FUNCTION`, so its `context_free_check` precondition fires before the body); `send_inline` with an empty span pinned to throw during action unpack (no silent no-op).

## Coverage boundary

The compiler_builtins probes pin the adversarially-distinct representatives of the int128 / float128 host surface (mul, signed and unsigned div, arithmetic-left-shift, f128 add/mul/div, float->int128 saturating conversions) across the legacy_ptr -> aligned_ptr copy-out refactor in #314, which keeps the intrinsics.

`__modti3`, `__umodti3`, `__ashrti3`, `__lshlti3`, `__lshrti3`, `__subtf3` are declared in the probe contract for ABI completeness but intentionally NOT separately probed. This PR is the base of a stack (#310 -> #314 -> #318). #318 deletes the host compiler_builtins outright (`compiler_builtins.cpp` + `libraries/builtins/`, all 43 registrations) and relocates equivalent coverage -- including the `INT128_MIN/-1` wrap and shift-count >= 128 saturation edges -- into a contract-side `librt` suite (`test_compiler_builtins.cpp` / `test_f128_builtins.cpp`); its diff against this file is a single -289-line hunk deleting the whole compiler_builtins probe section. Host-ABI probes for those entry points would pin behavior #318 removes two PRs up the same stack. This boundary is documented at the declaration block in `intrinsic_probe.cpp`.

## Host behaviors pinned for regression

- `assert_sha` family throws `crypto_api_exception` "hash mismatch" via SYS_ASSERT; not a generic `fc::exception`. The separate exception-cleanup follow-on noted below can tighten the other `recover_key` paths toward the same shape; this pin gives that refactor a baseline.
- `recover_key` for K1 with a pub buffer smaller than 34 bytes throws (fc::datastream fixed-size pack `FC_ASSERT`). Variable-size keys (ed, wa) silently truncate via memcpy -- an API inconsistency worth flagging to the follow-on cleanup but NOT a regression surface today (not reached with the K1 test vector this suite uses).
- `apply_context::get_action` with type not in {0, 1} throws `action_not_found_exception`. The -1 sentinel applies only to valid-type / out-of-range-index.
- `get_context_free_data` is context-free-only; from a non-context-free apply its `context_free_check` precondition throws `unaccessible_api` before the host body runs -- the same "gate fires before body" shape as `privileged_check`.
- `__divti3` / `__udivti3` throw `arithmetic_exception` on rhs==0 (SYS_ASSERT in compiler_builtins.cpp).
- `__ashlti3` with shift >= 128 returns 0 -- well-defined saturation, not UB.
- `__divtf3(1.0, 0.0)` returns +Inf, not a throw -- IEEE 754 semantics preserved through softfloat.
- `privileged_check` fires BEFORE the host body, so non-priv-rejection probes never reach their digest / buffer-content validation -- pinned uniformly across `preactivate_feature` / `set_resource_limits` / `set_proposed_producers` / `set_blockchain_parameters_packed`.
- `sysio_assert` and `sysio_assert_message` both throw `sysio_assert_message_exception` on test==0, NOT `sysio_assert_code_exception` (those are distinct intrinsics with distinct exception types).

## Exception-cleanup follow-on (not in this PR)

`recover_key` / `assert_recover_key` currently surface three categories of failure with three different exception types: structural unpack errors leak as `fc::exception`, secp256k1 math errors leak as `fc::assert_exception`, and SYS_ASSERT-gated errors are the only ones that produce a typed `crypto_api_exception`. A follow-on should wrap `fc::raw::unpack` and `public_key::recover` calls in try/catch and rethrow as `crypto_api_exception` with a descriptive message so contracts and oncall can catch a single typed exception for "signature recovery failed". The small-pub-buffer inconsistency (fixed-size keys throw, variable-size keys truncate silently) should be normalized in the same PR. This probe suite pins current behavior; when the follow-on lands, several `BOOST_CHECK_THROW` types here tighten from `fc::exception` to `crypto_api_exception`.

## Verification

98 test cases x 3 WASM runtimes (sys-vm, sys-vm-jit, sys-vm-oc) = 294 invocations.

## Refs

Companion to PR #308 (feature/kv-intrinsic-probe-tests). Same layout: dedicated test-contract directory, extern "C" raw imports at the call site for ABI explicitness, one `[[sysio::action]]` per probe, shared-tester singleton for cost amortization, pre-built `.wasm`/`.abi` committed alongside source.


unittests/intrinsic_probe_tests.cpp
```diff
@@ -0,0 +1,471 @@
...
+}
+
+BOOST_FIXTURE_TEST_CASE(recover_key_corrupt_rs, intrinsic_probe_fixture) {
+   // A single-bit flip in the r component has two acceptable outcomes: the
+   // secp256k1 recovery succeeds and yields a DIFFERENT pub (probe's
+   // in-contract memcmp passes), OR the math fails (host throws
+   // fc::exception). The bad outcome the probe is protecting against is
+   // "recovery silently succeeds and returns the ORIGINAL pub", which never
+   // happens with a real bit flip -- both branches are a correct pin.
+   try {
```
huangminghuang, line 275: This catch makes the probe pass on the exact bad outcome the contract is trying to detect. `recbadrs` enforces "corrupted r/s must not recover the original pubkey" with a `check(...)`; that throws `sysio_assert_message_exception`, which derives from `fc::exception`, so this handler swallows it as an acceptable math failure. Please distinguish host recovery failures from contract assertion failures, or have the action return a sentinel/result that the C++ side can assert.

heifner: Good catch -- you're right that `sysio_assert_message_exception` derives from `fc::exception` so the broad catch silently swallowed exactly the bad-outcome path the contract is designed to detect.

Fixed in a01d791b0c by narrowing the catch ladder: `sysio_assert_message_exception` is caught first and re-surfaces as a `BOOST_FAIL` with the contract's assert message, while plain `fc::exception` (the libfc secp256k1 math-failure path, which throws via `FC_THROW_EXCEPTION(exception, ...)` in `elliptic_secp256k1.cpp`) is still treated as acceptable. Verified locally across sys-vm / sys-vm-jit / sys-vm-oc.

huangminghuang (commented):
**Findings**

- **Medium**: The new probe suite still leaves several non-kv host intrinsics unexercised even though the PR frames this as coverage for the “remaining non-kv host intrinsics” before the legacy span/pointer cleanup. The contract declares `__modti3`, `__umodti3`, `__ashrti3`, `__lshlti3`, and `__lshrti3` at `unittests/test-contracts/intrinsic_probe/intrinsic_probe.cpp:116`, but the Boost driver only runs `multi3`, `divti3`, `udivti3`, and `ashlti3` cases around `unittests/intrinsic_probe_tests.cpp:352`. Same pattern for `__subtf3` declared at `intrinsic_probe.cpp:132`, while the driver only covers add/mul/div around `intrinsic_probe_tests.cpp:376`. The comments also call out `get_context_free_data` at `intrinsic_probe.cpp:23` and `intrinsic_probe.cpp:1394`, but there is no action/test for it in the P3 block around `intrinsic_probe_tests.cpp:460`. If the follow-up cleanup changes any of these host ABI paths, CI can still pass. I’d either add probes for these declared/named intrinsics or narrow the PR description/comments so the coverage boundary is explicit.

**Open Question**

- I’m assuming PR #310 is meant to be the comprehensive non-kv companion to the kv probe suite. If this PR is intentionally only a first slice, the missing-intrinsic point can be handled by making that scope explicit.

huangminghuang (approved):

PR #312, opened 2026-04-24
chain_plugin: get_table_rows pagination + fix reverse off-by-one

jglanz:
## Summary
- Fold `limit == 0` = paginate-all, `unwrap_rows`, and a `filter` predicate into `chain_apis::read_only::get_table_rows`; the original body moves to a single-page primitive `get_table_rows_page`. HTTP callers on the default `limit=50` / no-filter / no-unwrap path take a fast-path forward with zero added indirection.
- Fix a one-row skip at every reverse-page boundary: the wrapper resumes reverse scans using the last returned row's `key` (exclusive `upper_bound`) instead of `next_key` (which points at the first *unseen* row and would be excluded by the exclusive bound).
- `batch_operator_plugin` and `underwriter_plugin` drop their hand-rolled `read_table` pagination/filter/unwrap helpers. Plugin call sites now construct `get_table_rows_params` directly and go through a ~10-line plumbing wrapper that just binds each plugin's timeout. Underwriter gains two indexed queries: `parse_swap_from_attestation` (primary-key `find` on `attestations.id`) and `scan_pending_requests` (`bystatus` range on `uwreqs` for `UNDERWRITE_REQUEST_STATUS_PENDING`).
- Fixes the deterministic epoch-25 → 26 stall in `wire-test-cluster`: the old batch operator read clipped to 50 rows, so once `outenvelopes` had 2 outposts × 25 epochs = 50 rows, epoch-26 envelopes at ids 50-51 were invisible and delivery stopped forever.
- `CLAUDE.md` "Enums over raw values" now mandates `magic_enum::enum_integer|enum_name|enum_value|enum_cast` instead of `static_cast` / hand-rolled tables, with the single exception of protobuf-generated enums (keep using the generated `<Enum>_Name` / `<Enum>_Parse`).

See the commit message (`ff752fd2`) for the line-level walk-through and rationale.

## Test plan
- [ ] `cmake --build build -- -j$(nproc)` — `all` target builds (nodeop, contracts, plugin_test, unit_test, contracts_unit_test, test_fc, per-plugin tests).
- [ ] `./build/tests/plugin_test --run_test=table_reader_tests` — 8/8 green (default wrapped single page, `limit==0` pagination across internal 100-row pages, `unwrap_rows`, `filter` after unwrap, `reverse` + small `limit` latest-N, `reverse` + `limit==0` multi-page without skips, empty-table, `find`+`index_name` exits on single match even with `limit==0`).
- [ ] `./build/plugins/batch_operator_plugin/test_batch_operator_plugin` — 21/21 green.
- [ ] `./build/plugins/underwriter_plugin/test_underwriter_plugin` — 3/3 green.
- [ ] `./build/tests/plugin_test --run_test=get_table_tests` — existing suite unaffected (HTTP fast path preserved).
- [ ] Live validation on a fresh `wire-test-cluster` (`--prod-count=5 --batch-operators=3 --underwriters=1 --epoch-duration=60`): epochs 25 → 26 → 27 advance on 60s cadence; zero `no pending outbound for epoch 2X` log lines; `outenvelopes` grows past 50 rows without stalling delivery. (Passed on dev-023 with the refactored binary; confirmed via batch-op log line `batch_operator_plugi:386 do_poll_epoch_state ELECTED for epoch 27`.)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -52,9 +52,40 @@ Any value drawn from a closed set — chain kind, envelope status, operator type
 
 - Prefer `FC_REFLECT_ENUM`-reflected protobuf enums (`sysio::opp::types::ChainKind`, `sysio::opp::types::EnvelopeStatus`) over hand-rolled sentinels.
 - Decode variants as the enum type: `obj["status"].as<EnvelopeStatus>()`, never `static_cast<EnvelopeStatus>(obj["status"].as_uint64())`.
-- Enum-to-string: `sysio::opp::types::ChainKind_Name(kind)`. Don't hand-roll switches for human-readable names.
+- Don't hand-roll switches for human-readable names.
 
 A rename of `CHAIN_KIND_ETHEREUM` propagates through the compiler automatically when the code holds the enum. It doesn't when the code holds the bare string `"CHAIN_KIND_ETHEREUM"` or the integer `2`.
+
+#### Enum conversions — use `magic_enum`, never `static_cast`
+
+Every enum ↔ raw value conversion goes through `magic_enum` (`#include <magic_enum/magic_enum.hpp>`). `static_cast` and hand-rolled switches survive rename refactors silently; `magic_enum` does not.
+
+| Need | Use |
+|---|---|
+| Enum → underlying integer | `magic_enum::enum_integer(v)` |
+| Enum → string name (runtime) | `magic_enum::enum_name(v)` |
+| Integer → enum (checked `std::optional<E>`) | `magic_enum::enum_cast<E>(n)` |
+| String name → enum (checked `std::optional<E>`) | `magic_enum::enum_cast<E>(name)` |
+| N-th member by declaration order | `magic_enum::enum_value<E>(i)` |
+
+Forbidden once a `magic_enum` form exists:
+
+- `static_cast<uint64_t>(SomeEnum::X)` — use `magic_enum::enum_integer(SomeEnum::X)`.
+- `static_cast<SomeEnum>(int_var)` at a trust boundary — use `magic_enum::enum_cast<SomeEnum>(int_var).value_or(SomeEnum::DEFAULT)`; static_cast past the enum's declared range is UB and it hides bad input.
+- Hand-rolled `if/switch` tables that map an enum to its spelling — use `magic_enum::enum_name`.
+
+**Sole exception — protobuf-generated enums.** `protoc` emits a `<EnumName>_Name(int value)` free function (and `<EnumName>_Parse`) for every proto enum. Use those instead of `magic_enum` for the generated types under `sysio::opp::types::` and any other `.pb.h` enums:
+
+```cpp
+// Protobuf-generated enum — prefer the generated helper.
+auto s = sysio::opp::types::ChainKind_Name(kind);        // ✓
+auto n = magic_enum::enum_name(kind);                    // ✗ works, but skip for proto enums
+
+// Non-proto enum — magic_enum.
+auto s = magic_enum::enum_name(EnvelopeStatus::PENDING); // ✓
+```
+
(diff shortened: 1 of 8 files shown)


heifner:
**1. Reverse-off-by-one is only half-fixed**

The wrapper sidesteps the bug by resuming reverse scans with `last_returned_row["key"]`, but `get_table_rows_page` itself still sets `next_key = sk of first unseen row` for reverse scans (chain_plugin.cpp, reverse branch around the `count >= limit` check). Since `upper_bound` is exclusive and the reverse loop starts at `lower_bound(upper_bound)` then `--itr`, any HTTP caller that follows the natural pagination pattern `upper_bound = prev.next_key` silently drops one row per page boundary.

The existing `get_kv_rows_reverse_pagination_test` (tests/get_table_tests.cpp:1097) actually encodes the dropped-row behavior as expected — the comment literally says `next_key from page1 points to "c" (exclusive), so page2 gets b, a` and the asserts require only `b, a` with no row `c` anywhere.

I put up #313 which fixes this at the page-impl level (reverse `next_key` is the last returned row's key, so `upper_bound = next_key` resumes naturally). #313 also fixes a second related bug in the same spot — secondary-index `next_key` was always hex even with `json=true`, so the bound parser (which expects JSON when `json=true`) rejected it on round-trip. Once #313 lands and this branch rebases on top, the wrapper's reverse-specific `last_row["key"]` cursor branch becomes dead code and can be dropped — this PR shrinks back to its pagination / filter / unwrap scope.

**2. `limit=0` is HTTP-exposed and the deadline doesn't bound it**

`limit` is in the FC_REFLECT list, so callers can POST `{"limit": 0}` to `/v1/chain/get_table_rows` and trigger paginate-all. Per-page deadline is enforced inside `get_table_rows_page`, but the outer paginate-all loop has no deadline check between page calls. Once the deadline expires, each subsequent page call trips the per-row check on its first row, returns `more=true`, the outer loop advances `lower_bound = next_key` and calls again. Net effect: the scan walks the whole table at ~1 row per page call, wall time bounded by table size, not by the HTTP deadline.

Two options, not mutually exclusive:
- Add a deadline check in the outer loop that breaks with `more=true, next_key=...` so the caller resumes.
- Gate paginate-all to C++ callers by moving the trigger to a separate non-reflected field (e.g. `paginate_all`), and reject `limit==0` at the HTTP surface. Stronger guarantee — the two in-tree callers just set the new field.

**3. Threading — pre-existing, but this PR widens the race**

Both plugin callers (batch_operator, underwriter) dispatch their reads via `cron_plugin.add_job(...)`, which runs on `cron_service`'s private `boost::asio::thread_pool`. `chain_plug->get_read_only_api(...)` is a bare constructor — when the cron thread invokes `ro.get_table_rows(...)()`, the scan runs on the cron worker thread with no coordination against the controller's write window. If a tick lands during the write window, the cron thread iterates `kv_index` while the main thread is mutating it. `get_table_rows_page` doesn't take `with_read_lock`. Possible outcomes: torn pointer reads during rebalance, silently inconsistent row sets.

This was already wrong pre-PR — old `limit=50` scans just kept the race window small. The PR widens it two ways: `limit=0` scans O(table size), and multi-page iteration isn't a consistent snapshot — a block can apply between pages.

Fix belongs in the `read_table` helper in each plugin: post the scan onto `app().executor()` via the read_only queue and await, so the scan runs during the read window (same discipline as the HTTP path). Rough shape:

```
std::promise<get_table_rows_result> pr;
app().executor().post(priority::read_only, exec_queue::read_only,
   [this, p, &pr]() mutable {
       auto ro = chain_plug->get_read_only_api(timeout);
       auto r = ro.get_table_rows(p, deadline)();
       pr.set_value(std::get<get_table_rows_result>(std::move(r)));
   });
return pr.get_future().get();
```

**4. Follow-on, not a blocker — architectural layering**

Extending `get_table_rows_params` with C++-only fields (`filter`, `unwrap_rows`) treats the HTTP shape as canonical for in-process callers. Every row pays ABI lookup + variant wrapping + `{key, value, payer?}` construction + `fc::variant_object` field-poking. For plugins that already have typed row structs (`outbound_envelope_record`, etc.), a plugin-facing typed scan below the HTTP layer (raw bytes -> `fc::raw::unpack<T>`) would skip all that, and the filter/unwrap escape hatches would stop being necessary. I would be happy to work on that refactor.

brianjohnson5972 (approved):

PR #313, opened 2026-04-24
Fix get_table_rows reverse pagination and secondary next_key format

heifner:
Two related fixes to `read_only::get_table_rows`.

(1) Reverse off-by-one. Reverse iteration set `next_key` to the first unseen row's key. Since callers resume with `upper_bound = next_key` and `upper_bound` is exclusive, the reverse loop starts at `lower_bound(upper_bound)` then `--itr` and yields the first row strictly below next_key -- so next_key itself was skipped. Each reverse page boundary dropped one row. Fix: in reverse, `next_key` is now the LAST RETURNED row's key. `upper_bound = next_key` yields the first previously unseen row on the next page. Applies to both primary- and secondary-index reverse branches; forward is unchanged (lower_bound is inclusive).

Example (keys 1..5, reverse, limit=2): pre-fix [5,4] -> [2,1] (3 dropped); post-fix [5,4] -> [3,2] -> [1] (every row exactly once).

(2) Secondary index next_key format. Secondary `next_key` was always hex of the raw sec_key bytes, regardless of `p.json`. But the bound parser expects JSON when `p.json` is set, so `upper_bound = next_key` did not round-trip on json=true. Fix: when `p.json` is set, secondary `next_key` is now a JSON object matching the bound syntax (e.g. `{"byowner":"u4"}`), decoded via `be_key_codec::decode_key` over `{si.name, si.key_type}` after stripping any scope prefix. Falls back to hex on decode failure. Hex behavior is preserved for json=false.

Tests: primary and secondary reverse pagination over 5 rows with limit=2, mirrored json=true and json=false cases (four test cases total). `get_table_next_key_test` (sec-6) updated to use the natural json=true round-trip instead of the old json=true -> json=false hex workaround.

### DIFF
--- a/plugins/chain_plugin/src/chain_plugin.cpp
+++ b/plugins/chain_plugin/src/chain_plugin.cpp
@@ -1869,6 +1869,29 @@ read_only::get_table_rows( const read_only::get_table_rows_params& p, const fc::
          return r;
       };
 
+      // Format a secondary-index `next_key` value. When `p.json` is set, emit
+      // a JSON object matching the bound syntax (e.g. `{"byowner":"u4"}`) so
+      // `upper_bound = next_key` round-trips through the bound parser, which
+      // expects JSON when `p.json` is set. Otherwise emit hex of the raw sk
+      // bytes. Falls back to hex on any decode failure.
+      auto emit_secondary_next_key = [&](std::string_view sk) {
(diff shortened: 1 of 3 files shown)


heifner:
Fix being rolled into #312

PR #315, opened 2026-04-25
Fc: variant performance improvements

heifner:
## Summary

Speed up `fc::variant` for the dominant create-then-iterate workload (JSON serialization, pass-by-value across thread boundaries, ABI-decoded rows). 11 individually-reviewable commits: test coverage and a microbenchmark scaffold first, then a sequence of focused perf changes, each with a bench delta in its commit message.

## Per-commit summary

**Setup**

1. `delete unused arithmetic operators` — `fc::variant operator+/-/*//` were unreachable dead code; one had a never-reached loop bug. `= delete` so any future caller fails at compile time with a clear message.
2. `add microbenchmark scaffold` — `libraries/libfc/benchmark/variant_bench` modeled on the existing `abi_serializer_bench`. Plain chrono, `EXCLUDE_FROM_ALL`, manual run; per-scenario warmup + median-of-10.
3. `capture benchmark baseline` — starting numbers with observations on each scenario (RelWithDebInfo, 12th Gen i9-12900K).
4. `expand test coverage` — 102 new test cases in 7 files locking down current behaviour before any perf change.

**Perf changes** (each commit message has its own bench delta)

5. `as_enum_value uses from_chars` — replace `std::stoll + catch(...)` with non-throwing `std::from_chars`. 11.6 → 4.6 ns valid path (2.5x), 3976 → 2965 ns invalid path.
6. `lazy-allocate variant_object vector` — default-constructed `vo`/`mvo` no longer call `make_shared<vector<entry>>` until first write. As a side effect, fixes a latent aliasing bug in `vo::operator=(const mvo&)` where writing through a shared `shared_ptr` could mutate any sibling vo that shared the same backing vector. 7.5 → 1.4 ns empty-mvo ctor.
7. `add non-throwing find_or helper` — `variant_object::find_or(key, default)` collapses the common `contains() ? op[k] : default` double-scan + throw-on-miss into one scan. 19.9 ns vs 38.9 ns for `contains_then_op_50key`.
8. `get_string() returns std::string_view` — API change, prep for SSO. `as_string()` (returns `std::string` by value) unchanged for callers that need an owning copy. Cascaded through `fc::reflector<E>::from_string`, `fc::throw_bad_enum_cast`, `fc::from_hex`, `fc::to_int64/uint64/double`, `asset::from_string`, `symbol::from_string`, `string_to_symbol` so they take `string_view` internally.
9. `add SSO for short strings` — strings ≤ 14 bytes stored inline in the variant's 16-byte buffer instead of heap-allocating a `std::string`. Wire format unchanged: `raw::pack(variant)` translates `string_sso_type` to `string_type` at the wire boundary, so existing peers and chainbase-resident buffers deserialize identically. 14.2 → 3.6 ns short-string ctor (3.9x), 11.0 → 2.2 ns short-string copy (5x).
10. `same-type op=(const variant&) reuses heap object` — `op=(const&)` was doing `clear()` (delete) + `new` even when old type == new type. Route same-type assign through the existing heap object's own assignment operator. 17.0 → 3.3 ns long-string assign (5.2x), 10.3 → 2.0 ns object assign (5.1x), 32.9 → 12.1 ns array assign (2.7x).
11. `re-baseline bench at Release (-O3)` — BASELINES.md gains a Release column for absolute deployment-build numbers; per-commit log stays at RelWithDebInfo for comparable deltas.

## Headline numbers (Release -O3)

| Scenario | Baseline | After | Speedup |
|---|---:|---:|---|
| `ctor_short_string` (5 chars) | 14.2 | 3.6 | 3.9x |
| `copy_short_string` | 11.0 | 2.8 | 3.9x |
| `ctor_empty_mvo` | 7.5 | 1.5 | 5.0x |
| `as_enum_string_valid` | 11.6 | 3.6 | 3.2x |
| `assign_long_string_to_long` | 17.0 | 3.3 | 5.2x |
| `assign_object_to_object` | 10.3 | 2.0 | 5.1x |
| `assign_array_to_array` | 32.9 | 11.2 | 2.9x |
| `find_or_50key_hit` (new helper) | (n/a) | 19.3 | vs `contains_then_op` 38.4 = 2x |
| `ctor_long_string` (heap path) | 19.7 | 20.2 | unchanged |

Workload-shaped scenarios (`json_parse_50key`, `walk_50key_by_name`) are within run-to-run noise — the per-allocation savings exist but are diluted by parser/iteration overhead in the 50-key fixture.

## Compatibility notes

- `fc::variant::get_string()` returns `std::string_view` (was `const std::string&`). `.size()`, `.empty()`, `==`, range-based for, passing to `string_view`-typed parameters all still work. Callers passing the result to APIs that take `const std::string&` need to materialise a `std::string` explicitly; this PR includes the migrations for every in-tree caller.
- `fc::raw::pack(variant)` wire format is unchanged — the new `string_sso_type` tag translates to `string_type` at the wire layer. Persisted variants and net-protocol bytes round-trip without code changes on the receiver.
- `fc::variant::visitor::handle(const std::string&)` is now `handle(std::string_view)`. Only one production override (`fc::raw::variant_packer`) needed updating; included.
- `fc::variant` thread-safety contract is documented in the lazy-alloc commit and in feedback memory: copy/move to another thread is safe, concurrent reads on the same instance are safe, concurrent read+write on the same instance is the caller's problem (POD-like). The lazy-alloc and SSO changes preserve this.

## Decisions documented in commit messages

- COW for variant string/blob payloads is **deferred** — technically permissible under the thread-safety contract, but implementation cost is high relative to remaining wins. Revisit only if a workload surfaces it.
- Hash side-table on `variant_object` for O(1) lookups is **rejected** — dominant lifecycle is build-once + iterate, neither calls `find()`. Hash index would tax every construction for a find-heavy minority.


libraries/libfc/src/variant.cpp
```diff
@@ -266,27 +340,47 @@ variant& variant::operator=( const variant& v )
...
+   // delete-then-new.  For string/object/array/blob (and the std::string-
+   // backed multi-precision integer encodings) this skips an alloc+free
+   // pair.  Inline encodings fall through to the bytes-copy default arm.
+   if( get_type() == src_type ) {
+      switch( src_type ) {
+         case object_type:
+            **reinterpret_cast<variant_object**>(this) =
+               **reinterpret_cast<const const_variant_object_ptr*>(&v);
+            return *this;
+         case array_type:
```
huangminghuang, line 393: assigns array-to-array directly through the same-type fast path. If the RHS is nested inside the LHS, e.g. v = v.get_array()[0] where both are array_type, the destination vector assignment can destroy the outer vector and invalidate the nested source while it is being read. The different-type path correctly deep-copies through variant tmp(v) first; the same-type array path needs equivalent alias protection or a test covering nested array self-assignment.

libraries/libfc/include/fc/variant.hpp
```diff
@@ -290,8 +298,19 @@ namespace fc
...
+              // std::from_chars is non-throwing: avoids the stoll exception
+              // round-trip on the invalid-text fallback path, which was the
+              // dominant cost on bad input (~4 us per call) and ~25% of the
+              // valid-text path too.  Behaviour matches stoll for the
+              // domain we hit: leading minus accepted, leading whitespace
+              // and leading '+' rejected, suffix garbage silently ignored
+              // (so "1abc" still yields 1, matching the prior behaviour).
+              std::string_view s = get_string();
+              int64_t parsed = 0;
+              auto [ptr, ec] = std::from_chars(s.data(), s.data() + s.size(), parsed);
```
huangminghuang, line 312: replaces std::stoll with std::from_chars, but stoll(" 1") and stoll("+1") both returned 1; from_chars rejects both. The comment says behavior is equivalent, and the new tests cover suffix garbage but not leading whitespace or +. If preserving behavior is required, trim/skip leading whitespace and handle optional + before from_chars.

libraries/libfc/src/variant.cpp
heifner: Thanks. After looking at it, both the same-type and different-type aliased self-assign cases were UB in the previous fc::variant via `clear()`-then-`new`; the protection in e6e198a386 was incidental to the refactor.

Rather than extending protection to the same-type fast path (the safe options are either always deep-copy the source vector -- ~+20 ns on the `assign_array_to_array` microbench -- or a partial pointer-aliasing check that catches direct cases but misses deep nesting like `v = v.get_array()[i].get_array()[j]`), I reverted to UB everywhere and added a debug-only assertion that catches the common direct cases:

```
assert(rhs_not_aliased(this, v) && "...");
```

The helper short-circuits on array_type/object_type only and is only called from `assert()` so it (and the call site) compile out under NDEBUG. Deeper nesting still slips past, but direct-aliasing regressions will trip the assertion in our assertion CI builds.

See 2e4ab5268c.

libraries/libfc/include/fc/variant.hpp
heifner: You're right, the previous comment overstated the equivalence. Updated to spell out the actual differences:

- leading minus accepted (matches stoll)
- leading `+` REJECTED (stoll accepted)
- leading whitespace REJECTED (stoll accepted)
- suffix garbage silently ignored, e.g. `"1abc"` yields 1 (matches stoll)

ABI-emitted enum strings never contain whitespace or sign prefixes, so the stricter contract surfaces malformed input as an exception rather than silently parsing it. Added `string_source_with_leading_whitespace_throws` and `string_source_with_leading_plus_throws` to pin the new behaviour.

See 659cd5bc49.

libraries/libfc/src/variant.cpp
```diff
@@ -190,6 +247,31 @@ typedef const variants* const_variants_ptr;
...
+// Deeper nesting (rhs reachable through an inner array/object owned by lhs) remains UB but may slip past undetected.
+// Only ever called from assert(), so it (and the operator= call site) compile out under NDEBUG.
+inline bool rhs_not_aliased( const variant* lhs, const variant& v )
+{
+   const auto t = lhs->get_type();
+   if( t == variant::array_type ) {
+      const variants& dst_vec = **reinterpret_cast<const variants* const*>(lhs);
+      const variant* begin = dst_vec.data();
+      const variant* end   = begin + dst_vec.size();
+      return &v < begin || &v >= end;
```
huangminghuang: Those comparisons are undefined behavior when &v is not a pointer into the same array as begin/end, which is exactly the normal non-aliased assignment case this helper is trying to validate. It is debug-only via assert, but assert-enabled CI can still execute UB here. Use address integers or a defined total-order comparator instead, e.g. compare reinterpret_cast<std::uintptr_t> ranges after handling the empty-vector case.

heifner: Fixed in https://github.com/Wire-Network/wire-sysio/pull/315/commits/f5b8e9c7c0b06bb0cfebf35dd0090db90725ffab

huangminghuang (approved):

PR #316, opened 2026-04-27
chain_plugin: run read_table_rows inline on main thread

heifner:
`read_table_rows` posts onto the executor's read_only queue and blocks the caller on the future. From a main-thread context (e.g. a plugin's `plugin_startup`), the main thread is the only drain for that queue during write window, so the post sits until the deadline and returns empty rows. This silently broke `batch_operator_plugin::refresh_outposts` at startup -- the cron pool came up sized for zero OPP jobs.

When `std::this_thread::get_id() == executor().get_main_thread_id()`, run the scan inline; the read-window discipline is already satisfied on the main thread. Off-thread callers keep the post + wait path so cron threads still iterate chainbase during the read window. Scan body is factored into a `run_scan` lambda shared by both paths.

Follow-on to #312.


jglanz (approved):
approved once tests pass (I did not actually review, but I'm going to cherry pick to my workspace and confirm)

PR #317, opened 2026-04-27
Achieved memory & on-chain storage stability (`>2400` epochs over `42hrs+`) with integrated pruning, cross-chain data consistency & chain-agnostic orchestration layer

jglanz:
# Achieved memory & on-chain storage stability (`>2400` epochs over `42hrs+`) with integrated pruning, cross-chain data consistency & chain-agnostic orchestration layer

## Overview

Achieved memory & on-chain storage stability (`>2400` epochs over `42hrs+`) with integrated pruning, cross-chain data consistency & chain-agnostic orchestration layer.

This commit implements a comprehensive architectural refactoring to achieve cross-chain data-state consistency between WIRE and external blockchains. It introduces a chain-agnostic orchestration layer (`depot_ops` + `outpost_opp_job`) that cleanly separates WIRE table operations from chain-specific client implementations, fixing critical bugs in secondary index queries, row unwrapping, and EIP-1559 RLP signature encoding that previously broke batch operator envelope delivery.

Additionally, the commit removes obsolete auto-generated contract files, adds devcontainer awareness to build tooling, eliminates deprecated macOS signing infrastructure, and standardizes CMake module naming conventions.

---

### **Removed Deprecated Auto-Generated Contract Files**

Cleaned up stale build artifacts that should never have been committed to version control:

- **Deleted** `contracts/sysio.chalg/sysio.chalg.cpp.actions.cpp` (138 lines of codegen action wrappers)
- **Deleted** `contracts/sysio.chalg/sysio.chalg.dispatch.cpp` (40 lines of codegen dispatch handlers)  
- **Deleted** `contracts/sysio.chalg/sysio.chalg.sysio.chalg.cpp.desc` (264 lines of ABI metadata)

These files are regenerated during the build process and do not belong in the repository.

---

### **Chain-Agnostic Orchestration Layer**

Introduced two new abstractions to decouple WIRE operations from outpost-specific client logic:

#### **New Interface: `depot_ops` (`batch_operator_plugin/depot_ops.hpp`)**

Provides chain-agnostic WIRE table/action operations:

- `read_pending_outbound()`: Queries `sysio.msgch::outenvelopes` for pending delivery  
- `has_delivered_envelope()`: Checks `sysio.msgch::envelopes` secondary index for delivery status  
- `deliver_to_depot()`: Pushes inbound messages via `sysio.msgch::deliver` action  
- `emit_debug_envelope()`: Publishes debugging events to `external_debugging_plugin`  
- `within_epoch_window()`, `is_elected()`, `current_epoch()`: Epoch state accessors

Concrete implementation (`depot_ops_impl_t`) translates high-level requests into `read_table()` / `push_action()` calls with row unwrapping and batch operator authorization checks.

#### **New Orchestrator: `outpost_opp_job` (`batch_operator_plugin/outpost_opp_job.hpp`)**

Per-outpost job lifecycle manager:

- Owns a single `outpost_client` instance (Ethereum or Solana)  
- Runs on dedicated cron thread to avoid blocking other outposts  
- **Outbound Flow (WIRE → L1):**  
  Queries pending envelope → submits to outpost → waits for confirmation → emits debug event  
- **Inbound Flow (L1 → WIRE):**  
  Polls L1 events → checks delivery status → submits to WIRE → prevents duplicates

#### **Refactored `batch_operator_plugin::impl`**

Replaced monolithic `run_epoch_cycle()` with job-based architecture:

- Removed direct outpost client management (`opp_client`, `sol_outpost_client`, transient state tracking)  
- Introduced `std::map<uint64_t, std::shared_ptr<outpost_opp_job>> opp_jobs` (one job per outpost)  
- Added `build_opp_jobs()` factory using chain plugin `make_outpost_client()` APIs  
- Centralized WIRE contract constants (`msgch` and `epoch` namespaces) to prevent string literal duplication

---

#### **1. Secondary Index Query Fix (`chain_plugin.cpp`)**

**Problem:** Secondary indexes on `multi_index` tables store keys as `[scope:8B][value:N]`, but `json=true` queries only provided the value portion in bounds, causing incorrect row lookups.

**Solution:** Prepend `scope_prefix_bytes` to lower/upper bounds when:
- Query uses secondary index (`!resolved_index_name.empty()`)  
- Scope is set (`!scope_prefix_bytes.empty()`)  
- JSON mode is enabled (`p.json == true`)

**Test Coverage:**
- `(sec-10)`: `find` on `multi_index` secondary index returns exact match  
- `(sec-11)`: `find` miss returns zero rows (not lexicographically-next row)

#### **2. Row Unwrapping Fix (`batch_operator_plugin.cpp`)**

**Problem:** `chain_plugin::get_table_rows` wraps rows as `{"key", "value", "payer"}`, breaking direct field access in plugin code.

**Solution:** Post-process `combined.rows` vector to extract `"value"` object using safe pattern:
```cpp
fc::variant value{row_obj["value"]};  // Copy to temporary
row = std::move(value);                // Move-assign to avoid self-destruction
```

**Critical Pattern:** Direct assignment `row = row["value"]` causes undefined behavior because `variant::operator=` calls `clear()` before reading the source.

#### **3. EIP-1559 RLP Signature Encoding Fix (`ethereum_rlp_encoder.cpp`)**

**Problem:** Signature components `r` and `s` were encoded as fixed 32-byte strings. When the MSB was `0x00`, strict RLP decoders (alloy-rs in reth/anvil) rejected transactions with "leading zero" errors (~1/256 signatures affected).

**Solution:** Implemented `encode_sig_scalar()` to strip leading zeros and emit minimal big-endian integers per EIP-2718.

**Test Coverage:**
- `signed_rlp_strips_leading_zeros_in_r_and_s`: Verifies 31-byte encoding when MSB is `0x00`  
- `eip1559_signed_rlp_strips_leading_zero_in_s`: Regression test for production failure case (612-byte payload, s=`0x00 9b bd d7 ...`)

---

#### **Outpost Client Factory Methods**

**Ethereum:** `outpost_ethereum_client_plugin::make_outpost_client()`  
Creates `outpost_ethereum_client` from ETH client ID, outpost ID, chain ID, contract addresses. Validates client existence and contract address configuration.

**Solana:** `outpost_solana_client_plugin::make_outpost_client()`  
Creates `outpost_solana_client` from SOL client ID, outpost ID, chain ID, program ID. Filters loaded IDL set to match `OPP_SOLANA_OUTPOST_PROGRAM_NAME` and asserts IDL availability.

#### **Debug Event Type Relocation**

Moved `DebugEnvelopeEvent` to dedicated header `batch_operator_plugin/debug_envelope_event.hpp` to break circular dependency between `batch_operator_plugin.hpp` and `depot_ops.hpp`. Remains in `sysio::opp::debugging` namespace.

---

#### **CMake Module Standardization**

- Renamed `VersionUtils.cmake` → `version-tools.cmake`  
- Renamed `AddTestHelpers.cmake` → `test-helpers.cmake`  
- Follows lowercase-hyphenated naming convention

#### **Devcontainer Detection in Version Generation**

Enhanced `version-tools.cmake` to gracefully handle non-git environments:
```cmake
if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/.git AND NOT "$ENV{IN_DEVCONTAINER}" STREQUAL "1")
```
Sets `V_HASH="unknown"` and `V_DIRTY="true"` in devcontainers/tarballs to prevent build failures.

#### **Removed Deprecated MAS Signing Infrastructure**

Deleted `cmake/MASSigning.cmake` (22 lines of unused App Store code signing macros) and removed `mas_sign(${KEY_STORE_EXECUTABLE_NAME})` invocation from `programs/kiod/CMakeLists.txt`.

#### **OPP Bundle Generation Enhancements**

**`generate-opp-bundles.fish`:**
- Added `


heifner:
### 1. `deliver` epoch-mismatch assert builds the message unconditionally

`contracts/sysio.msgch/src/sysio.msgch.cpp::deliver`:
```cpp
check(env_epoch == epoch,
   "envelope epoch_index mismatch: envelope=" + std::to_string(env_epoch) +
   " current=" + std::to_string(epoch));
```
The triple-string concatenation runs on every successful delivery (`check`'s message argument is evaluated before the predicate), so each successful delivery allocates a ~60-byte temporary that's then thrown away. Use the lambda overload of `sysio::check` (defined in `wire-cdt/libraries/sysiolib/core/sysio/check.hpp:81`) so the format work only fires on the failure branch:
```cpp
check(env_epoch == epoch, [&] {
   return "envelope epoch_index mismatch: envelope=" + std::to_string(env_epoch) +
          " current=" + std::to_string(epoch);
});
```

### 2. Redundant `contains` before known-safe `erase` (perf nit)

`contracts/sysio.msgch/src/sysio.msgch.cpp::evalcons`:
```cpp
msgs.emplace(get_self(), id_key{msg_id}, ...);
// ... ~30 lines later ...
if (msgs.contains(id_key{msg_id})) {
   msgs.erase(id_key{msg_id});
}
```
Nothing in between deletes the row, so `contains` is provably true and the check costs an extra key lookup per evalcons. Drop it and `erase` directly — `kv::table::erase(K, msg)` asserts on a missing row, which is a louder signal than a silent no-op if the invariant ever changes.

### 3. `read_u32_le` is implicitly little-endian-only (nit)

`plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp`:
```cpp
uint32_t v;
std::memcpy(&v, buf.data() + off, 4);
return v;
```
Wire is x86_64 only today, so this is correct. A self-documenting `static_assert(std::endian::native == std::endian::little)` at the top of the function would catch a future port to a BE host instead of silently producing wrong epoch values.

### 4. Why is `commitment_t::confirmed` hardcoded on the Solana read path?

`plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp::read_inbound_envelope`:
```cpp
auto info = _entry->client->get_account_info(
   _program_client->latest_outbound_envelope_pda,
   fc::network::solana::commitment_t::confirmed);
```
The new write paths in this PR thread `solana_confirm_options` through to give callers control over commitment level — but the read path locks `confirmed` at the call site. Is this an intentional asymmetry (e.g. inbound reads must always be confirmed for consensus reasons)? If so, a one-line comment would close the gap.

### 5. `dlog` of full `getLatestOutboundEnvelope raw_hex={}` floods debug logs (nit)

`plugins/outpost_ethereum_client_plugin/src/outpost_ethereum_client.cpp`:
```cpp
dlog("outpost_ethereum_client[{}]: getLatestOutboundEnvelope raw_hex={}",
     to_string(), raw_hex);
```
Hex blob can be hundreds of bytes per call. If debug logging ever runs in production this floods. Truncate to first ~64 bytes + size suffix in the format string.

heifner (commented):
Three items from a focused review of the consensus-zone changes (contracts + outpost relay). Inline comments below.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -84,115 +75,136 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+   // the **batch operator's only Solana-side tx is `epoch_in`**: the
+   // last-chunk call triggers the program's `finalize_envelope`, which
+   // (a) records the operator's delivery, (b) on consensus reach also
+   // fires `emit_outbound_inner` inline (drains the queued outbound
+   // attestations into a packed envelope and writes it to the
+   // `latest_outbound_envelope` PDA), and (c) self-closes this
+   // operator's chunk_buffer. No separate `emit_outbound_envelope` or
+   // `cleanup_envelope_chunks` tx is needed in the relay.
+   std::string last_sig;
+   for (uint16_t i = 0; i < total_chunks; ++i) {
```
heifner, line 102: Envelope-cap mismatch with batch-op deadline.

Contract allows 64 KiB envelopes (`MAX_ENVELOPE_BYTES = 65'536` in `sysio.msgch.cpp`). Solana relay chunks at 672 B, so worst case is 98 chunks. Each chunk now goes through `execute_tx_and_confirm` which awaits `processed`-commitment (~400 ms - 2 s round-trip on a healthy validator).

`_outpost_deadline` defaults to `DELIVERY_TIMEOUT_MS = 15000`. The math doesn't close:

| envelope | chunks | min wall time @ 400 ms/chunk | deadline |
|---|---|---|---|
| 2.5 KB (dev-026) | 4 | ~1.6 s | 15 s |
| 20 KB | 30 | ~12 s | 15 s |
| 32 KB | 49 | ~20 s | 15 s |
| 64 KB (cap) | 98 | ~40 s | 15 s |

When the deadline expires mid-stream, partial chunks sit in `chunk_buffer` (only the original uploader can reap them via `cleanup_envelope_chunks`), `_last_outbound_epoch` doesn't advance, next tick re-attempts -- but `buildenv` will have already overwritten the on-chain outenvelope row (one-deep retention), the WIRE-side attestations were marked PROCESSED and erased inside that same `buildenv`, and `envelope_log` only carries the checksum. Payload unrecoverable.

Fix options, in order of preference:
1. Lower `MAX_ENVELOPE_BYTES` on the WIRE side to whatever the relay can deliver under a tight deadline budget at production cadence (~13 chunks * 672 B = ~8.7 KB with margin).
2. Scale the per-call deadline by `total_chunks` inside the relay (`outpost_deadline + per_chunk_budget * total_chunks`).
3. Drop confirmation on intermediate chunks (only `_and_confirm` the last one) -- but this re-introduces the silent-drop hazard the `_and_confirm` change was meant to close.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -14,10 +14,36 @@ using opp::types::AttestationStatus;
...
+
+/// Hard cap on the encoded outbound envelope size, mirroring the Solana
+/// (`opp_outpost::MAX_ENVELOPE_BYTES`) and Ethereum (`OPP.MAX_ENVELOPE_BYTES`)
+/// caps. 64 KiB is the e2e-supported maximum across WIRE / Ethereum / Solana,
+/// bound by Solana's 256 KiB BPF heap divided by ~3.3× envelope-size peak heap
+/// usage during the finalising chunk's `Envelope::decode + keccak::hash + clone`.
+/// The `buildenv` packing loop uses this to decide how many READY attestations
+/// to bundle into the current epoch's envelope; any that don't fit stay in the
+/// `attestations` table with status READY for the next epoch's `buildenv` call.
+constexpr size_t   MAX_ENVELOPE_BYTES         = 65'536;
```
heifner, line 33: Cap of 64 KiB is set on the contract side but the Solana relay can't deliver that within `DELIVERY_TIMEOUT_MS = 15000`. See the comment on `outpost_solana_client.cpp` chunk loop -- the contract permits envelopes the relay cannot move. Recommend lowering this to ~8-10 KiB pre-launch unless the relay deadline is also scaled by chunk count.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -378,6 +533,13 @@ void msgch::buildenv(uint64_t outpost_id) {
    auto out = zpp::bits::out{packed, zpp::bits::no_size{}};
    (void)out(env);
 
+   // Hard backstop — the estimator should always over-count, but a final
+   // size check guarantees the envelope cannot exceed the cross-chain cap
+   // even if the conservative estimator drifts.
+   check(packed.size() <= MAX_ENVELOPE_BYTES,
+         "sysio.msgch::buildenv: serialised envelope exceeds MAX_ENVELOPE_BYTES "
```
heifner, line 540: This is a dead-letter trap.

The Phase 2 estimator (`ENVELOPE_BASELINE_BYTES = 512` + `ATTESTATION_OVERHEAD_BYTES = 24` per entry) under-counting fires this `check` *after* the candidates have been marked PROCESSED. When it trips, the action aborts and every state mutation rolls back -- PROCESSED reverts to READY, no outenvelope row created. Next epoch's `buildenv` collects the same set of READY attestations, runs the same estimator to the same conclusion, trips the same check. Permanent loop until an attestation is added/removed or the contract is upgraded.

The 24 B/attestation budget is plausibly tight enough to be wrong on edge cases:
- `vuint32_t data_size` for >= 16 KB payloads serialises to 3-4 bytes
- protobuf field tag (1-2 B) + length prefix (1-5 B) per nested attestation
- the `Message` and `Envelope` outer length prefixes scale with included payload size

Suggested fix: pop-and-resize on overflow rather than abort.

```cpp
while (included_count > 1) {
   <serialise>
   if (packed.size() <= MAX_ENVELOPE_BYTES) break;
   --included_count;
   entries.pop_back();
   included_ids.pop_back();
}
check(packed.size() <= MAX_ENVELOPE_BYTES, "...");
```

Then revert PROCESSED status on the popped entries before the early-return. Cap stays inviolable, estimator drift becomes self-correcting.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -220,6 +299,44 @@ void msgch::evalcons(uint64_t outpost_id, uint32_t epoch_index) {
...
+           it != evict_idx.end() && it->by_outpost_epoch() == composite; ) {
+         it = evict_idx.erase(std::move(it));
+      }
+
+      // Drop the just-inserted `messages` row. Its raw_payload mirrors
+      // the envelope bytes we already discarded; downstream consumers
+      // read the audit log for trail and the attestations table for
+      // queued outbound work.
+      if (msgs.contains(id_key{msg_id})) {
+         msgs.erase(id_key{msg_id});
```
heifner, line 336: Emplace + erase in the same action.

The row was just emplaced ~60 lines above (`msgs.emplace(...)` with `.raw_payload = raw`, up to 64 KiB) and is erased here. Net storage delta is zero but the action pays for:
- a deep copy of `raw` into the row payer's RAM (peak charge happens before the erase rolls it back)
- KV serialisation for `message_entry`
- chainbase undo-log entries for both the insert and the delete
- the `contains` probe on the previous line (an extra KV lookup that always returns true since we just inserted `msg_id`)

Also: if `sysio.msgch`'s RAM payer is near its limit, `evalcons` can start failing on the temporary peak even though the steady-state delta is zero. Pre-launch this is theoretical; post-launch it's a footgun.

Suggested fix: if the `messages` table really has no downstream consumer (per the inline comment), drop the emplace + the contains/erase pair entirely. If it ever gets one, reintroduce the emplace then but skip the inline erase.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
jglanz: the solana max and envelope max are UNRELATED.  the solana max is per chunk

contracts/sysio.msgch/src/sysio.msgch.cpp
jglanz: I believe this is the same misunderstanding?

contracts/sysio.msgch/src/sysio.msgch.cpp
jglanz: Can you add a JIRA umbrella ticket assigned to both you and I and drop these performance comments/changes/fixes into there as subtasks?

heifner (approved):

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
heifner: Agreed the math doesn't close at the 64 KiB cap, but I think the deeper issue is structural: `buildenv` drops the `outenvelope` row + PROCESSED attestations before delivery is observed at finalized commitment on the destination, so a deadline overrun, an RPC failure, or a Solana micro-fork after processed-commitment all lose the payload. 

I think we need an outbound delivery-confirmation + finality round-trip. I'll create a JIRA issue for this.

PR #318, opened 2026-04-28
Drop host compiler-builtin intrinsics; retarget tests at contract-side librt

heifner:
## Summary

Drops the 43 host implementations of compiler-rt intrinsics (9 int128 + 34 softfloat-backed) from wire-sysio. Contracts now resolve these locally from CDT's `librt.a` (default-linked) and `libsf.a` (opt-in via `--use-rt`).

## Why

Pre-launch cleanup. Each host intrinsic is a place that has to behave bit-identically across architectures and across the three runtimes. Moving the implementations into per-contract WASM lets compiler-rt + softfloat (both upstream-tested) provide them, and reduces our consensus question to "do all three runtimes execute the same WASM identically" -- something the standard test framework already exercises across sys-vm, sys-vm-jit, and sys-vm-oc.

This is the wire-sysio half. The CDT-side changes (librt standalone rewrite, default-link, `cdt.imports` prune, `cdt-ld` hint, build-fail test) are on `Wire-Network/wire-cdt#cleanup/drop-builtin-host-fallback`.

## Commits

1. `tests: retarget compiler-builtin tests at contract-side librt` -- `test_compiler_builtins.cpp` drops `sysio_wasm_import` so calls resolve to local librt; adds tests for `INT128_MIN`/-1 wrap and shift-count >= 128 saturation. New `test_f128_builtins.cpp` covers all 34 softfloat-backed entry points -- arithmetic, comparisons (NaN edges), conversions (extend/truncate, int<->tf, sf/df<->ti). `api_tests.cpp` gains `compiler_builtins_tests` and `f128_builtins_tests` Boost cases. `int128_tests.cpp` keeps its fc-library section (string conversion, variant roundtrips) and drops the host-direct compiler-builtin section. `float128_builtin_tests.cpp` and `divmod_host_function_overflow_wast` deleted. Long-double-using test contracts recompiled with `--use-rt`. A pre-existing copy-paste bug in `test_umodti3` (signed expectations on unsigned operations) is corrected.
2. `chain: drop host compiler-builtin registrations` -- drops 43 entries from the intrinsic signature registry, sys-vm.cpp registrations, and the OC PIC table. Deletes `libraries/chain/webassembly/compiler_builtins.cpp` and `libraries/builtins/`. Updates `sys-vm-oc/gs_seg_helpers.h::SYS_VM_OC_MEMORY_STRIDE` (the prologue shrinks past a 4KB rounding boundary; the static_assert in `memory.hpp` is the deliberate tripwire -- pre-launch the renumber is fine). Drops the unused `<compiler_builtins.hpp>` include from `wasm_interface.cpp` and the 43 method declarations from `interface.hpp`.

## Behavior changes

- `int128_t / 0` and `int128_t % 0` in contracts throw `sysio_assert_message_exception` "divide by zero" (was `arithmetic_exception`).
- `INT128_MIN / -1` wraps to `INT128_MIN`; `INT128_MIN % -1` is 0 (librt force-handles the signed-overflow UB case).
- Shift count >= 128 on int128 ops saturates: 0 for left/logical-right, sign-extension for arithmetic right.
- Contracts that use long double must pass `--use-rt` to the CDT toolchain or get a clean wasm-ld error pointing at it.

## Stacked

Targets `feature/intrinsic-span-cleanup` (#314).

## Coordination

Depends on https://github.com/Wire-Network/wire-cdt/pull/51


huangminghuang (commented):
sys-vm.cpp at PR head removes the host registrations for the compiler builtins, but the PR does not rebuild the shipped system contract WASMs. I parsed PR-head imports and these still import env.__ashlti3: contracts/sysio.authex, sysio.bios, sysio.chalg, sysio.epoch, sysio.msgch, sysio.msig, sysio.opreg, sysio.roa, sysio.system, and sysio.uwrit. These prebuilt contracts are copied when BUILD_SYSTEM_CONTRACTS=OFF, and are also used by tests/performance flows, so deploying or executing them after this PR will fail once the host no longer provides __ashlti3. Please rebuild/update those checked-in WASMs with the new CDT/librt path, or keep the host intrinsic until all bundled contracts no longer import it.

heifner:
Addressed. The checked-in system-contract WASMs were rebuilt against the PR #51 CDT/librt path in `e681b6516a` and `8bfc006959`; none of the listed contracts (`sysio.{authex,bios,chalg,epoch,msgch,msig,opreg,roa,system,uwrit}`) import `env.__ashlti3` (or any int128/softfloat builtin) anymore.

The latest merge of master (`153e014a19`) regenerated all reference data (snapshots, consensus_blockchain, deep-mind, `head_block_id`) with that toolchain and re-verified 0 env builtin imports across every bundled contract. Pairs with wire-cdt #51.

huangminghuang (commented):
libraries/chain/genesis_intrinsics.cpp (line 5) still seeds the protocol intrinsic whitelist with the removed compiler-builtin names (__divti3, __addtf3, etc.). This PR removes their interface declarations, signature pins, and sys-vm registrations, so sys_vm_host_functions_t::resolve() will reject any module importing them before the whitelist check can help. New chains/snapshots will therefore advertise these intrinsics as whitelisted even though they are no longer resolvable. Please remove the dropped builtin names from genesis_intrinsics as part of the same change, or explain why they must remain as inert state.

huangminghuang (commented):
cmake/SysioTester.cmake.in (line 163) still requires libbuiltins and links it into SysioChain (line 191), but this PR deletes libraries/builtins and stops building/installing that archive. Any downstream project using the installed sysio CMake package will fail at find_package(sysio) / include(SysioTester) because find_sysio_library(libbuiltins ...) can no longer resolve. Remove libbuiltins from SysioTester.cmake.in along with the in-tree target removal.

huangminghuang (approved):

PR #319, opened 2026-04-28
Copyable wrapper over std::move_only_function for next_function

heifner:
## Why this change

This is prep work for a streaming/move-response pivot on the async APIs (`push_transaction`, `send_transaction`, `compute_transaction`, the read-only path, etc.). Today those callbacks are stored in a `std::function`, which forces every wrapped lambda to be copy-constructible. As soon as the response path wants to *move* the result into a streaming writer (instead of copying or going through a serialize-then-send detour) the lambda needs to capture by `std::move` and consume those captures on call - and `std::function` cannot hold such a callable.

Switching the type erasure to `std::move_only_function` fixes that, but as a bare alias it does not compile against the existing pipelines (boost::signals2, boost::multi_index, several `[=]` and `CATCH_AND_CALL` capture sites all require the outer type to be copyable). Wrapping `move_only_function` behind a shared_ptr gives the storage-side win - mutable consume-on-call captures - without rewriting the surrounding plumbing. None of the consumer sites need to change.

Doing this on master rather than on the streaming branch keeps the type change isolated and benefits any other consumer of these async APIs.

## Summary

`next_function<T>` becomes a small class holding `shared_ptr<std::move_only_function<...>>` instead of a `std::function` alias. The user-visible API matches the old alias: default-constructible, null-checkable via `operator bool`, throws `std::bad_function_call` when invoked empty.

## Why a wrapper instead of pure std::move_only_function

A direct alias swap to `std::move_only_function` was the first thing tried. The callback travels through several pipelines that need a copyable type:

- `boost::signals2`, used by `appbase::method_decl` for `incoming::methods::transaction_async` (the dispatch backing `chain_plugin::accept_transaction` and the `read_write::push_transaction` pipeline). Slot storage is `boost::function`, which copies its bound callable when forwarding through the signal pipeline. Tested with parameter forms of value, `&`, and `&&` - all fail at slot-dispatch time.
- `boost::multi_index` value storage in `unapplied_transaction_queue`, in `trx_retry_db`'s `tracked_transaction`, and in `snapshot_scheduler`'s `pending_snapshot_index`. Insertion paths use `insert(value_type)` with a default-copyable expectation.
- Many lambda captures across `chain_plugin` and `producer_plugin` that capture by `[=]` or by value, plus paired `CATCH_AND_CALL(next)` fallbacks that need access to `next` after it has been captured into an async closure.

The shared_ptr indirection keeps the outer type copyable while leaving the inner `move_only_function` move-only, so it accepts non-copyable callables. All copies of a given `next_function` share the same underlying callable.

## Performance

Per callback lifecycle:

| Phase | Before (`std::function`) | After (wrapper) |
|---|---|---|
| Construct from a lambda | 0 allocs if it fits SBO, 1 alloc otherwise | Always 1 alloc (`make_shared`) |
| Copy | Full alloc when heap-stored, byte copy when SBO | Atomic refcount bump |
| Move | Pointer swap or byte copy | Pointer swap |
| Invocation | One indirect call | Two indirect calls |
| Sizeof | ~32B (pointer + SBO buffer + manager) | 16B |

Net per request (typical: construct, 2-4 copies, one invocation): the construction cost shifts from inline-SBO to one allocation, and copies get cheaper. Multi-copy paths like `push_recurse`, the snapshot_scheduler fanout, and the read_only_trx executor chain see a real win on copies. End-to-end through `chain_plugin::push_transaction`, the wrapper cost is dominated by signature recovery, ABI serialization, and transaction execution.

## Semantic shift to flag for review

`std::function` copies are independent. Wrapper copies share state. For async-callback usage that's the desired model (one logical callback held by several lambdas in the same pipeline), and survey of the call sites did not find any code that depended on the old independence semantics. Worth a second look during review against future call sites.

### DIFF
--- a/plugins/producer_plugin/src/producer_plugin.cpp
+++ b/plugins/producer_plugin/src/producer_plugin.cpp
@@ -1061,16 +1061,17 @@ public:
 
       auto is_transient = (trx_type == transaction_metadata::trx_type::read_only || trx_type == transaction_metadata::trx_type::dry_run);
       if (!is_transient) {
-         next = [this, trx, next{std::move(next)}](const next_function_variant<transaction_trace_ptr>& response) {
-            next(response);
-
+         next = [this, trx, next{std::move(next)}](next_function_variant<transaction_trace_ptr>&& response) {
+            // Extract the ack metadata (cheap shared_ptr copies) before moving the
+            // variant into the inner next; that keeps both consumers correct under
+            // the rvalue-only contract without an extra variant copy.
             fc::exception_ptr except_ptr; // rejected
             if (std::holds_alternative<fc::exception_ptr>(response)) {
                except_ptr = std::get<fc::exception_ptr>(response);
             } else if (std::get<transaction_trace_ptr>(response)->except) {
                except_ptr = std::get<transaction_trace_ptr>(response)->except->dynamic_copy_exception();
             }
-
+            next(std::move(response));
             _transaction_ack_channel.publish(priority::low, std::pair<fc::exception_ptr, packed_transaction_ptr>(except_ptr, trx));
          };
       }
--- a/plugins/producer_plugin/src/producer_plugin.cpp
+++ b/plugins/producer_plugin/src/producer_plugin.cpp
@@ -1062,17 +1062,16 @@ public:
       auto is_transient = (trx_type == transaction_metadata::trx_type::read_only || trx_type == transaction_metadata::trx_type::dry_run);
       if (!is_transient) {
          next = [this, trx, next{std::move(next)}](next_function_variant<transaction_trace_ptr>&& response) {
-            // Extract the ack metadata (cheap shared_ptr copies) before moving the
-            // variant into the inner next; that keeps both consumers correct under
-            // the rvalue-only contract without an extra variant copy.
+            // Publish before invoking next so the one-shot callback is the wrapper's last act; a throw from publish
+            // would otherwise re-enter next via the outer CATCH_AND_CALL after captures were already consumed.
             fc::exception_ptr except_ptr; // rejected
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/src/producer_plugin.cpp
```diff
@@ -1061,16 +1061,17 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+            // variant into the inner next; that keeps both consumers correct under
+            // the rvalue-only contract without an extra variant copy.
             fc::exception_ptr except_ptr; // rejected
             if (std::holds_alternative<fc::exception_ptr>(response)) {
                except_ptr = std::get<fc::exception_ptr>(response);
             } else if (std::get<transaction_trace_ptr>(response)->except) {
                except_ptr = std::get<transaction_trace_ptr>(response)->except->dynamic_copy_exception();
             }
-
+            next(std::move(response));
```
huangminghuang: The wrapper invokes the original next and then calls _transaction_ack_channel.publish(...). If publish or a subscriber path throws, process_incoming_transaction_async() catches it at line 1172 and calls the same outer next again. With the new shared move_only_function backing, that can re-enter a callback whose captures were already consumed by the first invocation. I’d move the publish before next(std::move(response)), or isolate/log publish failures so no work that can throw runs after the one-shot callback.

heifner: Good catch -- fixed in 13b1e24635. Reordered so publish runs first and the inner next(std::move(response)) is the wrapper's last act. If publish throws synchronously, next is never invoked here, and CATCH_AND_CALL(next) at the call site invokes it exactly once with the exception, satisfying the one-shot contract.

huangminghuang (approved):

PR #321, opened 2026-04-30
Use wire-vcpkg-registry boringssl-custom 1.0.3 (drops libssl-dev runtime dep)

heifner:
Picks up Wire-Network/wire-vcpkg-registry#11 which adds `libssl.pc` and `libcrypto.pc` to the boringssl-custom port.

With those .pc files installed, vcpkg curl's `dependencies.patch` (`pkg_check_modules(REQUIRED libssl)` / `pkg_check_modules(REQUIRED libcrypto)`) resolves to our boringssl-custom static archives instead of falling back to a system `libssl-dev` install. Removes a latent system dependency that surfaced in CICD when curl's binary cache was invalidated by the prior boringssl-custom 1.0.1 -> 1.0.2 bump.

### DIFF
--- a/vcpkg-configuration.json
+++ b/vcpkg-configuration.json
@@ -9,7 +9,7 @@
     {
       "kind": "git",
       "repository": "https://github.com/wire-network/wire-vcpkg-registry",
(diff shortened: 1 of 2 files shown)


jglanz (approved):
awesome

PR #323, opened 2026-05-01
Block production diagnostic logging

heifner:
## Summary

Five operator-facing diagnostic logging improvements that surfaced as gaps during the recent Vaulta BP investigations (the always-11-orphaned-blocks pattern). All small, low-volume, and structured for grepping. Independent of the producer-side fixes in #320 and #322 -- this PR is just diagnostics, no behavior changes to block validation, application, or fork choice.

## What you get

1. **wire latency** added to "Received block" log lines. Splits the existing combined `latency` (block_timestamp -> apply complete) into a wire-side number (block_timestamp -> first network arrival, set by net_plugin) and the existing latency. The difference is the local apply-queue / in-producing-mode delay. In the Vaulta incidents we spent hours chasing a "9-second wire gap" that was actually a ~1-second wire gap plus 8 seconds of local hold; this would have surfaced that immediately. Existing `latency` field unchanged.

2. **Entering producing round**. One info line at the start of each producing round, with head, fork_db_head, and the gap. If the gap is non-zero at round start, the producer is entering its slot with un-applied blocks queued in fork_db -- the exact condition that triggered the always-11 collapse. New-round detection compares head.producer() to scheduled_producer.

3. **Round complete**. Companion to (2): one info line at round transition summarizing the just-finished round -- producer, starting block num, signed count, finalized count, orphaned count. Surfaces "always 11 missed" patterns directly without operators having to hand-correlate produce_block logs against fork-switch logs. Survival check is via `head.extends(...)`, valid at next-round-start (sub-second window where a fork-switch could still shift counts by a block or two).

4. **Reformatted clock-offset log** at over-threshold time-message asymmetry. Old form printed raw NTP arithmetic (`rec`/`org`/`xmt`/`dst` nanoseconds plus an `offset` integer), which operators rationally read as "clock drift" and dismissed. New form expresses the same data as outbound vs inbound one-way latency plus the offset. Same trigger threshold, same volume, much more actionable.

5. **block_notice mode transitions**. One info line when a peer crosses the `consecutive_block_nacks` threshold (we've stopped sending them full blocks, only notices) and one when they fall back. Helps operators understand when peers are normally faster than them (and therefore become slow request-respond fallback paths when the usual fast path fails).

Plus a small helper `block_handle::extends(block_id_type)` shared by (3); same as the helper added in #322.

### DIFF
--- a/plugins/producer_plugin/src/producer_plugin.cpp
+++ b/plugins/producer_plugin/src/producer_plugin.cpp
@@ -710,8 +710,7 @@ public:
    std::set<chain::account_name>                     _producers;
    chain::db_read_mode                               _db_read_mode = db_read_mode::HEAD;
    pending_block_mode                                _pending_block_mode = pending_block_mode::speculating;
-   block_timestamp_type                              _last_round_entry_logged{};                      // dedup the round-entry log line to once per producing round
-   // Tracks blocks signed during the current producing round so we can summarize survived-vs-orphaned at round exit.
+   // Tracks blocks signed during the current producing round so we can summarize on-head-vs-orphaned at round exit.
    struct producing_round_state {
       account_name                producer;
       block_timestamp_type        round_start;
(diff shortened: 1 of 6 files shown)


plugins/producer_plugin/src/producer_plugin.cpp
```diff
@@ -2265,6 +2274,35 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
    _time_tracker.clear(); // make sure we start tracking block time after `apply_blocks()`
 
+   // Round transition diagnostics: summarize the previous round (if any) when the scheduled producer differs from the
+   // tracked producer, then log round entry. Survival check uses head.extends, valid for blocks still above LIB which
+   // is true at next-round-start; sub-second fork switches after round end could shift the count by a block or two.
+   if (_producing_round && _producing_round->producer != scheduled_producer.producer_name) {
+      // signed_blocks form one chain in the common case (no mid-round fork switch); iterate reverse and break on the
+      // first canonical id, since all earlier signed blocks are then ancestors and necessarily canonical too.
+      size_t survived = 0;
+      for (auto it = _producing_round->signed_blocks.rbegin(); it != _producing_round->signed_blocks.rend(); ++it) {
```
huangminghuang: counts earlier signed blocks as survived once it finds the latest signed block on the current head branch. That only works if every block signed during the round stayed on one chain. If the producer signs on branch A, then a mid-round fork switch occurs and it signs on branch B, the reverse search can find the branch-B block and count the earlier branch-A blocks as “finalized/survived” even though they were orphaned. This is exactly the kind of forky case these diagnostics are meant to clarify. Safer fix: check each signed id independently against head.id() || head.extends(id), or reset the round tracker when the producer switches branches.

plugins/producer_plugin/src/producer_plugin.cpp
```diff
@@ -2265,6 +2274,35 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
+            break;
+         }
+      }
+      const auto signed_count = _producing_round->signed_blocks.size();
+      ilog("Round complete for {} starting #{} at {}: signed {}, finalized {}, orphaned {}",
+           _producing_round->producer, _producing_round->first_block_num, _producing_round->round_start,
+           signed_count, survived, signed_count - survived);
+      _producing_round.reset();
+   }
+   if (in_producing_mode() && head.producer() != scheduled_producer.producer_name && _last_round_entry_logged != block_time) {
```
huangminghuang: only starts a producing-round record when head.producer() != scheduled_producer.producer_name. On a single-producer network, or any schedule where the same account continues across a round boundary, this suppresses both the “Entering producing round” log and the later summary. If the goal is round diagnostics, keying this off the round boundary/block number is more reliable than comparing current head producer to scheduled producer.

plugins/producer_plugin/src/producer_plugin.cpp
```diff
@@ -2265,6 +2274,35 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
+      // first canonical id, since all earlier signed blocks are then ancestors and necessarily canonical too.
+      size_t survived = 0;
+      for (auto it = _producing_round->signed_blocks.rbegin(); it != _producing_round->signed_blocks.rend(); ++it) {
+         if (head.id() == *it || head.extends(*it)) {
+            survived = static_cast<size_t>(std::distance(it, _producing_round->signed_blocks.rend()));
+            break;
+         }
+      }
+      const auto signed_count = _producing_round->signed_blocks.size();
+      ilog("Round complete for {} starting #{} at {}: signed {}, finalized {}, orphaned {}",
```
huangminghuang: logs the survived count as finalized. The code is checking whether signed blocks are on the current head branch, not whether they are irreversible/final. In this codebase “finalized” has a specific consensus meaning, so this can mislead operators reading the new diagnostics. Consider renaming the field to survived or canonical.

plugins/producer_plugin/src/producer_plugin.cpp
heifner: Fixed in 4ad3b543e4. Switched to a per-id `head.id() == id || head.extends(id)` loop over signed_blocks, so a mid-round branch switch can no longer let earlier branch-A blocks be counted as on-head once the producer fork-switches to branch B. The reverse-iter shortcut was a premature optimization for a 12-iteration loop.

plugins/producer_plugin/src/producer_plugin.cpp
heifner: Fixed in 4ad3b543e4. Both transitions now hang off the round-start slot (`block_timestamp_type(block_time).slot % chain::config::producer_repetitions == 0`), the same idiom used a few lines up at line 2174. Single-producer schedules and same-account-across-boundary rotations now log round-entry and round-complete normally. `_last_round_entry_logged` was redundant once the slot edge plus the `_producing_round` optional dedup the calls, so it's gone.

plugins/producer_plugin/src/producer_plugin.cpp
heifner: Renamed to `on head` in 4ad3b543e4 (`signed S, on head H, orphaned O`). Agreed that `finalized` collided with the consensus meaning.

huangminghuang (approved):

huangminghuang (approved):

PR #324, opened 2026-05-06
Migrate boost::asio::deadline_timer to chrono-based timers

heifner:
Replaces deprecated `boost::asio::deadline_timer` with `boost::asio::steady_timer` (interval / backoff sites) or `boost::asio::system_timer` (wall-clock deadline at the http_client API boundary).

`deadline_timer` is deprecated in Boost 1.87 and removed from the `<boost/asio.hpp>` umbrella include in 1.89; the next Boost bump past 1.89 would break the build. Side benefit: chrono-native at every site (drops Boost.DateTime dependency from these sources), and steady-clock semantics at backoff/poll sites avoid NTP/wall-clock-jump sensitivity.

Three commits, by category:

1. `cleanup: drop dead boost::asio::deadline_timer surfaces` — `accept_error_timer` member in net_plugin (declared but never read/written) and `mock_deadline_timer` typedef in fc::mock_time (zero callers, kept alive only by supporting asio TimeTraits methods on `mock_time_traits` which nothing instantiates). `mock_time_traits::set_now/fc_now/is_set` are kept; `boost::posix_time::ptime` remains the public `time_type` for the existing `test_trx_retry_db.cpp` callers.

2. `migrate periodic poll/backoff timers to boost::asio::steady_timer` — wallet_plugin lock-file watch (1 s poll), resource_monitor disk-space poll, and fc::listener accept-error backoff after EMFILE. The listener's `accept_timeout` parameter type changes from `boost::posix_time::time_duration` to `std::chrono::milliseconds`, cascading through three callers of `fc::create_listener` (http_plugin, net_plugin, state_history_plugin).

3. `migrate http_client deadline to boost::asio::system_timer` — the `sync_do_with_deadline` deadline derives from an `fc::time_point` at the public `post_sync()` boundary, so it carries wall-clock semantics. Side effect: collapses the `gregorian::date(1970, 1, 1) + posix_time::microseconds(...)` epoch-arithmetic dance to a single `std::chrono::system_clock::time_point` construction.

Ports Spring open issue [#1830](https://github.com/AntelopeIO/spring/issues/1830).


libraries/libfc/src/network/http/http_client.cpp
```diff
@@ -248,8 +248,7 @@ class http_client_impl {
    };
 
    variant post_sync(const url& dest, const variant& payload, const fc::time_point& _deadline) {
-      static const deadline_type epoch(boost::gregorian::date(1970, 1, 1));
-      auto deadline = epoch + boost::posix_time::microseconds(_deadline.time_since_epoch().count());
+      auto deadline = deadline_type(std::chrono::microseconds(_deadline.time_since_epoch().count()));
```
huangminghuang: post_sync() default deadline is fc::time_point::maximum(). Converting its int64_t microsecond count directly into std::chrono::system_clock::time_point can overflow/wrap on platforms where system_clock::duration is nanoseconds, e.g. libstdc++ Linux. That makes default/no-timeout HTTP calls look expired immediately. Fix: special-case fc::time_point::maximum() to deadline_type::max() before constructing from microseconds.

heifner: Fixed in 4c68844fee. Added a saturating `fc::time_point::to_system_clock()` helper that maps `maximum()` to `system_clock::time_point::max()` and otherwise does the natural microsecond conversion, and routed the three sibling sites in `producer_plugin.cpp` through it as well so the overflow can't resurface.

libraries/libfc/include/fc/time.hpp
```diff
@@ -69,6 +70,22 @@ namespace fc {
...
+         * Saturating conversion to std::chrono::system_clock::time_point.
+         *
+         * fc::time_point::maximum() carries microseconds(INT64_MAX); a direct construction of
+         * system_clock::time_point from that value silently overflows on platforms whose
+         * system_clock::duration is finer than microseconds (e.g. nanoseconds on libstdc++ Linux),
+         * which would yield a deadline in the past. This helper maps maximum() to
+         * system_clock::time_point::max() and otherwise performs the natural microsecond conversion.
+         */
+        std::chrono::system_clock::time_point to_system_clock() const {
+           if (*this == maximum())
```
huangminghuang: to_system_clock() says it is saturating, but only clamps exact fc::time_point::maximum(). Any other far-future value, like maximum() - fc::microseconds(1), can still overflow when converted to a nanosecond system_clock::time_point on libstdc++, producing a bogus past deadline. Since this helper is now the public safety point, it should clamp any value outside system_clock::time_point::{min,max}() range, not just exact maximum().

heifner: Fixed in https://github.com/Wire-Network/wire-sysio/pull/324/commits/e950e56f622715435722e7e70cbd6956446fa8a4

huangminghuang (approved):

PR #335, opened 2026-05-14
Use modular Boost vcpkg ports

huangminghuang:
## Summary
- replace the all-in-one `boost` vcpkg dependency with the specific modular Boost ports used by wire-sysio
- remove the obsolete `boost` override now that individual Boost ports resolve from the default vcpkg registry
- keep static Boost linkage via the existing CMake `Boost_USE_STATIC_LIBS` setup

## Verification
- `cmake --build /home/ssm-user/code/wire/wire-sysio/build/boost-modular-pruned-debug-clang -j2`
- confirmed the fresh vcpkg tree contains `libboost_*.a` archives and no `libboost_*.so` files


vcpkg.json
```diff
@@ -74,10 +105,6 @@
       "name": "protobuf",
       "version": "6.33.4#1"
     },
-    {
-      "name": "boost",
-      "version": "1.89.0#1"
```
heifner, line 79: We want to pin the version. How is that done with this change? Does the `wire-vcpkg-registry` boost port go away with this? If so we should remove that and update the reference in `vcpkg-configuration.json` as part of this same PR.

vcpkg.json
```diff
@@ -58,7 +58,38 @@
...
+    "boost-multiprecision",
+    "boost-process",
+    "boost-program-options",
+    "boost-property-tree",
+    "boost-rational",
+    "boost-regex",
+    "boost-signals2",
+    "boost-system",
+    "boost-test",
+    "boost-thread",
```
heifner, line 94: Boost::algorithm and Boost::lexical_cast are linked but not declared. libraries/chaindb/CMakeLists.txt:12 links Boost::lexical_cast and tests/CMakeLists.txt:142 links
  Boost::algorithm, but neither boost-algorithm nor boost-lexical-cast is in the new modular list, and neither is in dependencies.boost.cmake's BOOST_COMPONENTS loop. They work
  today via transitive find_dependency chains in upstream Boost CMake configs. A future baseline bump could break this silently at link time rather than at configure time. Adding
   both as explicit deps would be cheap insurance.

heifner (changes requested):

vcpkg.json
huangminghuang: Good catch. With the modular ports, Boost is currently pinned indirectly by the default vcpkg registry baseline, but that is less explicit than the old `boost` override.

I’ll update this PR to make the pin explicit for the modular Boost packages, and I’ll remove `boost` from `vcpkg-configuration.json` since wire-sysio no longer consumes the custom `wire-vcpkg-registry` Boost port. Removing the actual Boost port from `wire-vcpkg-registry` would need to happen in that registry repo, unless we know nothing else still uses it.

vcpkg.json
huangminghuang: Agreed. I’ll add both `boost-algorithm` and `boost-lexical-cast` as explicit vcpkg dependencies, and include them in the Boost discovery list so we fail at configure time if either target is missing. That avoids relying on transitive Boost CMake behavior from the current baseline.

heifner (approved):

PR #337, opened 2026-05-14
Base58: replace OpenSSL BIGNUM with pure-bytes encode/decode

heifner:
## Summary
Heaptrack on a node serving `/v1/trace_api/get_block` (5k TPS, 30s) flags base58-encoding of signatures as the dominant allocator -- ~5.7M `OPENSSL_malloc` calls through `BN_rshift` / `BN_div` inside `EncodeBase58`. The libfc port of base58 is a 2012-era Bitcoin Core fork that still walks the input through an OpenSSL BIGNUM, allocating per `BN_new`, per `BN_CTX_new`, and per internal `BN_div` step.

This PR replaces the BIGNUM path with Bitcoin Core's pure-bytes algorithm (repeated `divmod 58` on a working byte buffer). The algorithm is mathematically identical -- both compute the base-58 representation of the input treated as a big-endian integer -- so encoded outputs are bit-for-bit unchanged.

## Changes
- `libraries/libfc/src/crypto/base58.cpp` rewritten against the pure-bytes algorithm; `CBigNum` wrapper class and the `<openssl/bn.h>` include removed.
- Working buffers use `boost::container::small_vector<unsigned char, 128>`, so signatures / public keys / private keys all encode without a heap allocation for the working state.
- Public API (`fc::to_base58`, `fc::from_base58`) is unchanged; existing callers compile and behave identically.
- New `libraries/libfc/test/crypto/test_base58.cpp` covering Bitcoin Core's 12 canonical `encode_decode.json` vectors, empty input, leading-zero preservation, invalid-character rejection, whitespace tolerance, and randomized roundtrips at 32 / 33 / 65 byte input sizes.

## Allocations per call
| Input | Before (BIGNUM) | After |
|---|---|---|
| 32-byte private key | 12+ heap allocs | 1 (output string) |
| 33-byte public key | 12+ heap allocs | 1 (output string) |
| 65-byte signature | 12+ heap allocs | 1 (output string) |

### DIFF
--- a/libraries/libfc/src/crypto/base58.cpp
+++ b/libraries/libfc/src/crypto/base58.cpp
@@ -23,19 +23,19 @@
 namespace {
 
 /// Alphabet used for base58 encoding (no 0OIl to avoid visually similar glyphs).
-constexpr const char* k_b58_alphabet =
+constexpr const char* b58_alphabet =
    "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
 
 /// Inline capacity for the working buffer used inside encode/decode.
 /// Covers any input up to ~92 bytes without touching the heap; signatures
 /// are 65 bytes, public keys 33, private keys 32, so all production callers
 /// stay on the stack. Larger inputs fall back to a heap allocation.
-constexpr size_t k_b58_inline_capacity = 128;
+constexpr size_t b58_inline_capacity = 128;
 
 /// Reverse lookup table: maps an ASCII byte to its base58 digit value, or -1
 /// if the byte is not a valid base58 character. Sized at 256 so any unsigned
 /// byte indexes safely.
-constexpr int8_t k_b58_reverse[256] = {
+constexpr int8_t b58_reverse[256] = {
    -1,-1,-1,-1,-1,-1,-1,-1, -1,-1,-1,-1,-1,-1,-1,-1,
    -1,-1,-1,-1,-1,-1,-1,-1, -1,-1,-1,-1,-1,-1,-1,-1,
    -1,-1,-1,-1,-1,-1,-1,-1, -1,-1,-1,-1,-1,-1,-1,-1,
@@ -62,18 +62,18 @@ constexpr int8_t k_b58_reverse[256] = {
 /// the partial quotient in base58. Leading zero bytes in the input are
 /// preserved as leading '1' characters in the output.
 std::string encode_base58(const unsigned char* pbegin, const unsigned char* pend) {
-   int zeroes = 0;
+   size_t zeroes = 0;
    while (pbegin != pend && *pbegin == 0) {
       ++pbegin;
       ++zeroes;
    }
    // log(256) / log(58) ~= 1.366, round up via *138/100 + 1.
-   const int size = static_cast<int>((pend - pbegin) * 138 / 100 + 1);
-   boost::container::small_vector<unsigned char, k_b58_inline_capacity> b58(size);
-   int length = 0;
+   const size_t size = static_cast<size_t>(pend - pbegin) * 138 / 100 + 1;
+   boost::container::small_vector<unsigned char, b58_inline_capacity> b58(size);
+   size_t length = 0;
    while (pbegin != pend) {
       int carry = *pbegin;
-      int i = 0;
+      size_t i = 0;
       // b58 = b58 * 256 + carry, walking from least to most significant digit.
       for (auto it = b58.rbegin(); (carry != 0 || i < length) && it != b58.rend(); ++it, ++i) {
          carry += 256 * (*it);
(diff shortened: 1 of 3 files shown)


libraries/libfc/src/crypto/base58.cpp
```diff
@@ -1,650 +1,188 @@
...
+      ++zeroes;
+   }
+   // log(256) / log(58) ~= 1.366, round up via *138/100 + 1.
+   const size_t size = static_cast<size_t>(pend - pbegin) * 138 / 100 + 1;
+   boost::container::small_vector<unsigned char, b58_inline_capacity> b58(size);
+   size_t length = 0;
+   size_t outer = 0;
+   while (pbegin != pend) {
+      if ((++outer & 0x3F) == 0) // every 64 input bytes
+         yield();
```
huangminghuang, line 90: This regresses yield semantics. The new encoder skips all leading zero bytes before the only yield site, then only calls yield() every 64 non-zero input bytes. That means empty input, 32-byte private keys, 33-byte public keys, and all-zero or mostly-leading-zero large inputs can complete without checking an expired deadline at all. The old implementation called yield() at entry/exit and inside the conversion loop, so callers like ABI serialization that pass deadline-checking yields could rely on to_base58 honoring cancellation even for small key encodes. I’d add at least one unconditional yield() near function entry, and probably another after leading-zero handling or before return, plus a unit test that a small input invokes the yield.

heifner: Restored in 3ab9954dc2. `encode_base58` now calls `yield()` unconditionally on entry (before the leading-zero skip) and again after the conversion loop, keeping the periodic in-loop yield for large inputs. The entry yield covers empty / small-key / all-zero inputs that never reach the in-loop site; the exit yield catches a deadline that expired at a length that is not a multiple of the cadence. Added a test (`to_base58_invokes_yield_even_for_small_inputs`) asserting that 32-byte, 65-byte, empty, and all-zero inputs each invoke the yield, and that a 256-byte input invokes it more often than a 32-byte one.

heifner:
Done in 3ab9954dc2. `encode_base58` now takes `const char*` and uses `static_cast<unsigned char>(*pbegin)` for the byte arithmetic, so `to_base58` forwards without a pointer cast. `from_base58` builds its result with the value-based range constructor `std::vector<char>(out.begin(), out.end())`. Both `reinterpret_cast`s are gone.

libraries/libfc/src/crypto/base58.cpp
```diff
@@ -1,650 +1,200 @@
...
-    for (const char* p = psz; *p == pszBase58[0]; p++)
-        nLeadingZeros++;
-    vchRet.assign(nLeadingZeros + vchTmp.size(), 0);
+namespace fc {
 
-    // Convert little endian data to big endian
-    reverse_copy(vchTmp.begin(), vchTmp.end(), vchRet.end() - vchTmp.size());
-    return true;
+std::string to_base58(const char* d, size_t s, const fc::yield_function_t& yield) {
+   return encode_base58(d, d + s, yield);
```
huangminghuang, line 180: It still forms d + s unconditionally, and the new tests now call fc::to_base58(nullptr, 0, ...) at libraries/libfc/test/crypto/test_base58.cpp:45. Pointer arithmetic on a null pointer is undefined behavior even when adding zero, so the test is exercising UB rather than proving empty/null input is supported. Same risk appears in the empty Bitcoin vector path via bytes.data(), bytes.size(). Please special-case s == 0 before doing pointer arithmetic, e.g. call encode_base58(d, d, yield) for zero length, and assert/reject d == nullptr && s != 0.

heifner: Addressed in `ff5be1ee9a`. I used an early `return std::string()` for the zero-length case rather than `encode_base58(d, d, yield)`:

```cpp
std::string to_base58(const char* d, size_t s, const fc::yield_function_t& yield) {
   if (s == 0)
      return std::string();
   FC_ASSERT(d != nullptr, "to_base58 called with null data pointer and non-zero length {}", s);
   return encode_base58(d, d + s, yield);
}
```

Two reasons for returning directly instead of routing the empty range through `encode_base58`:

- It makes this overload consistent with `to_base58(const std::vector<char>&, ...)`, which already early-returns an empty string for an empty vector without calling through.
- Empty input does no encoding work, so there is nothing for a deadline-bound caller to interrupt. The entry/exit yields added in the previous round are still reached by every non-empty input: 32/33/65-byte keys and signatures, and all-zero / leading-zero inputs (those have `s != 0`, hit the unconditional entry yield, then skip the conversion loop). Only a genuinely empty input skips the yield.

The `s != 0 && d == nullptr` case is now rejected with `FC_ASSERT` rather than forming `nullptr + s`.

On the tests: the empty-input case keeps the `to_base58(nullptr, 0, ...)` call (now well-defined, since it returns before any pointer arithmetic), `to_base58_invokes_yield_even_for_small_inputs` asserts empty input yields zero times while all-zero / key / signature / large inputs still yield, and a new `null_data_with_nonzero_length_throws` covers the assert.

huangminghuang (approved):

PR #339, opened 2026-05-15
Added AWS KMS signing

brianjohnson5972:
**What the feature does**

  Adds a third signature-provider backend, KMS:, to signature_provider_manager_plugin, alongside the existing KEY: (raw private key) and KIOD: (key daemon) forms. With KMS:, the signing key lives in AWS
   KMS and never appears on the host or in process memory — nodeop/cranker calls KMSClient::Sign over the wire for each signature.

  The first concrete consumer is the Ethereum cranker, replacing a hot KEY:0x… literal in the operator's config with an IAM-gated KMS key (CloudTrail-audited, rotatable without redeploy).

The feature will not be automatically available when signature_provider_manager_plugin is added to the command line, unless it is added specifically to an executable. This feature will be used for the cranker application that is being added in a separate Pull Request. It is specifically not made available in nodeop, since AWS KMS would not sign blocks in an appropriate timeframe.

  **Spec grammar**

  --signature-provider eth-01,ethereum,ethereum,0x045a87…,KMS:<key-ref>

  <key-ref> is either a full ARN (arn:aws:kms:<region>:<account>:(key|alias)/<id>) or the shorthand <region>:<key-id-or-alias>. Region is mandatory — no silent fallback to AWS_REGION, so
  misconfiguration fails loud.

  **Implementation (6 commits, ~1,480 lines)**

  New plugin-private TU — src/kms_signature_provider.{hpp,cpp}:
  - parse_kms_spec — ARN/shorthand parser, throws plugin_config_exception on malformed input.
  - der_to_compact / normalise_low_s / recover_v — convert KMS's ASN.1 DER ECDSA output to a 65-byte Ethereum recoverable signature. KMS doesn't return a recovery id, so v is derived by trial-recovery
  against the pinned public key; high-S is normalised to low-S (EIP-2).
  - get_kms_client — process-wide per-region KMSClient cache; aws_sdk_lifecycle Meyers-singleton guards Aws::InitAPI/ShutdownAPI with deliberate static-destruction ordering.
  - make_kms_signature_provider — builds the sign_fn closure; signs with MessageType=DIGEST + ECDSA_SHA_256.

  Plugin hook — signature_provider_manager_plugin.cpp gets a KMS branch in create_provider_from_spec (returns std::nullopt for private_key, same shape as KIOD) and extended help text.

  Build wiring — aws-sdk-cpp[kms] (v1.11.665, default-features:false) added to vcpkg.json; plugin CMakeLists.txt links aws-cpp-sdk-kms. The boringssl-custom collision with aws-c-cal/s2n (both declare
  openssl) was resolved via overlay ports in wire-vcpkg-registry; baseline bumped to c739f08.

  **Scope boundaries**

  - Supported now: secp256k1 / chain_key_type_ethereum only.
  - Out of scope: Wire K1, R1/P-256, Ed25519 (Solana), BLS — all throw pending_impl_exception at construction.
  - Explicitly not for producer block signing — KMS Sign is ~30–100 ms/call and non-deterministic (not RFC-6979), so it's fine for cranker/maintenance traffic but unsafe for consensus paths.

  **Tests**

  test_kms_signature_provider.cpp (461 lines) — fully offline suite: parser cases, DER round-trips, low-S flipping, v-recovery, per-region client caching, plus an env-gated
  (KMS_LIVE_SPEC/KMS_LIVE_PUBKEY) live round-trip. test_create_provider_specs.cpp adds two end-to-end cases proving the plugin routes KMS: specs through the parser and rejects Solana KMS specs at parse
  time.


OVERVIEW.md
```diff
@@ -87,6 +87,8 @@ Elected via Appointed Proof of Stake (APoS). They validate transactions and prod
 - Building and submitting outbound envelopes to external chains
 - Responding to challenges if disputed
 
+Signing keys for outbound submissions can be held locally (`KEY:`), in a `kiod` daemon (`KIOD:`), or in AWS KMS (`KMS:` — secp256k1 asymmetric keys). See `plugins/signature_provider_manager_plugin/KMS_SIGNING_DESIGN.md` for the KMS spec format and operational notes.
```
huangminghuang: I don't see the KMS_SIGNING_DESIGN.md file in this PR.

plugins/signature_provider_manager_plugin/src/kms_signature_provider.cpp
```diff
@@ -0,0 +1,620 @@
...
+                 chain::plugin_config_exception,
+                 "KMS ARN tail must start with 'key/' or 'alias/', got \"{}\" in \"{}\"",
+                 tail, spec_data);
+      // Reject bare prefixes: `key/` or `alias/` with nothing after them.
+      SYS_ASSERT(tail.size() > tail_prefix_key.size() &&
+                 (!tail.starts_with(tail_prefix_alias) || tail.size() > tail_prefix_alias.size()),
+                 chain::plugin_config_exception,
+                 "KMS ARN tail \"{}\" has empty key/alias name", tail);
+
+      return kms_key_ref{region, tail};
```
huangminghuang: This strips full KMS ARNs down to key/<uuid> or alias/<name> before calling KMS. AWS documents valid KeyId values as key ID, key ARN, alias name, or alias ARN, not key/<uuid>. This means key ARNs will fail at first GetPublicKey/Sign, and alias ARNs lose their account identity and may resolve to a same-named alias in the caller’s account. Keep the full ARN as key_id for ARN specs, or strip only key ARNs to the bare UUID while preserving alias ARN/account semantics. Source: AWS KMS Sign docs, KeyId forms: https://docs.aws.amazon.com/cli/latest/reference/kms/sign.html

plugins/signature_provider_manager_plugin/src/signature_provider_manager_plugin.cpp
```diff
@@ -110,6 +125,25 @@ class signature_provider_manager_plugin_impl {
...
+         auto ref = sysio::sigprov::kms::parse_kms_spec(spec_data);
+         auto kms = sysio::sigprov::kms::make_kms_signature_provider(ref, key_type, public_key);
+
+         // Retain the startup probe so plugin_startup() can — when the opt-in
+         // KMS startup check is enabled — fail fast on a missing credential,
+         // bad region, absent IAM grant, or wrong pinned key, instead of
+         // discovering it deep into production on the first sign.
+         {
+            std::scoped_lock lock(_signing_providers_mutex);
+            _kms_startup_probes.push_back(std::move(kms.warm_up));
```
huangminghuang: This records the KMS warm-up probe before the provider is inserted at :400. If set_provider() then rejects a duplicate name/public key, the failed provider’s probe remains in _kms_startup_probes and plugin_startup() can later probe or fail on a KMS key that was never registered. Defer adding the probe until after successful insertion, or roll it back on failure.

libraries/libfc/include/fc/crypto/signer.hpp
```diff
@@ -19,6 +19,55 @@ namespace fc::crypto {
...
+      const auto& em_key = p.private_key->get<em::private_key_shim>();
+      return signature(signature::storage_type(em_key.sign_keccak256(digest)));
+   }
+
+   FC_ASSERT(static_cast<bool>(p.sign),
+             "signature provider has neither a local private key nor a sign function");
+
+   // Carry the keccak digest into the sha256-typed closure argument unchanged.
+   const auto digest_bytes = digest.to_char_span();
+   const signature sig     = p.sign(sha256(digest_bytes.data(), digest_bytes.size()));
```
heifner, line 59: This should not be using `sha256` as a bucket of bits. Just use a bucket of bits or add a `sign_keccak256`.

libraries/libfc/include/fc/crypto/signer.hpp
```diff
@@ -19,6 +19,55 @@ namespace fc::crypto {
...
+
+   // Self-verify: the remote signer must have raw-signed `digest` with the
+   // provider's key. Recover and compare; reject anything else loudly.
+   FC_ASSERT(sig.contains<em::signature_shim>(),
+             "remote signer returned a non-Ethereum signature");
+   const auto recovered = public_key(public_key::storage_type(
+      sig.get<em::signature_shim>().recover_eth(digest)));
+   FC_ASSERT(recovered == p.public_key,
+             "remote signer produced a signature that does not recover to the "
+             "provider's public key — the signer's key or digest framing is wrong");
```
heifner: No need to recover the key on every sign. Just have a test that verifies we can recover they key from a AWS signature.

plugins/signature_provider_manager_plugin/kms/src/kms_signature_provider.cpp
```diff
@@ -0,0 +1,620 @@
...
+      // Build a Sign request. MessageType=DIGEST tells KMS the 32 bytes are
+      // already a hash; otherwise it would re-hash with SHA-256 and break
+      // any chain that hashes with anything other than SHA-256.
+      Aws::KMS::Model::SignRequest req;
+      req.SetKeyId(Aws::String{state->key_id.c_str()});
+      req.SetMessage(Aws::Utils::ByteBuffer{
+         digest.to_uint8_span().data(),
+         digest.to_uint8_span().size()});
+      req.SetMessageType(Aws::KMS::Model::MessageType::DIGEST);
+      req.SetSigningAlgorithm(Aws::KMS::Model::SigningAlgorithmSpec::ECDSA_SHA_256);
```
heifner, line 660: The AWS interface is rather strange here. They should have a ECDSA_KECCAK_256 but apparently they don't. But we shouldn't duplicate that strangeness. Make the comment above a bit more clear that we are only using the provided AWS interface and the apparently normal workaround by using `ECDSA_SHA_256` as a way to say here is a 256 bit hash of data, don't rehash it.

plugins/signature_provider_manager_plugin/CMakeLists.txt
```diff
@@ -10,6 +12,7 @@ plugin_target(
 
         LIBRARIES
         custom_appbase
-        http_client_plugin 
-        wallet_plugin_headers)
+        http_client_plugin
+        wallet_plugin_headers
+        aws-cpp-sdk-kms)
```
heifner: I would prefer this not to be linked into every `nodeop` instance. Can we factor the kms_signature_provider into its own plugin and only link it into the non-`nodeop` executables.

OVERVIEW.md
brianjohnson5972: yeah, I'll pull that out and replace it with the README under the test directory.

huangminghuang (approved):

plugins/signature_provider_manager_plugin/kms/src/kms_signature_provider.cpp
```diff
@@ -0,0 +1,733 @@
...
+      // does NOT apply SHA-256 to the input. It treats the 32 bytes as an
+      // already-hashed value and signs them via raw ECDSA over secp256k1; the
+      // "SHA_256" portion of the spec only constrains the input length to 32
+      // bytes (the SHA-256 output size), which Ethereum's Keccak-256 digest
+      // also satisfies. ECDSA's math is hash-agnostic -- it reduces the
+      // 32-byte digest modulo the curve order and emits (r, s) -- so the
+      // signature KMS produces over `keccak256(rlp(tx))` is byte-identical to
+      // what a local secp256k1 signer would produce over the same 32 bytes.
+      // The self-verify in `em_sign_keccak` (recover + compare to the pinned
+      // public key) catches any deviation on every signature.
```
heifner: The self-verify in `em_sign_keccak` is gated to once-per-provider (`signature_provider_t::self_verified`), so it does not run "on every signature". The per-sign guarantee is `recover_v` below, which throws on a key mismatch on every call.

```suggestion
      // `recover_v` below recovers the signer from each signature and rejects any
      // that does not match the pinned key, so a deviation is caught on every
      // sign; the once-per-provider self-verify in `em_sign_keccak` is a separate,
      // structural check that the keccak digest reached the closure intact.
```

plugins/signature_provider_manager_plugin/kms/src/kms_signature_provider.cpp
```diff
@@ -0,0 +1,733 @@
...
+      const std::span<const unsigned char> der{
+         der_buf.GetUnderlyingData(), der_buf.GetLength()};
+
+      // `der_to_eth_signature` -> `recover_v` recovers the pubkey to pick the
+      // recovery id and throws if it doesn't match the pinned key. `em_sign_keccak`
+      // (the caller) then recovers a SECOND time and compares against the
+      // provider's public key. Two EC recoveries per signature is intentional
+      // defense-in-depth, not an oversight: this is the external-chain submission
+      // path, not a hot loop, so the cost is irrelevant -- do not "optimize" either
+      // recovery away.
```
heifner: It is two recoveries on the *first* sign, one on every sign after (the self-verify is gated once-per-provider). Worth fixing precisely because this comment says "do not optimize either recovery away" -- a maintainer reconciling code to comment could make the self-verify run every sign and undo the once-per-provider gate.

```suggestion
      // `der_to_eth_signature` -> `recover_v` recovers the signer to pick the
      // recovery id AND throws if it does not match the pinned key. That runs on
      // every sign and is the per-signature guarantee that KMS signed with the
      // expected key -- keep it. `em_sign_keccak` (the caller) recovers once more,
      // but only on the first sign per provider (gated by
      // `signature_provider_t::self_verified`): a one-time structural check that
      // the keccak digest survived the `sha256`-typed closure boundary. So the
      // first sign does two recoveries and every later sign one; this is the
      // external-chain submission path, not a hot loop, so neither cost matters.
```

libraries/libfc/include/fc/crypto/signature_provider.hpp
```diff
@@ -2,10 +2,40 @@
...
+//        deliberately constructs a `sha256` from a keccak digest's bytes and hands
+//        it to this closure. What keeps that safe is the self-verify in
+//        `em_sign_keccak`: it recovers the public key from the produced signature
+//        and asserts it matches the provider's pinned key, so a closure wired to
+//        the wrong bytes (wrong digest, framing, or key) is rejected before its
+//        output is used. That recover-and-compare is the compensating control for
+//        the lost compile-time digest distinction -- DO NOT remove it as
+//        "redundant." It is distinct from the KMS provider's own `recover_v`,
+//        which only picks the recovery id inside the closure; this validates, from
+//        outside, that the returned signature recovers to the pinned key.
```
heifner: Minor / optional, for consistency with the two `kms_signature_provider.cpp` comments: `recover_v` does not *only* pick the recovery id -- it also rejects a signature that does not recover to the pinned key, on every sign.

```suggestion
//        "redundant." It is distinct from the KMS provider's own `recover_v`,
//        which runs inside the closure on every sign -- it picks the recovery id
//        and also rejects a signature that does not recover to the pinned key;
//        this self-verify instead runs once per provider and validates, from
//        outside the closure, that the digest crossed the `sha256`-typed boundary
//        uncorrupted.
```

heifner (commented):
A few follow-ups from re-reviewing the latest revision -- all about the once-per-provider self-verify and the magic_enum range. The KMS crypto path, the plugin split, and the offline test coverage look solid; these are comment-accuracy fixes plus one scoping nit, no functional change requested.

plugins/signature_provider_manager_plugin/kms/src/kms_signature_provider.cpp
```diff
@@ -0,0 +1,733 @@
...
+// AWS KMS's service-specific error codes (Aws::KMS::KMSErrors) begin at
+// Aws::Client::CoreErrors::SERVICE_EXTENSION_START_RANGE + 1 = 129 and run up
+// to 176, so under the default ceiling magic_enum::enum_name() would return ""
+// for every KMS-specific error in throw_kms_error() -- precisely the transient
+// vs. permanent triage cases that matter most. Raise the ceiling to 256 (well
+// above the highest KMS error value, 176). This must be defined before
+// <magic_enum/magic_enum.hpp> is first included anywhere in this translation
+// unit -- including transitively through the headers below -- so it sits ahead
+// of every #include.
+#define MAGIC_ENUM_RANGE_MAX 256
```
heifner: `#define MAGIC_ENUM_RANGE_MAX 256` widens the range for *every* enum named via magic_enum in this TU, including chain enums pulled in transitively through the headers below. If one of those is also named under the default range in another TU, that is an ODR mismatch (benign today since chain enums are < 128, but latent UB). magic_enum's own header recommends a per-type specialization for exactly this case.

Suggest dropping this macro and its comment block (lines 1-13) and adding, after the includes and before `namespace sysio::sigprov::kms`:

```cpp
// KMSErrors codes begin at 129 (SERVICE_EXTENSION_START_RANGE + 1) and run to
// 176, above magic_enum's default ceiling of 127, so enum_name() returns "" for
// them without a wider range. Scope the widening to this one enum rather than the
// global MAGIC_ENUM_RANGE_MAX macro, which would affect every enum in the TU
// (incl. transitively-included chain enums) and risk an ODR mismatch with other
// TUs at the default range.
template<>
struct magic_enum::customize::enum_range<Aws::KMS::KMSErrors> {
   static constexpr int min = 0;
   static constexpr int max = 256;
};
```

heifner (approved):

PR #352, opened 2026-05-22
Unify sysio::name and slug_name onto a shared basic_name<Traits> template

heifner:
## Change Description

`name` and `slug_name` are near-duplicate 64-bit packed-identifier types that differ only in alphabet and length. This introduces `fc::basic_name<Traits>` holding the shared encoding, comparisons, and value semantics, and migrates both types onto it.

- `sysio::chain::name` now derives `fc::basic_name<sysio_name_traits>`, keeping `prefix()`/`suffix()`, `_n`, `string_to_name`, `std::hash`, and the variant conversions. Encoding is byte-identical.
- `fc::slug_name` becomes a `using` alias of `basic_name<slug_name_traits>` and switches from LSB-first to MSB-first packing, so integer order matches string order. **This changes the `slug_name` wire format** — confirm no persisted `slug_name` data exists before merging. The `sysio.opp.common` contract `slug_name` switches to MSB-first to match.
- `_n` and `_s` literals are now compile-time validated against the alphabet and length. This caught six literals that silently encoded `_` as `.` (`"sysio_token"_n` and similar), now spelled with the dot.
- `fc::unsigned_int`'s converting constructor is SFINAE-constrained: once `name` joins namespace `fc` through its base, that unconstrained constructor poisoned fmtlib's `format_as` detection for containers of `name`.
- Comparisons use a defaulted `operator<=>` / `operator==` (C++20). `name`'s characterization tests move into a new `unittests/name_tests.cpp`.

The CDT side (`sysio::basic_name`, `sysio::name`) is a parallel change in wire-cdt — Wire-Network/wire-cdt#64; the two are meant to be merged together.


libraries/libfc/include/fc/basic_name.hpp
```diff
@@ -0,0 +1,177 @@
...
+ * @file fc/basic_name.hpp
+ * @brief Generic MSB-first packed 64-bit identifier.
+ *
+ * basic_name<Traits> is the shared implementation behind sysio::chain::name and
+ * fc::slug_name. Both pack a short string into a uint64_t, most-significant
+ * symbol first, so that integer ordering matches string ordering. They differ
+ * only in alphabet and length, which Traits supplies.
+ *
+ * A Traits type must provide:
+ *   static constexpr int              max_len;   // number of symbols
```
huangminghuang: I would use the following concept instead comment
```
template <typename Traits>
concept basic_name_traits =
   requires(std::string_view in, const char* why) {
      { Traits::max_len } -> std::convertible_to<int>;
      { Traits::alphabet } -> std::convertible_to<std::string_view>;
      { Traits::throw_invalid(in, why) } -> std::same_as<void>;
   } &&
   Traits::max_len > 0 &&
   std::string_view{Traits::alphabet}.size() > 0;
 ```

File: libraries/libfc/include/fc/basic_name.hpp [hunk 2]
```diff
@@ -0,0 +1,177 @@
...
+   std::string to_string() const {
+      std::string s;
+      for (int i = 0; i < Traits::max_len; ++i) {
+         const uint64_t sym = (value >> shift(i)) & width_mask(i);
+         s.push_back(char_of(sym));
+      }
+      // trailing symbol-0 slots are padding, not content
+      const char pad = char_of(0);
+      while (!s.empty() && s.back() == pad)
+         s.pop_back();
```
huangminghuang: This only trims trailing pad symbols, but contract-side sysio::slug_name::to_string() stops at the first zero slot at slug_name.hpp (line 66). For raw/deserialized slug values with an interior zero slot, host fc::slug_name can return strings containing embedded NULs and later characters while the contract mirror returns the prefix. If fc::slug_name is meant to remain byte/behavior-identical with the contract type, basic_name needs a trait for “zero is terminator” or slug_name needs its own to_string().

huangminghuang (commented):
contracts/tests/sysio.msgch_tests.cpp (line 334) still calls fc::slug_name_literals::operator""_s("ETH", 3) directly, but fc/slug_name.hpp (line 61) replaced that literal with the GNU template literal form. There is no (const char*, size_t) overload anymore, so this test target will not compile. Use "ETH"_s.value with the literal imported, or keep a compatible overload.

libraries/libfc/include/fc/basic_name.hpp
heifner: Done in 887ea3facd — added a `basic_name_traits` concept and constrained `basic_name<Traits>` with it; the prose trait-requirements list in the file header now points to the concept instead. Thanks for the suggestion.

libraries/libfc/include/fc/basic_name.hpp
heifner: Good catch. Fixed in 887ea3facd with option (a): a `Traits::zero_terminates` flag on the policy. `slug_name` sets it `true`, so `to_string()` stops at the first symbol-0 slot — identical to the contract-side `sysio::slug_name`. `name` sets it `false` and keeps trimming only trailing padding, since for name symbol 0 (the dot) is an ordinary interior character. The `basic_name_traits` concept requires the new member.

libraries/libfc/include/fc/basic_name.hpp
jglanz: I dig! @huangminghuang good call on the `concept`

contracts/sysio.opp.common/include/sysio.opp.common/slug_name.hpp
```diff
@@ -58,7 +58,7 @@ struct slug_name {
       for (std::size_t i = 0; i < s.size(); ++i) {
          const auto v = char_to_slot(s[i]);
          sysio::check(v != INVALID_SLOT, "slug_name: invalid character (alphabet is [A-Z0-9_])");
-         out |= (v << (i * 6));
+         out |= (v << (42 - i * 6));   // most-significant-symbol-first
```
jglanz, line 61: Glad i caught this; the template should allow for both MSB and LSB (maybe via an `enum class` with the template defaulting to MSB

jglanz (approved):
Approving, but please apply the comment first...

"the template should allow for both MSB and LSB (maybe via an `enum class` with the template defaulting to MSB"

huangminghuang (approved):

PR #354, opened 2026-05-23
Emissions: fund sysio.dclaim immediately on incoming OPP claim

heifner:
## Summary

Replaces the curve-driven capital BPS share with immediate funding tied to actual OPP claim load. `sysio.dclaim::onreward` now inline-calls a new `sysio.system::fundclaim` action that transfers the exact credit amount from `sysio`'s pool to `sysio.dclaim`, so a staker can claim in the next block instead of waiting for the next pay-epoch.

## Model

- `payepoch` transfers compute + capex + governance only. The "capital bucket" is gone -- the slot it used to occupy is now an *implicit capital reserve* (`period_emission - compute - capex - governance`) that stays in `sysio`'s balance.
- `fundclaim` drains the pool against the implicit reserve in real time, one inline action per onreward. The reserve naturally absorbs the rounding dust from per-share `split_bps` floors.
- When `fundclaim` is asked for more than the pool has left, it transfers what's available and bumps `t5_state.capital_shortfall_total` by the unfunded delta. It never throws (the OPP inbound dispatch chain must not be aborted by emissions-side conditions).

## ABI changes

- `emission_config.capital_bps` removed. `setemitcfg` now requires `compute_bps + capex_bps + governance_bps <= 10000`.
- `epoch_log.capital_amount` removed.
- `t5_state.capital_shortfall_total` (int64) added.
- New `sysio.system::fundclaim(int64_t amount)` action, auth=`sysio.dclaim`.

Pre-launch, so no migration shim.


contracts/sysio.system/src/emissions.cpp
```diff
@@ -728,6 +747,49 @@ void system_contract::payepoch(uint32_t epoch_index,
...
+   require_auth(CAPITAL_ACCOUNT);
+
+   if (amount <= 0) return;
+
+   t5state_t t5s(get_self());
+   if (!t5s.exists()) return; // pre-init: silently absorb
+   auto state = t5s.get();
+
+   const auto cfg = get_emit_cfg(get_self());
+   const int64_t remaining = cfg.t5_distributable - cfg.t5_floor - state.total_distributed;
```
huangminghuang: This caps fundclaim only by lifetime emission headroom, but ignores state.pending_emission_amount, which has already accrued for the current pay period at emissions.cpp (line 474). With pay_cadence_epochs > 1, a capital draw between accrue and payepoch can consume funds already scheduled for compute/capex/governance. The next payepoch then either blocks on insufficient balance or treasury exhaustion while pending_emission_amount remains unpaid. Cap the claim pool to the current period’s implicit capital reserve, or at least exclude pending payepoch obligations from the transferable amount.

contracts/sysio.system/src/emissions.cpp
```diff
@@ -728,6 +747,49 @@ void system_contract::payepoch(uint32_t epoch_index,
...
+   const int64_t to_transfer = (amount <= remaining) ? amount : (remaining > 0 ? remaining : 0);
+   const int64_t shortfall   = amount - to_transfer;
+
+   if (to_transfer > 0) {
+      send_wire_transfer(get_self(), CAPITAL_ACCOUNT, to_transfer, memo::capital);
```
huangminghuang, line 801: This still performs a real token transfer after only checking accounting headroom. If sysio’s actual WIRE balance is below to_transfer, sysio.token::transfer throws overdrawn balance at sysio.token.cpp (line 98). Since sysio.dclaim::onreward sends this inline at sysio.dclaim.cpp (line 252), the inbound OPP handler can still abort despite the “never throws” guarantee. Include actual sysio WIRE balance in the cap before sending the transfer.

contracts/sysio.system/src/emissions.cpp
heifner: Addressed in 2b4e0d1b1d. fundclaim now subtracts state.pending_emission_amount from the lifetime headroom cap, so a capital draw between accrueepoch and payepoch cannot consume the funds already scheduled for the next payepoch's compute/capex/governance distribution.

contracts/sysio.system/src/emissions.cpp
heifner: Addressed in 2b4e0d1b1d. fundclaim now caps to sysio's actual WIRE balance (also less pending_emission_amount) before sending the transfer, so sysio.token::transfer cannot throw overdrawn balance from inside the inline STAKING_REWARD dispatch. The final transfer amount is std::min({amount, lifetime_headroom - pending, balance - pending}); any unfunded delta accrues to capital_shortfall_total.

jglanz (approved):
looks good; personally I'd like to chat about how the different basis points (bps) are applied/related, but generally the code seems fine

PR #359, opened 2026-05-28
Opp: node-owner NFT registration — create-in-flow depot dispatch

heifner:
## Summary
Implements the full inbound OPP NodeOwnerRegistration claim ("trust-OPP", create-in-flow). When sysio.msgch processes an ATTESTATION_TYPE_NODE_OWNER_REG attestation it decodes it and inline-sends sysio.roa to create the claim account from the supplied Wire key, register the node owner at the requested tier, and record the depositor's external-chain key in sysio.authex. The depot is trusted, so claim-payload problems soft-fail into an audit row rather than aborting the dispatch.

## What's included
- Proto (libraries/opp/proto): add WireKey { WireKeyType key_type; bytes key } + WireKeyType (K1/R1/WA/EM/ED/BLS). WireKey is a standalone message rather than a field on the shared ChainAddress — a usually-default field on ChainAddress is omitted by proto3 on the wire but expected by the regenerated abi, which breaks abi decode of every table storing a ChainAddress. NodeOwnerRegistration: wire_account -> account; wire_pub_key is now a WireKey (the new account's owner/active key).
- sysio.msgch: dispatch_node_owner_reg decodes the account name, WireKey, and ETH key, then inline-sends sysio.roa::newnameduser (create) and ::nodeownreg (register) as {sysio.roa, active}; malformed envelopes are silently dropped.
- sysio.roa: nodeownreg(owner, tier, eth_pub_key, wire_pub_key) creates in-flow — confirms an existing account is controlled by the claimed key (active_key_matches: the key satisfies the active authority by itself), records the ETH link via inline sysio.authex::recordlink, registers, and writes a nodeownerreg audit row. Claim-payload errors (NAME_INVALID / OWNER_NOT_ACCOUNT / ACCOUNT_KEY_MISMATCH / DUPLICATE) soft-fail; tier-range / non-EM key / ROA-inactive stay hard checks. newnameduser is now non-throwing.
- sysio.authex: recordlink(account, chain_kind, pub_key) — trusted depot-only link insert (require_auth self), idempotent per account+chain; one external key may back several accounts.
- clio: convert eth_to_k1_private / eth_to_k1_public (reuse an Ethereum keypair as a Wire K1 signing key, printing the derived Ethereum address) + convert keccak256 / sha256 helpers.
- Docs (docs/key-formats.md): reference for all Wire key types (string + binary forms, encodings, signing-performance note) and the OPP carriage forms incl. WireKey.
- Tests: contracts/tests/sysio.roa_tests.cpp (create-in-flow nodeownreg + newnameduser cases); contracts/tests/sysio.dispatch_tests.cpp (a msgch->roa dispatch integration test); tests/nodeownreg_test.py (TestHarness integration, rewritten for create-in-flow); tests/clio_em_key_test.py (eth_to_k1 EIP-155 known-answer test).

The two sysio.code delegations the inline sends require (sysio.roa.active <- sysio.msgch@sysio.code, sysio.authex.active <- sysio.roa@sysio.code) are wired in wire-e2e-tests ClusterManager.


contracts/sysio.authex/src/sysio.authex.cpp
```diff
@@ -141,6 +141,42 @@ namespace sysio {
...
+   // return (never throw -- the trust-OPP depot dispatch must not abort).
+   //
+   // We deliberately do NOT reject when `pub_key` is already linked to a *different* account: one
+   // external wallet can hold several WireNodes NFTs and therefore legitimately back several Wire
+   // accounts, so one ETH key -> many accounts is allowed on this path. The operator path
+   // (createlink) keeps its 1:1 `bypubkey` check, which also guarantees an operator's link is the
+   // lowest-primary-key row for that key; sysio.msgch's resolve_account_from_op_address does a single
+   // `bypubkey.find()` (lowest-primary-key match), so it still lands on the operator's account rather
+   // than a later node-owner duplicate.
+   if (by_namechain.find(to_namechain_key(account, chain_kind)) != by_namechain.end()) return;
```
huangminghuang, line 167: recordlink returns whenever (account, chain) already exists, without checking that the existing pub_key matches the depositor key from this claim. nodeownreg then continues to regnodeowner and records CONFIRMED at sysio.roa.cpp (line 677). So a pre-existing account that already has an EVM link to key A can successfully claim an NFT with depositor key B, but the table still links the account to key A. I’d make the idempotent path only no-op when the stored key matches; otherwise have nodeownreg soft-fail/audit or update the link intentionally, with a regression test for mismatched pre-existing EVM links.

contracts/sysio.authex/src/sysio.authex.cpp
```diff
@@ -141,6 +141,42 @@ namespace sysio {
...
+
+   uint64_t next_key = 0;
+   if (links.cbegin() != links.cend()) {
+      auto last = --links.cend();
+      next_key = last->key + 1;
+   }
+   links.emplace("sysio"_n, links_key{next_key}, links_s{
```
jglanz: is there a globally available `constexpr` of `"sysio"_n`?

contracts/sysio.roa/CMakeLists.txt
```diff
@@ -1,10 +1,19 @@
...
+   find_cdt_magic_enum()
    add_contract(sysio.roa sysio.roa sysio.roa.cpp)
 
    target_include_directories(sysio.roa
       PUBLIC
-      ${CMAKE_CURRENT_SOURCE_DIR}/../sysio.system/include)
+      ${CMAKE_CURRENT_SOURCE_DIR}/../sysio.system/include
+      ${CMAKE_CURRENT_SOURCE_DIR}/../sysio.authex/include
+      ${CMAKE_CURRENT_SOURCE_DIR}/../sysio.opp.common/include
+      ${CMAKE_CURRENT_SOURCE_DIR}/../../libraries/libfc-lite/include
```
jglanz, line 11: i think there are interface targets (header only) that you can use instead of the `../../libraries/...`

jglanz (approved):
Approved, BUT please fix cmake comments before landing

contracts/sysio.authex/src/sysio.authex.cpp
heifner: Fixed in a016b9b49c. `nodeownreg` now reads the account's existing EVM link before the inline `recordlink` and soft-fails `REJECTED/LINK_KEY_MISMATCH` (new reject_reason 5) when the stored key differs from the claim's depositor key -- so a pre-existing link to key A can no longer be silently kept while recording CONFIRMED for a claim with key B. `recordlink` stays idempotent on an exact match; the mismatch is caught and audited in roa rather than overwriting. Added `nodeownreg_link_key_mismatch` (seeds a link with key A, claims with key B, asserts REJECTED/LINK_KEY_MISMATCH and not registered).

contracts/sysio.authex/src/sysio.authex.cpp
heifner: No global one -- added a file-local `constexpr name link_row_payer = "sysio"_n;` in the anonymous namespace and used it for both the `createlink` and `recordlink` emplace payers (a016b9b49c). The compiled wasm is byte-identical, so it's purely a readability / no-magic-literal change.

contracts/sysio.roa/CMakeLists.txt
heifner: I tried it, but CDT's contract compiler (`cdt-cpp`) only honors `target_include_directories` -- includes propagated via `target_link_libraries` from the sibling targets (`sysio.system`/`sysio.authex`, both PUBLIC) don't reach the `cdt-cpp` invocation, so the build fails with `'sysio.system/native.hpp' file not found`. The existing `INTERFACE magic_enum` link only "works" because magic_enum comes from the vcpkg include path, not the target. All 9 contracts use the same raw `$<BUILD_INTERFACE:.../include>` paths for this reason, so I've kept sysio.roa consistent with them.

huangminghuang (approved):

PR #363, opened 2026-05-31
ROA node-owner account-creation pre-work + exact sysio-funded RAM model

heifner:
Pre-work for the WireNodes NFT node-owner claim flow, stacked on the ex.* removal PR. Adds the account-creation primitive and makes all sysio-funded RAM exact and fully accounted.

## Exact sysio-funded RAM model
- **`giftram`** fixed + generalized: was minting chain quota via `add_system_resources` while drawing the reslimit from sysio (desynced, over-allocated). Now a conserving transfer — moves the delta `sysio.pool ↔ sysio.acct` bucket and `sysio.quota ↔ account.quota`. Bidirectional (a shrink reclaims back to sysio) and authorized by `sysio.roa` only (`setsyscode`/`setsysabi` call it; createlink no longer calls giftram after the links-only pivot).
- **`setsyscode`/`setsysabi`**: deploy a system contract's code/abi and gift exactly the RAM it consumes out of sysio's pool (measured after, via `giftram`). `setsyscode` also sets the account privileged (inline `setpriv`, ordered after `setcode` — the chain rejects `setpriv` on a codeless account). Re-callable; a smaller re-deploy reclaims.
- **`activateroa`**: partition `T = node reserve + roa + sysio` with the `sysio.acct` bucket seed taken from sysio's share. Dropped the per-contract carve-outs — system contracts self-fund exactly via `setsyscode`.

## Account creation
- **`newnameduser(account, pubkey, tier)`**: the create step of the claim flow. Creates an account with a user-chosen vanity name and the holder's K1 key as owner/active, funded the fixed `newaccount_ram` from sysio's pool. `require_auth(get_self())` (depot-dispatched as `{sysio.roa, active}` via the `msgch@sysio.code` delegation, like `nodeownreg`). Idempotent if the account already exists. Tier name rules: tier-1 = 2–6 char prefix, tier 2/3 = up to 12 chars.
- **`newuser`**: flip sub-account naming `<random>.<suffix>` → `<prefix>.<random>` (a tier-1 owner is a short prefix; ROA is privileged so it can mint any name). Replaced the per-attempt `sha256` in the generator with a cheap splitmix64 finalizer — no crypto intrinsic needed since uniqueness is enforced by the `is_account` retry.

## Cleanup
- Remove the legacy `initnodereg → setpending → finalizereg` state machine (superseded by the trust-OPP flow). `forcereg` and the `nodeownerreg` table are kept.

## Tests
- `setsyscode` exact-gift + full conservation, re-deploy reclaim, `setsysabi` gift; `newnameduser` creation/funding/idempotency + tier name rules; `newuser` prefix naming. The createlink test asserts the account bears no RAM cost (the link row is sysio-paid). Full roa + authex suites pass; `nodeop_run_test` integration passes.

## Deferred (depot-dispatch deliverable, separate PR)
`createlink` depot-auth, the bios-boot `setsyscode` reorder, and the msgch `NODE_OWNER_REG` dispatch + `sysio.roa.active ← msgch@sysio.code` delegation.


contracts/sysio.roa/sysio.roa.cpp
```diff
@@ -846,6 +829,52 @@ namespace sysio {
...
+        // Create the account with the holder's K1 key as both owner and active.
+        auto auth = sysiosystem::authority{1, {{pubkey, 1}}, {}};
+        action(permission_level{get_self(), "active"_n}, "sysio"_n, "newaccount"_n,
+               std::make_tuple(get_self(), account, auth, auth)).send();
+
+        // Fund the fixed newaccount_ram from sysio's pool (same model as newuser): bucket the RAM
+        // under sysio.acct and draw it from sysio. The inline newaccount's transfer_ram (sysio.system)
+        // moves the matching chain quota from sysio to the new account.
+        auto sys_symbol = state.total_sys.symbol;
+        int64_t ram_weight_amount = sysiosystem::newaccount_ram / state.bytes_per_unit;
```
huangminghuang: This converts newaccount_ram to policy units with integer division. setbyteprice rejects non-divisible bytes_per_unit, but activateroa does not, so the initial activation can still choose a value that causes every newnameduser to move 1144 bytes while under-recording the sysio.acct policy. Add the same divisibility check to activateroa, or store/account the exact byte delta.

contracts/sysio.roa/sysio.roa.hpp
```diff
@@ -158,21 +127,63 @@ namespace sysio {
...
+             * @param pubkey  The holder's K1 public key (becomes owner and active).
+             * @param tier    Node-owner tier 1/2/3 (selects the name-length rule).
+             */
+            [[sysio::action]]
+            void newnameduser(const name& account, const public_key& pubkey, uint8_t tier);
+
             /**
              * @brief Gifts an account exactly the RAM consumed since `usage_before`.
              *
              * Called by sysio.authex after createlink adds an external-chain key to an
```
huangminghuang: This still documents giftram as callable by sysio.authex after createlink adds an external key to active. The implementation now explicitly requires sysio.roa and says sysio.authex is not an authorizer, so the public contract docs need to be updated.

contracts/sysio.roa/sysio.roa.cpp
heifner: Fixed in aad85f286a. `activateroa` now runs the same divisibility check `setbyteprice` had, factored into a shared `check_divisible_byte_price` helper, so a non-divisible `bytes_per_unit` is rejected at initial activation (no more under-recording the sysio.acct policy). Also added a `bytes_per_unit > 0` guard since the unit conversion would otherwise divide by zero. Added `byteprice_divisibility_guard` covering the shared check (non-divisible, zero, and the accepted case) via `setbyteprice`.

contracts/sysio.roa/sysio.roa.hpp
heifner: Fixed in aad85f286a. Updated the `giftram` header doc to match the implementation: it is authorized by `sysio.roa` itself (setsyscode/setsysabi), and since createlink now records the external-chain link only, `sysio.authex` no longer calls it and is not an authorizer.

contracts/sysio.roa/sysio.roa.cpp
```diff
@@ -200,25 +217,96 @@ namespace sysio {
...
-        // Gift exactly the RAM the caller's preceding action consumed on `account`
-        // (e.g. createlink adding an external key to `active`). RAM is checked at
-        // transaction end, so usage already reflects that growth when this runs.
+        // Unlimited-RAM accounts (privileged/system, or not yet ROA-managed) have no RAM
+        // constraint to satisfy and must NOT be shrunk to a finite limit — skip the gift.
+        // (The delta>0 path is exercised when a finite account's RAM grows -- e.g. setsyscode
+        //  deploying a contract, or the depot-driven node-owner flow; see the nodeownreg tests' TODO.)
+        int64_t cur_ram, cur_net, cur_cpu;
+        get_resource_limits(account, cur_ram, cur_net, cur_cpu);
+        if (cur_ram < 0) return;
```
huangminghuang: This skips giftram whenever the target account still has unlimited RAM, but real system accounts are created with default ram_bytes = -1 and this PR removed the old sysio.authex finite carve-out. So calling setsyscode/setsysabi for those accounts will deploy code/ABI, then return before moving bytes out of sysio or into sysio.acct, breaking the “system contracts self-fund exactly” model. The tests only cover alice after regnodeowner gives it finite RAM, so this misses the actual system-account path.

contracts/sysio.roa/sysio.roa.cpp
huangminghuang, line 195: This allows changing bytes_per_unit, but the existing sysio.acct policy keeps the original bytes_per_unit from activation while later newuser/newnameduser add RAM units computed from the new global price at line 830 (line 830) and line 880 (line 880). Any accepted divisor other than the original, for example 8, makes the policy’s ram_weight no longer represent the bytes moved, so exact accounting drifts again. Either freeze this price once the bucket exists or roll/account the bucket in a way that preserves units.

contracts/sysio.roa/sysio.roa.cpp
heifner: Fixed in a0fec0f7e1. `giftram` now rejects an unlimited (`-1`) RAM target with `check(cur_ram >= 0, ...)` instead of silently returning -- a system account that wasn't first brought under ROA management (finite 0 quota) now fails loudly rather than deploying code with the bytes unfunded. In the prod flow the account is created with a finite quota before `setsyscode`/`setsysabi` runs, so the gift is always accounted; `setsyscode`'s inline `setpriv` flips only the privileged flag, not the RAM limit, so finite targets are unaffected. Added `setsyscode_rejects_unlimited_ram_target` covering the raw system-account (`-1`) path you flagged.

contracts/sysio.roa/sysio.roa.cpp
heifner: Fixed in a0fec0f7e1. `newuser`/`newnameduser` now convert `newaccount_ram` to `sysio.acct` policy units at the bucket's own frozen `bytes_per_unit` (`acct_pol.bytes_per_unit`) rather than the live `state.bytes_per_unit` -- the "account the bucket in a way that preserves units" option -- so a later `setbyteprice` no longer makes the bucket's `ram_weight` diverge from the bytes actually moved. `setbyteprice` stays usable for network expansion since each node-owner policy stores its own creation price. Added `sysio_acct_bucket_uses_frozen_price`: after `setbyteprice(8)` a creation still records `newaccount_ram/104` units, not `newaccount_ram/8`.

contracts/sysio.roa/sysio.roa.cpp
```diff
@@ -116,31 +129,35 @@ namespace sysio {
...
+        // Convert the leftover (SYS units) to bytes and partition it so the grand total of all
+        // reslimits stays exactly total_sys * bytes_per_unit — nothing is minted on top:
+        //   T = node-owner reserve (allocated above) + roa allocation + sysio pool.
+        // sysio.roa keeps half the leftover for its own (growing) bookkeeping tables; sysio gets
+        // the rest as THE pool that funds account creation and every other system contract's RAM
+        // (deployed via setsyscode/setsysabi, which gift the exact bytes out of this pool). Other
+        // system contracts are deliberately NOT pre-allocated here — they self-fund exactly. The
+        // only deduction is the sysio.acct account-creation bucket seed, taken out of sysio's
+        // share so it stays conserved.
+        uint64_t leftover_bytes = (uint64_t)(leftover * bytes_per_unit);
```
huangminghuang: Validate leftover before converting it to bytes: sysio.roa.cpp (line 141)
leftover is signed, but line 141 multiplies it with bytes_per_unit and stores the result as uint64_t. For small valid-looking supplies, tier rounding can make allocated > total_amount; e.g. total_sys.amount == 13 gives allocated == 21, so leftover == -8. That underflows into a huge unsigned byte count, and the new sysio_gross > acct_seed_bytes check passes instead of rejecting activation. I’d add explicit checks before this conversion, such as total_sys.amount > 0, allocated <= total_amount, and enough positive leftover bytes for newaccount_ram, plus an activation regression test with a tiny supply.

heifner: Fixed in 743a99aaa4. `activateroa` now validates the supply before the signed->unsigned conversion: `check(total_amount > 0)` and `check(allocated <= total_amount)` reject a supply too small for the node-owner reserve (your `total==13 -> allocated==21` case) instead of letting `leftover` go negative and underflow `leftover_bytes`. The guarded value is then cast explicitly with `(uint64_t)leftover * bytes_per_unit`. Added `roa_unactivated_tester` (deploys sysio.roa privileged but skips init_roa's auto-activation) with `activateroa_rejects_tiny_supply` (`0.0013 SYS` rejected) and `activateroa_accepts_normal_supply`.

huangminghuang (approved):

PR #367, opened 2026-06-03
Chain: fix kv same_payer billing and reject payer==0 on insert

heifner:
## Problem
`apply_context::kv_set` (update branch) and `kv_idx_update` resolved a payer value of 0 -- the CDT `same_payer` sentinel (`name{}`) -- to the **receiver** (the contract account) instead of keeping the row's existing payer. A `modify(same_payer, ...)` on a kv row therefore moved the row's RAM from its real payer onto the contract.

This was invisible while every kv row was billed to the contract itself (`same_payer -> receiver` then equalled the real payer). It surfaces the moment a row is billed to a different account -- e.g. a privileged system contract billing its rows to the `sysio` RAM pool: on the next `same_payer` modify the contract is re-billed and overflows its exact, gifted RAM limit.

## Fix
`payer_val == 0` is the `same_payer` sentinel, now handled with full classic `db_*` parity:

- **Update** (`kv_set` update branch, `kv_idx_update`): `payer_val == 0` keeps the row's existing payer (`itr->payer`); it never moves billing onto the receiver.
- **Insert** (`kv_set` create branch, `kv_idx_store`): a new row has no existing payer to keep, so `payer_val == 0` is rejected with `invalid_table_payer` ("must specify a valid account to pay for new record"). To self-pay, a contract names `get_self()`.

This mirrors classic EOSIO/Spring exactly: `db_store_i64` and `generic_index::store` assert `payer != account_name()`, while `db_update_i64` and the secondary `update` do `if (payer == account_name()) payer = obj.payer`. The previous receiver-default-on-insert was a Wire-only convenience that no production contract relied on -- the CDT `kv::table`/`kv_multi_index` wrappers forward `payer.value` verbatim and every system contract names `get_self()` on insert. `invalid_table_payer` (3160001) is re-added; it had been dropped as "legacy db_* only". The `kv_set` / `kv_idx_store` / `kv_idx_update` intrinsic docs are updated for the insert (error) vs update (keep existing) distinction.

## Coverage
`payer_choice_test` exercises every RAM-billing kv route with a payer distinct from the contract -- the gap that let this ship:

- `kv_set` insert -- explicit payer bills it; `payer == 0` throws `invalid_table_payer`
- `kv_set` update -- payer change moves billing; `same_payer` keeps the existing payer
- `kv_erase` -- refunds the row's payer
- `kv_idx_store` insert -- explicit payer bills it; `payer == 0` throws `invalid_table_payer`
- `kv_idx_update` -- payer change moves billing; `same_payer` keeps the existing payer
- `kv_idx_remove` -- refunds the entry's payer

The `same_payer`-on-update and insert-0-throw cases are regression tests written to fail against the pre-fix host. `kv_api_tests` is updated to match (`payer_change_back_to_self` now names the receiver explicitly, since `0` no longer means "receiver" on update). The `test_kv_api` contract's raw `(0, ...)` "receiver pays" sites now name `get_self()`, with `tstpayerself` and a new `idxnopay` action probing primary/secondary insert-0 rejection (committed wasm/abi regenerated).


huangminghuang:
Update the public KV intrinsic docs for payer == 0 semantics.
PR #367 changes update behavior so payer == 0 keeps the existing payer for kv_set updates and kv_idx_update, but interface.hpp (line 786) and interface.hpp (line 948) still document 0 as “bill receiver.” That is now only true for inserts/new entries. This is a public contract for intrinsic users, so it should be updated with the create-vs-update distinction.

huangminghuang (approved):

PR #372, opened 2026-06-05
P2P: Fix trx listener connection type handling

huangminghuang:
## Why

`p2p_no_blocks_if_test` failed in the sharded CI run because a node that should only receive transactions over a `:trx` listener could still receive blocks.

Failed run reference: https://github.com/Wire-Network/wire-sysio/actions/runs/27022195099/job/79758140428

The root cause had two pieces:

- Incoming connections accepted on a `:trx` listener were initialized as `both` and could be widened by the peer's advertised normal `p2p_address`.
- `p2p-listen-endpoint` values were sorted before being paired with positional `p2p-server-address` values, which could swap the advertised normal and `:trx` addresses for multi-listener nodes.

## What changed

- Initialize incoming connection type from the listener that accepted the socket.
- For incoming handshakes, let the peer-advertised address only narrow the connection type.
- Preserve `p2p-listen-endpoint` order while de-duplicating so `p2p-server-address` pairing remains correct.
- Multi-listener nodes now advertise the `p2p-server-address` paired positionally with the listener endpoint that accepted the connection; this preserves configured order rather than sorted endpoint order.
- Extract endpoint de-duplication and connection-type selection into `net_utils.hpp` so the behavior is covered by deterministic unit tests.
- Add regression coverage for the order-preserving de-dupe case and for preventing a configured incoming `:trx` listener from being widened by a plain peer address.

## Verification

- `cmake --build build/codex-system-contracts --target nodeop -- -j6`
- Cherry-picked the fix onto post-sharding branch `codex/port-sharded-np-lr-tests` and ran:
  - `ctest -R '^p2p_no_blocks_if_test$' --output-on-failure --timeout 1000`
  - Result: passed, 98.69 sec
- Compiled touched host objects:
  - `plugins/net_plugin/CMakeFiles/net_plugin.dir/src/net_plugin.cpp.o`
  - `plugins/net_plugin/test/CMakeFiles/test_net_plugin.dir/connection_type_unittest.cpp.o`
- Relinked `test_net_plugin` and ran:
  - `./build/codex-system-contracts/plugins/net_plugin/test/test_net_plugin --run_test=listen_endpoint_dedupe_preserves_server_address_pairing,incoming_trx_listener_type_cannot_be_widened`
  - `./build/codex-system-contracts/plugins/net_plugin/test/test_net_plugin`
  - Result: passed, 66 test cases

### DIFF
--- a/plugins/net_plugin/include/sysio/net_plugin/net_utils.hpp
+++ b/plugins/net_plugin/include/sysio/net_plugin/net_utils.hpp
@@ -6,9 +6,12 @@
 
 #include <boost/asio/ip/address_v6.hpp>
 
+#include <set>
 #include <string>
+#include <string_view>
 #include <sstream>
 #include <regex>
+#include <vector>
 
 namespace sysio::net_utils {
 
@@ -20,6 +23,12 @@ namespace sysio::net_utils {
 // Allow for future extentions as well, hence 384.
 constexpr size_t max_p2p_address_length = 253 + 6 + 30;
 constexpr size_t max_handshake_str_length = 384;
+/// Address suffix for transaction-only p2p connections.
+inline constexpr std::string_view trx_connection_type = "trx";
+/// Address suffix for block-only p2p connections.
+inline constexpr std::string_view blk_connection_type = "blk";
+/// Length of a recognized p2p connection type suffix.
+inline constexpr size_t connection_type_length = 3;
 
 namespace detail {
 
@@ -134,11 +143,54 @@ namespace detail {
       if (host.empty() || port.empty()) return {};
 
       std::string type;
-      if (remainder.starts_with("blk") || remainder.starts_with("trx")) {
-         type = remainder.substr(0, 3);
+      if (remainder.starts_with(blk_connection_type) || remainder.starts_with(trx_connection_type)) {
+         type = remainder.substr(0, connection_type_length);
       }
 
       return {std::move(host), std::move(port), std::move(type)};
+   }
+
+   /// Connection traffic classes requested by local configuration or a peer's advertised endpoint.
+   enum class connection_type : char {
+      both,
+      transactions_only,
+      blocks_only
+   };
+
+   /// De-duplicate listen endpoints while preserving first-seen order so positionally paired
+   /// p2p-server-address values remain aligned with their p2p-listen-endpoint.
+   inline std::vector<std::string> dedupe_preserve_order(const std::vector<std::string>& addresses) {
+      std::vector<std::string> result;
+      result.reserve(addresses.size());
+      std::set<std::string> seen;
+      for (const auto& addr : addresses) {
+         if (seen.insert(addr).second) {
+            result.emplace_back(addr);
+         }
+      }
+      return result;
+   }
+
+   /// Return the requested connection type encoded in an address, or both when the address is invalid
+   /// or does not explicitly select transaction-only or block-only traffic.
+   inline connection_type type_from_address(const std::string& addr) {
+      const auto& [host, port, type] = split_host_port_type(addr);
+      if (host.empty() || type.empty()) {
+         return connection_type::both;
+      }
+      if (type == trx_connection_type) {
+         return connection_type::transactions_only;
+      }
+      if (type == blk_connection_type) {
+         return connection_type::blocks_only;
+      }
+      return connection_type::both;
+   }
+
+   /// Apply a peer's advertised type only when the local configuration has not already selected
+   /// transaction-only or block-only traffic.
+   inline connection_type narrow_connection_type(connection_type current, const std::string& peer_addr) {
+      return current == connection_type::both ? type_from_address(peer_addr) : current;
    }
 
    /// @return listen address and block sync rate limit (in bytes/sec) of address string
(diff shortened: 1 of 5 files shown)


heifner (commented):
Nice fix — and it closes **both** halves of the root cause, which is easy to half-miss here: incoming connections are now typed from the listener that accepted them (ctor `set_connection_type`), and the order-preserving dedup keeps `p2p-listen-endpoint` aligned with the positional `p2p-server-address` so a `:trx` port that sorts before the normal port no longer swaps their advertised addresses. Switching the incoming handshake path from `set_connection_type` to `set_peer_connection_type` is the right call: the configured listener should be authoritative, and the peer's advertised address should only ever *narrow* (`both → trx/blk`), never widen.

### Thread safety — traced it, looks sound
`connection_type` is touched from several threads, so I checked the new ctor write and the narrowing RMW:
- it's `std::atomic`, so concurrent readers (`is_blocks_connection()` et al.) never tear;
- post-construction the only writer is `set_peer_connection_type`, which runs on the connection strand (`handle_message`), so the non-atomic `if (== both) … = trx` has a single writer and can't lose an update;
- the new ctor `set_connection_type` runs before the connection is published — `connections.add()` happens inside the strand-posted lambda in `create_session`, so the ctor write happens-before any reader.

Bonus: this actually *shrinks* the prior race window — an inbound `:trx` connection is `transactions_only` from construction instead of defaulting to `both` until the handshake lands.

### Main ask: a deterministic regression test
The only current guard is `p2p_no_blocks_if_test`, which only *caught* this because the harness happened to assign node02's trx port (17891) below its normal port (18115), so the old `std::sort` reordered them. On a different port draw the buggy code passes CI — so this class of failure stays flaky. `plugins/net_plugin/test/test_net_plugin` is a natural home; two small extractions make the logic unit-testable (and line up with the repo's "no duplicated helpers / testable units" guidance).

**1) Dedup (the sort/pairing bug) → `net_utils.hpp`:**
```cpp
/// De-duplicate listen endpoints while preserving first-seen order, so positionally
/// paired p2p-server-address values stay aligned with their endpoint.
inline std::vector<std::string> dedupe_preserve_order(const std::vector<std::string>& addresses) {
   std::vector<std::string> result;
   result.reserve(addresses.size());
   std::set<std::string> seen;
   for (const auto& addr : addresses)
      if (seen.insert(addr).second)
         result.emplace_back(addr);
   return result;
}
```
```cpp
BOOST_AUTO_TEST_CASE(listen_endpoint_dedupe_preserves_server_address_pairing) {
   using namespace sysio::net_utils;
   // node02's config from the failing run: trx port (17891) sorts BEFORE the normal port
   // (18115), so the pre-fix std::sort moved ":17891:trx" ahead of ":18115" while the
   // server-address list stayed in CLI order — pairing the trx listener with the normal
   // address and typing the connection `both`.
   const std::vector<std::string> listen = { "localhost:18115", "localhost:17891:trx", "localhost:18115" };
   std::vector<std::string>       server = { "localhost:18115", "localhost:17891:trx" };

   const std::vector<std::string> deduped = dedupe_preserve_order(listen);
   BOOST_REQUIRE_EQUAL(deduped.size(), 2u);            // duplicate dropped
   BOOST_CHECK_EQUAL(deduped[0], "localhost:18115");   // order preserved (std::sort would flip these)
   BOOST_CHECK_EQUAL(deduped[1], "localhost:17891:trx");

   server.resize(deduped.size());
   for (size_t i = 0; i < deduped.size(); ++i)          // each listener's type matches its paired address
      BOOST_CHECK_EQUAL(std::get<2>(split_host_port_type(deduped[i])),
                        std::get<2>(split_host_port_type(server[i])));
}
```

**2) Connection-type decision (the widening bug):** move the `connection_types` enum out of the `connection` class into `net_utils.hpp` and split the setters into pure functions the methods call:
```cpp
enum class connection_type : char { both, transactions_only, blocks_only };

inline connection_type type_from_address(const std::string& addr) {            // = set_connection_type
   auto [host, port, t] = split_host_port_type(addr);
   if (host.empty() || t.empty()) return connection_type::both;
   if (t == "trx") return connection_type::transactions_only;
   if (t == "blk") return connection_type::blocks_only;
   return connection_type::both;
}
inline connection_type narrow_type(connection_type cur, const std::string& peer) { // = set_peer_connection_type
   return cur != connection_type::both ? cur : type_from_address(peer);            // never widen
}
```
```cpp
BOOST_AUTO_TEST_CASE(incoming_trx_listener_type_cannot_be_widened) {
   using namespace sysio::net_utils;
   connection_type t = type_from_address("localhost:17891:trx");
   BOOST_CHECK(t == connection_type::transactions_only);
   t = narrow_type(t, "localhost:18114");                                  // plain peer addr must NOT widen
   BOOST_CHECK(t == connection_type::transactions_only);                   // pre-fix this became `both`
   BOOST_CHECK(narrow_type(type_from_address("localhost:18115"), "p:1:trx") == connection_type::transactions_only);
   BOOST_CHECK(narrow_type(connection_type::transactions_only, "p:1:blk") == connection_type::transactions_only);
}
```

### Minor
- The ctor comment "Incoming listener type is authoritative" is slightly stronger than the code: the value passed is the paired `p2p-server-address`, not the raw listen endpoint. Identical in normal configs; only diverges if a `:trx` listen endpoint is given a non-`:trx` server address.

heifner (approved):

PR #377, opened 2026-06-06
Handle unavailable sys-vm-oc test options

huangminghuang:
## Summary

Handle test code paths that reference `sys-vm-oc` when the current `nodeop` build does not expose OC options.

## What changed

- Adds `Utils.nodeopSupportsOption()` so Python tests can check the actual built `nodeop --help` output.
- Gates read-only test `--sys-vm-oc-*` arguments on real `nodeop` option support instead of assuming every Linux build supports OC.
- Keeps forced-OC read-only tests as clean skips when OC is unavailable.
- Filters performance harness `ChainPluginArgs` constructor kwargs to the fields present in the generated plugin-args dataclass.

## Why

`sys-vm-oc` is only present on supported builds, currently Linux x86_64 with OC enabled. The generated `ChainPluginArgs` classes and `nodeop` command-line options therefore differ across platforms. These tests should keep exercising OC where it exists, skip tests that explicitly require unavailable OC, and avoid passing unsupported OC arguments on macOS arm64 or other non-OC builds.

## Testing

- `python3 -m py_compile tests/TestHarness/testUtils.py tests/read_only_trx_test.py tests/interrupt_read_only_trx_test.py tests/PerformanceHarness/performance_test_basic.py`
- Previous macOS arm64 focused CTest skip validation for the read-only forced-OC tests passed:

```bash
ctest --test-dir build/macos-arm64-jit-chain-release \
  -R '^read-only-trx-parallel-if-sys-vm-oc-test$|^interrupt-read-only-trx-parallel-if-sys-vm-oc-test$' \
  --output-on-failure \
  --timeout 120
```

Result: `2/2` passed in `0.52 sec`.

## Notes

This PR now includes the performance harness fix from PR #378. PR #378 is superseded by this combined PR.

### DIFF
--- a/tests/PerformanceHarness/performance_test_basic.py
+++ b/tests/PerformanceHarness/performance_test_basic.py
@@ -651,9 +651,14 @@ class PerformanceTestBasic:
     def setupClusterConfig(args) -> ClusterConfig:
 
         def supportedPluginArgs(pluginArgsClass, **kwargs):
-            """Return only constructor arguments supported by the generated plugin-args dataclass."""
+            """Drop only constructor arguments for chain-plugin options compiled out of this nodeop build."""
+            conditionallySupportedFields = {"sysVmOcCacheSizeMb", "sysVmOcCompileThreads"}
             supportedFields = {field.name for field in dataclasses.fields(pluginArgsClass)}
-            return {key: value for key, value in kwargs.items() if key in supportedFields}
+            return {
+                key: value
+                for key, value in kwargs.items()
+                if key in supportedFields or key not in conditionallySupportedFields
+            }
 
         chainPluginArgs = ChainPluginArgs(**supportedPluginArgs(
                                         ChainPluginArgs,
--- a/tests/PerformanceHarness/performance_test_basic.py
+++ b/tests/PerformanceHarness/performance_test_basic.py
@@ -85,7 +85,7 @@ class PerformanceTestBasic:
             def __str__(self) -> str:
                 args = []
                 for field in dataclasses.fields(self):
-                    match = re.search("\w*PluginArgs", field.name)
+                    match = re.search(r"\w*PluginArgs", field.name)
                     if match is not None:
                         args.append(f"{getattr(self, field.name)}")
                 return " ".join(args)
@@ -126,7 +126,7 @@ class PerformanceTestBasic:
             # Producer Nodes are index [0, producerNodeCount) and non-producer nodes (validationNodeCount, apiNodeCount) nodes follow the producer nodes [producerNodeCount, _totalNodes)
             self._producerNodeIds = list(range(0, self.producerNodeCount))
             self._validationNodeIds = list(range(self.producerNodeCount, self.producerNodeCount + self.validationNodeCount))
-            self._apiNodeIds = list(range(self.producerNodeCount + self.validationNodeCount, self.producerNodeCount + self.validationNodeCount + self.validationNodeCount))
+            self._apiNodeIds = list(range(self.producerNodeCount + self.validationNodeCount, self.producerNodeCount + self.validationNodeCount + self.apiNodeCount))
 
             def configureValidationNodes():
                 validationNodeSpecificNodeopStr = ""
@@ -654,11 +654,16 @@ class PerformanceTestBasic:
             """Drop only constructor arguments for chain-plugin options compiled out of this nodeop build."""
             conditionallySupportedFields = {"sysVmOcCacheSizeMb", "sysVmOcCompileThreads"}
             supportedFields = {field.name for field in dataclasses.fields(pluginArgsClass)}
-            return {
-                key: value
-                for key, value in kwargs.items()
-                if key in supportedFields or key not in conditionallySupportedFields
-            }
+            supportedArgs = {}
+            for key, value in kwargs.items():
+                if key in supportedFields or key not in conditionallySupportedFields:
+                    supportedArgs[key] = value
+                else:
+                    print(f"Skipping unsupported chain_plugin argument '{key}' for this nodeop build")
+            return supportedArgs
+
+        if args.non_prods_sys_vm_oc_enable and not Utils.nodeopSupportsSysVmOc():
+            Utils.errorExit("--non-prods-sys-vm-oc-enable requires a nodeop build with SYS VM OC support")
 
         chainPluginArgs = ChainPluginArgs(**supportedPluginArgs(
                                         ChainPluginArgs,
--- a/tests/PerformanceHarness/performance_test_basic.py
+++ b/tests/PerformanceHarness/performance_test_basic.py
@@ -650,12 +650,20 @@ class PerformanceTestBasic:
 
     def setupClusterConfig(args) -> ClusterConfig:
 
-        chainPluginArgs = ChainPluginArgs(signatureCpuBillablePct=args.signature_cpu_billable_pct,
+        def supportedPluginArgs(pluginArgsClass, **kwargs):
+            """Return only constructor arguments supported by the generated plugin-args dataclass."""
+            supportedFields = {field.name for field in dataclasses.fields(pluginArgsClass)}
+            return {key: value for key, value in kwargs.items() if key in supportedFields}
+
+        chainPluginArgs = ChainPluginArgs(**supportedPluginArgs(
+                                        ChainPluginArgs,
+                                        signatureCpuBillablePct=args.signature_cpu_billable_pct,
                                         chainThreads=args.chain_threads, databaseMapMode=args.database_map_mode,
                                         wasmRuntime=args.wasm_runtime, contractsConsole=args.contracts_console,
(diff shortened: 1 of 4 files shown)


heifner (commented):
Reviewed all four files plus the surrounding harness and C++ context. The approach is sound — probing the built `nodeop --help` is strictly better than the old `platform.system() != "Linux"` check, and it transparently fixes Linux `ENABLE_OC=OFF` builds, not just macOS arm64. I did find two real logic bugs in the new code plus a few smaller items. Everything below is test-harness scope (no consensus/production surface).

Premises verified while reviewing:
- The OC options are registered only under `#ifdef SYSIO_SYS_VM_OC_RUNTIME_ENABLED` (`plugins/chain_plugin/src/chain_plugin.cpp:371-386`), so probing `--help` is a sound capability check. No other option's help text mentions a literal `--sys-vm-oc-*` string, so the regex can't false-positive today.
- `ChainPluginArgs.py` is generated at build time from the built nodeop (`tests/PerformanceHarness/NodeopPluginArgs/CMakeLists.txt`), so the dataclass genuinely lacks the OC fields on non-OC builds.
- The log line these tests grep for is emitted only on speculative (non-block-application) execution: `wasm_interface_private.hpp:171-172` — the comment there literally says "read_only_trx_test.py looks for this log statement".
- Exercised `Utils.nodeopSupportsOption` against a real OC-enabled nodeop: `--sys-vm-oc-enable` → True, unknown options → False, and the prefix `--sys-vm-oc` → False (the suffix guard works).

### 1. Bug: interrupt test's `noOC` assertion checks the wrong node and will spuriously fail

`interrupt_read_only_trx_test.py:113-120`:

```python
found = producerNode.findInLog(f"executing {sysioCodeHash} with sys vm oc")
if noOC or not ocSupported:
    assert(not found)
else:
    assert(found)
```

`--sys-vm-oc-enable none` is applied only to node index `pnodes` (the API node) via `specificExtraNodeopArgs[pnodes]`. `producerNode` (node 0) runs with default settings — OC tier-up `auto` — so on an OC build it WILL speculatively execute sysio.token via OC during bootstrap and log the line. A manual run of `interrupt_read_only_trx_test.py --sys-vm-oc-enable none` on any Linux OC build trips `assert(not found)`.

This is exactly the bug commit 0db6efca42 ("fix the oc found assertion") fixed in `read_only_trx_test.py` by splitting the check (apiNode expect-absent for noOC, producerNode expect-present otherwise) — but the interrupt test, which has identical topology, didn't get the same fix. It's latent in CI only because `tests/CMakeLists.txt:182-184` never passes `none` to the interrupt test. Suggest mirroring the read_only split (`apiNode` is already assigned before the assertion).

### 2. Bug: the skip condition excludes the only valid forced-OC configuration

Both tests:

```python
if args.sys_vm_oc_enable != "none" and not ocSupported:
    if allOC or args.wasm_runtime == "sys-vm-oc-forced":
        ... skip ...
```

On an OC build, `--wasm-runtime sys-vm-oc-forced` requires `--sys-vm-oc-enable none`: `wasm_interface_private.hpp:114` asserts "You can't use SYS VM OC as the base runtime when tier up is activated" (default tier-up is `auto`). So the only combination that actually runs forced-OC mode is `none` + `sys-vm-oc-forced`.

On a non-OC build, that exact combination bypasses the skip (the value IS "none"), the test proceeds, appends `--wasm-runtime sys-vm-oc-forced`, and nodeop throws at startup ("wasm runtime not supported on this platform and/or configuration", `wasm_interface_private.hpp:109-110`) — a cluster-launch failure instead of the clean skip this PR sets out to provide.

The outer `!= "none"` adds nothing for the `allOC` half (`all` already implies not-none) and is actively wrong for the forced half. Suggested simplification in both files:

```python
if not ocSupported and (allOC or args.wasm_runtime == "sys-vm-oc-forced"):
    Print("sys-vm-oc is unavailable on this platform. Skip the test")
    exit(0)
```

### 3. The apiNode "expect absent" assertion is near-vacuous

`read_only_trx_test.py:146-149`: the OC log line is only emitted when `!context.is_applying_block()`. At assertion time (right after `waitOnClusterBlockNumSync(3)`, before any test traffic), the apiNode has mostly applied blocks; its only speculative executions are P2P-relayed bootstrap trxs, and even those only log after the async OC compile completes and a subsequent execution happens. So `assert(not found)` can never false-fail (good — no flake risk), but it would very likely still pass even if `--sys-vm-oc-enable none` were silently broken.

If the negative branch should actually verify `none` took effect, re-check the apiNode log after read-only transactions have run (e.g. at the end of `basicTests()`): read-only trxs execute on the apiNode via the speculative path, which is precisely where the tier-up decision and log statement live.

### 4. Perf harness: `--non-prods-sys-vm-oc-enable` still injects an unsupported option on non-OC builds

`performance_test_basic.py:137-138` appends `--sys-vm-oc-enable all` to validation nodes unguarded — a raw unknown-option failure at nodeop startup on non-OC builds. The PR filters the `ChainPluginArgs` ctor kwargs but leaves this second OC-arg injection path alone. Since the flag is an explicit user request it's defensible to let it fail, but a clear early error (or warn-and-ignore) in `setupClusterConfig` would be consistent with the PR's goal, and the support information is already computed two lines away.

### 5. Smaller items

- `supportedPluginArgs` silently drops explicitly-passed `--sys-vm-oc-cache-size-mb` / `--sys-vm-oc-compile-threads` on non-OC builds (argparse defaults make explicit-vs-default indistinguishable). A one-line `print` when a key is actually dropped would turn silent surprise into a visible note.
- `nodeopSupportsOption`: if the nodeop binary is missing entirely, `subprocess.run` raises `FileNotFoundError`, escaping unwrapped — the caller gets a raw traceback at module import instead of the carefully worded `RuntimeError` built for the nonzero-exit case. Wrap in `try/except OSError` and re-raise the same `RuntimeError`.
- `testUtils.py:94`: the `_check_output_lock` comment ("lock to serialize writes to subprocess_results.log") is now stale — the lock also guards the help cache. Update the comment or give the cache its own lock.
- The probe regex `(^|\s)<opt>([=\s]|$)` is safe today but would false-positive if a future option description mentions a literal `--sys-vm-oc-...` flag. Anchoring to the option-line shape (`re.MULTILINE` + `^\s*`) makes it immune to description text.
- The `ocSupported`/`ocRequested`/skip logic is duplicated verbatim across both tests — and the two copies already drifted once (item 1). A small shared helper (e.g. in `Utils` or `TestHelper`) would keep them in lock-step.
- `+= args.sys_vm_oc_enable` is no longer guarded by the old truthiness check; safe today because the AppArgs default is `"auto"`, but a `None` default would turn this into a `TypeError` on concat.
- `ocRequested` naming: `none` is also a "request"; the variable means "OC execution expected on the configured node" — `ocExpected`/`ocEnabled` would read better.
- `universal_newlines=True` → `text=True` is the modern spelling.
- The interrupt test still prints the generic `search:` message before branching while read_only prints branch-specific "(expect absent)/(expect present)" messages — worth harmonizing when fixing item 1.

### Pre-existing issues noticed nearby (NOT from this PR)

- `performance_test_basic.py:129`: `_apiNodeIds` end bound uses `validationNodeCount` instead of `apiNodeCount` — `range(P+V, P+V+V)` should be `range(P+V, P+V+A)`. Wrong API node IDs whenever the counts differ (consumed at lines 149 and 418); today's defaults mask it.
- `performance_test_basic.py:88`: `re.search("\w*PluginArgs", ...)` emits a SyntaxWarning; should be a raw string.
- `read_only_trx_test.py:346-348`: `assert((allOC and found) or not found)` in `basicTests()` checks producerNode, but `all` is only ever applied to the apiNode, so the `allOC` half is unreachable — same wrong-node theme as item 1.

### Validation gap

The PR body demonstrates the macOS skip path and `py_compile` only. Not yet demonstrated: the newly-reachable run path on non-OC builds (these tests previously exited early on non-Linux for any OC value — now they actually run with OC args stripped), and no-regression on Linux OC builds, especially `read-only-trx-parallel-no-oc-if-test`, which now exercises a strict assertion that was previously a tautology (`found or (noOC and not found)` ≡ `found or noOC`). Suggest `ctest -R "read-only-trx|interrupt-read-only-trx" --output-on-failure` on a Linux OC build before merge.

### What looks good

- Capability detection from the actual built binary is the right call, and the probe implementation is careful where it counts: cached, lock-guarded initialization, stderr folded into stdout, explicit error with command and output on failure, and a suffix guard that correctly rejects prefix matches.
- The kwargs filter drops only the two conditionally-compiled fields — unknown kwargs still raise `TypeError`, so dataclass typo-detection is preserved (and the docstring says so).
- Commit 0db6efca42's apiNode/producerNode split converts a tautological assertion into two meaningful one-sided assertions, with a comment explaining which node carries which setting.
- The old read_only skip path used `exit(True)` — exit code 1 — under a "Do not fail the test" comment, so the macOS "skip" actually failed the test; this PR quietly fixes that too.
- Gating `verifyOcVirtualMemory()` on `ocRequested` keeps the `/proc/<pid>/statm` read (Linux-only, OC-only semantics) off non-OC paths.

heifner (approved):

PR #382, opened 2026-06-06
Opp: node-owner dispute vote for conflicting envelopes

heifner:
## Summary

Adds the node-owner dispute vote that resolves conflicting OPP outpost envelopes, filling the placeholder `sysio.msig` integration scaffolded in `sysio.chalg`.

When the active batch-operator group delivers 3+ distinct envelope versions for one (outpost, epoch) with no majority — a Byzantine split the automatic consensus rules can't resolve — `sysio.msgch::evalcons` opens a dispute in `sysio.chalg` and pauses the epoch. The Tier-1 node owners vote on the canonical envelope; once a checksum wins, the winning envelope is dispatched and the epoch unpauses.

## Unifying principle

There is always one canonical envelope per (outpost, epoch) — chosen by majority when one exists, otherwise by the Tier-1 vote. Every operator that delivered a non-canonical checksum is slashed (bond → LP); an operator that delivered nothing is never slashed (it stays on the existing miss → terminate ladder).

## Tally

`N = sysio.system::nodecount.t1_count`, `Q = floor(N/2)+1`. Fast path (any time): a checksum with `votes >= Q` wins. After the 24h deadline the denominator relaxes to a quorum of cast votes (`cast >= Q`) plus a strict majority of cast (`2*votes > cast`). No plurality or tie-break.

## Changes

- **types.proto** — `DisputeStatus` enum.
- **sysio.chalg** — `dispute` / `disputevote` tables; `opendispute` (msgch-auth), `votedispute` (Tier-1 eligibility via `sysio.roa::nodeowners`), `chkdispute` (tally → dispatch winner via `msgch::resolvedisp` → unpause). `slashop` auth widened to accept `sysio.epoch`.
- **sysio.msgch** — `winning_checksum` on `outpcons`; the consensus-reached body is factored into `apply_consensus`, shared by the majority and dispute paths so the winning envelope's messages are dispatched in both; `resolvedisp`; the 3-way-split trigger plus an open-dispute gate in `evalcons`; an open-dispute gate in `chkcons`.
- **sysio.epoch** — `advance` slashes every non-canonical deliverer in one pass, deduped across outposts so an operator non-canonical on multiple outposts is slashed once.
- **sysio.roa** — `nodeowners` / `roastate` made public so `sysio.chalg` can read them for the Tier-1 voter lookup.

## Notes for review

- Two consensus-safety design points worth scrutiny: the slash is a **single path** in `advance` for both the majority and dispute cases — slashing in both `chalg::enforce` and `advance` would double-slash a disputed operator, and `opreg::slash` aborts on an already-slashed op, which would stall the epoch. And the winning envelope **is** dispatched on resolution (via `resolvedisp`), so the disputed epoch's cross-chain messages are not dropped.
- The legacy challenge scaffolding (`initchal` / `escalate` / `submitres` / `enforce` + `challenges` / `resolutions` tables) is superseded by the dispute vote and left in place; a follow-up can remove it.
- Contract tests live in `contracts/tests/sysio.dispute_tests.cpp`; a system-level e2e (`flow-batch-operator-slashing`) is in progress in wire-tools-ts.

## Follow-up: composite-key fix + test hardening (`ada34eacd9`)

Review of this PR surfaced a latent correctness bug in the `(outpost, epoch)` secondary-index key plus a few test/clarity gaps, addressed in a second commit:

- **uint128 composite key.** `(outpost, epoch)` buckets were keyed as `(uint64(chain_code) << 32) | epoch_index`. `chain_code` is a `slug_name` value (range `[0, 2^48)`, leading symbol at bits 42-47), so the shift truncated its high bits and distinct outposts collided on one key (every 1-2 char code mapped to the epoch alone) — cross-wiring envelope and dispute buckets in any multi-outpost deployment. Pre-existing in the envelope tables and now also reached by the dispute tables, so fixed across the board: widened to `uint128` via a shared `sysio::opp::outpost_epoch_key` helper, applied to the `byoutepoch` indexes (`sysio.msgch` envelope + outbound, `sysio.chalg` dispute) and every lookup (`deliver` / `evalcons` / `apply_consensus` / `resolvedisp` / `opendispute` / `advance`). `uint128` secondary keys are the existing idiom here (e.g. `sysio.reserv::by_chain_token`).
- **Double-slash invariant** documented at the `advance` slash site (a slashed op is dropped from scheduling by the ACTIVE-only filter, so `advance` never retries the `opreg::slash` that would abort on an already-slashed op), with a regression test covering that filter, the widened `slashop` auth, and the already-slashed guard.
- **chkdispute tests** added for the post-deadline relaxation and the deadline tie (previously only the fast path and sub-quorum wait were covered).
- **`DisputeStatus` / `NodeOwnerTier`** registered with `FC_REFLECT_ENUM` in `opp.hpp`; minor comment fixes (audit-only candidate operators, `roastate`, the open-dispute pause gate).


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -719,6 +720,187 @@ void dispatch_attestation(name self, uint64_t attestation_id,
...
+/// `raw` is the winning envelope's bytes; `winning_checksum` is its sha256 — recorded on `outpcons`
+/// so `sysio.epoch::advance` can classify each operator's delivery as canonical or slashable.
+void apply_consensus(name self, uint64_t chain_code, uint32_t epoch_index,
+                     const std::vector<char>& raw, const checksum256& winning_checksum) {
+   const uint128_t composite = opp::outpost_epoch_key(chain_code, epoch_index);
+
+   // Idempotency guard: if an INBOUND message for this (chain_code, epoch) is already on file the
+   // envelope was processed already (post-quorum late delivery, or a re-fired resolvedisp). The
+   // late delivery is a benign no-op.
+   {
```
huangminghuang, line 745: This checks messages for idempotency, but the same function erases that row at line 843 (line 843). After consensus, raw envelope bytes are cleared from existing delivery rows at line 837 (line 837). If a remaining batch operator delivers later, evalcons groups the old majority rows, copies empty raw_data for the winning checksum at line 1038 (line 1038), then calls apply_consensus again at line 1083 (line 1083), which decodes empty bytes and aborts. That makes “benign late delivery” fail, and a late non-canonical operator gets no persisted delivery row for epoch::advance to classify/slash. The idempotency guard should probably key off durable consensus state, e.g. outpcons/envlog, before decoding.

heifner: Good catch — confirmed. `apply_consensus` erases its `messages` row and clears `raw_data` from the delivery rows in the post-consensus cleanup, so the messages-based guard never fired on a re-invocation; a post-quorum re-fire then re-grouped the emptied winning row and aborted on an empty-envelope decode, reverting the late delivery (and the row `advance` needs).

Fixed in 27d6eabbf1 by keying the guard off the durable `outpcons` row. I used `outpcons.epoch_index` rather than `consensus_reached` because `chkcons` resets `consensus_reached` on every advance, whereas `epoch_index` is written only by `apply_consensus` and never rolled back — so `epoch_index == this epoch` reliably means "already processed". The no-op returns before any cleanup, so the late delivery's own envelope row persists for `epoch::advance` to classify/slash. Added a regression test (re-fire evalcons after consensus → benign no-op).

huangminghuang (approved):

PR #384, opened 2026-06-08
Opp: remove superseded sysio.chalg challenge flow

heifner:
## Summary

Follow-up to the node-owner dispute vote (#382). That PR added the Tier-1 dispute vote — the actual implementation of "resolve a contested epoch", the concept the old `sysio.chalg` challenge/response/manual-msig scaffolding was a placeholder for. The old flow was never wired up (stub `msig_proposal = name(0)`, `// TODO` lines) and had no callers, so it's removed here.

## Changes

- **sysio.chalg** — remove the dead actions (`initchal`, `submitresp`, `escalate`, `submitres`, `enforce`) and the `challenges` / `resolutions` tables, plus the now-unused `MSIG_ACCOUNT` / `UWRIT_ACCOUNT` constants, the `ChallengeStatus` alias, and `MAX_AUTOMATIC_ROUNDS`. `slashop` stays — repurposed as the slash chokepoint for the dispute + epoch path.
- **sysio.msgch** — update the `CHALLENGE_REQUEST` / `CHALLENGE_RESPONSE` dispatch comment, which referenced the removed entry points.
- **contracts/tests/sysio.chalg_tests.cpp** — removed; it only exercised the deleted actions. Chalg coverage now lives in `sysio.dispute_tests.cpp`.

The `ChallengeStatus` proto enum is left in `types.proto` (removing it would force a cross-repo OPP-bundle regeneration, out of scope here).

PR #386, opened 2026-06-09
Opp: sysio.msgch perf + robustness cleanup

heifner:
Three small perf/robustness cleanups to sysio.msgch, re-ported from the stale #328 (a PR #317 follow-on) onto the current dispute-vote contract. Stacked on #384.

**deliver -- defer error-string build to failure path.** The env_epoch mismatch check ran a triple std::to_string concat on every successful delivery. Switch to the sysio::check lambda overload so the message is built only on failure.

**apply_consensus -- drop the vestigial messages emplace+erase.** The raw envelope was emplaced into a messages_t row (raw_payload up to MAX_ENVELOPE_BYTES) and erased again later in the same action, for zero net storage. Idempotency already keys off the durable outpcons row, not the messages row, so nothing reads it. Drop the pair; keep the message_entry schema so a future consumer can reintroduce the write. Also removes the dead get_message test helper.

**buildenv -- pop-and-resize on size overflow.** The previous flow marked attestations PROCESSED before serialising and used a hard size check as a backstop; a trip rolled the whole action back, and the next epoch re-packed the same set into the same trip -- a dead-letter loop under estimator drift. Replace it with a trim loop that pops the last entry and re-serialises until under MAX_ENVELOPE_BYTES, deferring PROCESSED marking until the loop converges. Popped entries stay READY for the next epoch. Self-correcting under drift.

Regenerates sysio.msgch.wasm; the ABI is unchanged.


contracts/sysio.msgch/src/sysio.msgch.cpp
huangminghuang, line 748: This can permanently stall epoch advancement after an emissions block.
apply_consensus treats outpcons.epoch_index == epoch_index as an unconditional no-op. But chkcons clears consensus_reached before calling sysio.epoch::advance at line 1143 (line 1143), and advance can legally return without advancing when emissions are not ready at contracts/sysio.epoch/src/sysio.epoch.cpp (line 287). After that path, the chain is still on the same epoch, outpcons.consensus_reached == false, and future evalcons refires return at the idempotency guard before re-arming consensus. chkcons will then keep seeing !all_consensus and never retry advance, even once emissions become ready.
A fix needs to preserve “already dispatched” separately from “ready to advance”, or re-set consensus_reached = true on the same-epoch no-op path when it was cleared by a non-advancing advance.

heifner: Good catch -- confirmed and fixed in `bdf7a5c129`.

You're exactly right that `epoch::advance`'s emissions gate returns gracefully rather than throwing, so the pre-emptive `consensus_reached = false` in `chkcons` committed, and nothing re-armed it -- `apply_consensus` is the only writer and it doesn't re-fire once the delivery set is complete.

I took the "preserve *already-dispatched* separately from *ready-to-advance*" option: `chkcons` now gates on the durable `opc.epoch_index == current_epoch` signal (written by `apply_consensus` on dispatch, never cleared) and the premature reset is dropped. A stale row from a prior epoch fails the `epoch_index` check, so a successful advance still invalidates old consensus; a non-advancing advance leaves the row intact and the next `chkcons` re-attempts. `consensus_reached` stays as an audit field (no ABI change).

I didn't take the "re-set `consensus_reached` on the same-epoch no-op path" route because that path lives in `apply_consensus`, which never re-fires after a complete delivery set -- so it wouldn't trigger.

Added a regression test `sysio_dispatch_tests/chkcons_survives_non_advancing_advance`: after consensus, two `chkcons` calls across a gate-blocked advance must each re-attempt (`blocklog.retry_count` increments both times). I verified it fails on the pre-fix wasm (`retry_count 2 != 3`) and passes after.

huangminghuang (approved):

PR #387, opened 2026-06-09
Opp: read inbound envelopes at finalized commitment

heifner:
The Solana outpost client read inbound envelopes at `commitment_t::confirmed` and the Ethereum outpost client at `block_tag_latest`. WIRE consensus on inbound is committed forward against these reads, so a source-chain reorg below the read commitment leaves WIRE state derived from external history that no longer exists: an operator can reach WIRE-side consensus on a slot read at `latest`/`confirmed` and queue attestations off it, then watch that slot reorg away.

Read both at `finalized` -- the only commitment with cryptoeconomic finality on each chain (~12.8 min ETH, ~13-31 s SOL), which cannot be rolled back.

Intentionally hard-coded, not operator-configurable. The read commitment is effectively a consensus parameter: operators reading the same epoch at different commitments would deliver divergent envelopes and manufacture disputes among honest operators, and there is no safe production value other than `finalized`. (#328 proposed making these configurable with the unsafe values as defaults for upgrade-compat; OPP is pre-launch, so there is nothing to stay compatible with.)

Also adds `block_tag_finalized` to the Ethereum client header and a little-endian `static_assert` on the Solana client's `read_u32_le`. Scope is the inbound-consensus reads only; the outbound `epoch_in` tx-confirmation commitment is unchanged.

Stacked on #386.


huangminghuang (approved):

libraries/libfc/include/fc/network/ethereum/ethereum_client.hpp
```diff
@@ -59,6 +59,15 @@ constexpr std::string_view block_tag_pending = "pending";
...
 constexpr std::string_view block_tag_latest = "latest";
 
+/**
+ * @brief Block tag constant representing the most recent beacon-chain-finalized block.
+ *
+ * Unlike `latest` (canonical head, routinely reorged) this tag only advances on beacon-chain
+ * finality (~2 epochs, ~12.8 min), which carries cryptoeconomic finality and cannot revert without
+ * a third of staked ETH being slashed. Required for reads that downstream consensus commits against.
+ */
+constexpr std::string_view block_tag_finalized = "finalized";
```
heifner: Lets use enum instead of constexpr

heifner: Done in 89022ce392 — `enum class block_tag_t { earliest, finalized, safe, latest, pending }` (the full execution-apis closed set), mirroring the Solana side's `commitment_t`: the enumerator spelling is the wire format, converted via `magic_enum::enum_name`, with a unit test pinning the spellings. The old variant alias is now `block_number_or_tag_t` and flows through all block-parameter signatures, so the read site passes `eth::block_tag_t::finalized` directly — parallel to `commitment_t::finalized` on the Solana read.

huangminghuang (approved):

PR #391, opened 2026-06-10
Fix: consensus-review findings from the Wire-vs-Spring block-validation audit

heifner:
Fixes the confirmed findings from the 2026-06-09 Wire-vs-Spring block-validation / transaction-execution review (Wire deltas audited against the Spring fork point `e6a99f68b6`). No chain-split defect was found; these address host-intrinsic correctness, snapshot-load robustness, determinism of auxiliary state, and an OC cache-versioning hazard.

- **ed25519 assert_recover_key vs recover_key disagreement** -- assert_recover_key verified over the raw digest while recover_key and trx auth verify over the hex-encoded digest, so every SDK-signed ed25519 signature passed one intrinsic and failed the other. Both now share the public_key::recover path, with agreement pinned by intrinsic-probe tests driving both intrinsics with identical payloads.
- **Permission-depth snapshot DoS** -- snapshot load rejected parent chains deeper than 6 even though creation never caps depth, so an honest deep chain produced snapshots no node could restart from. The walk now detects only genuine corruption (cycles, dangling parents) and additionally validates authority structure on load, exempting the deliberately-empty genesis authorities.
- **sys-vm-oc stale code cache** -- the intrinsic-table renumber kept upstream cache version tags, so a stale code_cache.bin passed compatibility checks and dispatched to wrong host functions; cache magic and codegen version bumped.
- **transaction_dedup path-dependence** -- recording was gated on node-local pending LIB time and expiry pruning was insertion-order-bounded, so honest nodes at the same block could disagree on integrity hash and snapshot content. Recording is now unconditional and the index is (expiration, id)-sorted, making the serialized set canonical across replay, live sync, and fork switches.
- **kv iterator non-advance** -- erase + reinsert at the same key mid-iteration made kv_it_next/kv_idx_next return the same position forever; slow-path re-seek now uses upper_bound.
- **contract_action_match prefix rules** behaved as exact (latent).
- Minor findings: finalizer-policy log format, deep-mind RAM events from read-only transactions, platform_timer set_expired generation stamping, newaccount ram_delta narrowing, KV doc-comment ABI corrections, chaindb allocate(0) out-of-bounds and preallocate remnant stranding, and test_fc source-root discovery in git worktrees.


libraries/chain/transaction_dedup.cpp
```diff
@@ -187,11 +208,14 @@ void transaction_dedup::read_from_snapshot(const snapshot_reader_ptr& snapshot)
          snapshot_transaction_dedup_entry entry;
          more = section.read_row(entry);
          map_.emplace(entry.trx_id, entry.expiration);
```
huangminghuang: This ignores failed inserts while loading snapshot rows. A corrupted snapshot can contain the same trx_id twice, especially with different expirations: map_.emplace() keeps only the first entry, but index_.emplace_hint() can still add the second (expiration, id). That breaks the new map/index invariant, makes size() disagree with serialized contents, and later clear_expired() can leave phantom index entries or allow the same transaction id to be recorded again while an old index row remains. Snapshot/file load should assert both inserts succeed and reject duplicate transaction ids.

libraries/chain/authorization_manager.cpp
```diff
@@ -81,25 +81,40 @@ namespace sysio { namespace chain {
                              "Permission {}@{} contains unactivated key type", row.owner, row.name);
                }
 
+               // Reject structurally invalid authorities: unsorted or duplicate keys/accounts, or a
+               // threshold no weight combination can satisfy. newaccount and updateauth enforce the
+               // same predicate at creation, so any non-empty row failing it is corrupt or
+               // hand-crafted. Deliberately EMPTY authorities (no keys and no accounts) are exempt:
+               // genesis creates sysio.null's permissions and sysio.prods@owner with empty_authority,
+               // permanently unsatisfiable by design, and validate() rejects those as threshold-unreachable.
+               const bool deliberately_empty = row.auth.keys.empty() && row.auth.accounts.empty();
```
huangminghuang: This exempts every empty authority from validate(row.auth), not just the known genesis empty authorities. newaccount/updateauth reject these via validate(), so a hand-crafted snapshot can still load an arbitrary account@permission with threshold > 0 and no keys/accounts, bypassing the same structural checks this PR is adding. The exemption should be scoped to the deliberate genesis cases, e.g. sysio.null permissions and sysio.prods@owner.

libraries/chain/transaction_dedup.cpp
heifner: Fixed in 8dc37cf28e. Snapshot load now rejects a repeated transaction id with `snapshot_exception` instead of ignoring the failed map insert, and asserts `map_.size() == index_.size()` after the section is consumed (map uniqueness keys on the id alone, so equal sizes prove every index insert landed as well). The dedup-file path inherits the check through `read_from_file`'s existing catch: a corrupt file warns, resets, and returns false. Tests cover the duplicate id under both different and identical expirations, plus the file path.

libraries/chain/authorization_manager.cpp
heifner: Fixed in 8dc37cf28e. The exemption is now scoped to exactly the three rows genesis creates unsatisfiable (sysio.null@owner, sysio.null@active, sysio.prods@owner) and requires exactly the genesis value `authority(1, {}, {})`, now shared as `genesis_empty_authority()` between controller genesis code and this loader. Any other empty authority -- or any other value on those rows -- fails `validate()` like any corrupt row. Test covers an empty authority on a normal account, on sysio.prods@active, and a non-genesis threshold on sysio.null@owner.

huangminghuang (approved):

PR #392, opened 2026-06-10
sys-util: SHiP log inspection/repair/split/merge tooling + nodeop state-history-force-write

heifner:
Gives node operators in-box tooling for state history (SHiP) logs — the capabilities requested in AntelopeIO/leap#1664, motivated by AntelopeIO/leap#1593 (corruption at block 112M with 8.1 TB of salvageable history after it, and no recovery short of a multi-month replay).

## Problem

`state_history_log`'s constructor auto-mutates: merely opening a log truncates a corrupt tail, regenerates a bad index, and vacuums pruned logs, so a damaged log cannot even be inspected without modifying it — and only via a full nodeop. Anything beyond tail corruption ("is corrupted and cannot be repaired") is fatal, discarding everything after the first bad byte even when terabytes of valid data follow.

## sys-util ship-log

All subcommands operate on a `--state-history-dir` / `--log` pair, accept retained `<stem>-N-M` bundles, and the read-only ones really are read-only (the scanner never goes through `state_history_log`).

| Subcommand | Writes | Purpose |
|---|---|---|
| `info` | no | per-bundle endpoint report: version, pruned flag, block range, index health |
| `smoke-test [--deep]` | no | validate every entry (headers, position trailers, block continuity, index agreement); `--deep` also decompresses every payload (adler32 catches bit rot); prints a map of valid/damaged regions |
| `make-index` | index | standalone index rebuild; sequential forward walk, byte-identical to the library's backward regeneration |
| `trim --first/--last` | yes | trim to a block range; end trim is an in-place truncate, front trim rewrites position trailers via temp files + atomic rename |
| `extract-blocks` | new bundle | copy a block range into a fresh bundle, source untouched |
| `repair [--keep-tail --dry-run --deep --output-dir]` | yes | default truncates at first damage + rebuilds the index (offline, previewable equivalent of nodeop's auto-recovery); `--keep-tail` salvages the last valid range instead — the leap#1593 case |
| `vacuum` | yes | pruned → non-pruned |
| `split --stride` | new bundles | rotation-compatible retained bundles (boundaries on stride multiples) plus head remainder; loads directly into a `log_catalog` |
| `merge` | new bundle | merge contiguous retained bundles back into one log |

## nodeop --state-history-force-write

Never let damaged or inconsistent ship logs stop the node. An index that disagrees with its log is regenerated instead of fatal; a head log that fails its startup checks or cannot accept the next block (a gap after a snapshot restore, divergent fork history) is renamed aside to `<stem>-corrupt-<n>` — kept on disk, never deleted — and writing continues into a fresh log, escalating to setting aside the retained bundles only when a write predates the whole catalog. Holes in the retained set are tolerated with a warning, leaving just those blocks unserved. Orphaned bundles remain valid logs that `ship-log` can inspect, repair, and trim (and merge back once renamed off the `-corrupt-` suffix).

## Implementation notes

- Primitives live in `libraries/state_history/log_utils.{hpp,cpp}` next to the format they understand; the CLI is a thin binding in `programs/sys-util/actions/shiplog_actions.{hpp,cpp}` following the `block-log` action pattern.
- Entry position trailers store absolute file offsets, so every operation that moves entries (front trim, extract, split, merge, keep-tail) rewrites them — and must walk every entry including fork-superseded ones, not just the index's slots.
- After damage, the scanner resynchronizes by searching for the 8-byte ship magic and accepting a candidate only if its position trailer points back at it exactly, making a false resync on payload bytes practically impossible.
- A bundle salvaged from mid-log must start at its "canonical" entry (the earliest one not superseded by a later fork switch); starting at a stale fork entry would produce a log whose first block is not a floor for the rest, which `state_history_log` cannot open safely.
- A write below the whole catalog (chain wiped and resynced) is the one case force-write resolves by setting aside the retained bundles too; plain head-gap and fork-mismatch cases only rotate the head log out.

## Validation

- `tests/ship_log_utils.cpp` — 29 cases: torn tails, mid-file damage, payload bit flips, fork-overwritten entries, pruned logs, index corruption variants, trim/extract/split/merge round-trips all verified by reopening the results through real `state_history_log`/`log_catalog` instances, and every force-write tier.
- `tests/ship_log_util_test.py` — end-to-end against a real nodeop: corrupt, repair with sys-util, byte-compare against pristine files, relaunch, plus both force-write scenarios (lying index regenerated; gapped log orphaned while the head advances).
- Against a production 61 GB / 8.05M-block chain_state log: full structural scan + slot-by-slot index verification in ~3 minutes; deep payload validation of a 3.5 GB trace log in 38 s; 100-block extract from the 61 GB log in 37 ms; split→merge of the 3.5 GB log round-trips to a byte-exact prefix of the original; corrupt-then-repair cycles left both the 61 GB log and its rebuilt index byte-identical to a pristine backup.


libraries/state_history/log_utils.cpp
```diff
@@ -0,0 +1,1227 @@
...
+              "pruned log {}.log has a trailing block count {} reaching before its first block {}", stem.string(),
+              *e.pruned_count, ci.index_first);
+   grow_slots(ci.slots, ci.index_first, ci.last_block, stem);
+
+   window_reader   w(log, size);
+   progress_ticker ticker(progress, size);
+   uint64_t        next_begin = size - sizeof(uint32_t); //walking backward: where the entry after the current one starts
+   uint64_t        pos        = e.last_entry_pos;
+   while(true) {
+      const entry_check c = check_entry(w, log, pos, size, std::nullopt, std::nullopt, false);
```
huangminghuang: This validates the trailer chain, then immediately indexes ci.slots[c.block_num - ci.index_first]. A damaged pruned log can still contain a structurally valid entry whose decoded block number is below ci.index_first or above ci.last_block; that underflows or goes out of range before the tool can report “log damaged”. Since this PR is adding corruption-inspection/repair tooling, build_index, full check_index, and find_block_id should fail cleanly here. Add an explicit range check before line 403, or pass a floor into check_entry and also reject c.block_num > ci.last_block.

huangminghuang (commented):
I’d suggest a few non-blocking improvements:
- Add one focused regression test for the pruned-log out-of-range case
The current pruned tests cover a damaged index/trailer path, but not a structurally valid entry with a bogus block number. That’s the exact shape that can hit unchecked ci.slots[c.block_num - ci.index_first] in log_utils.cpp (line 403).

- Centralize the pruned backward-walk validation
compute_index_pruned() and the pruned branch of scan_log() both walk the trailer chain and validate similar invariants. They diverge slightly: scan_log() reports damage, while compute_index_pruned() can index before checking the block range. A small shared helper that yields validated entries, or at least a shared “validate pruned entry range” helper, would reduce future drift.

- Make the force-write behavior extra explicit in docs/help text
state-history-force-write is powerful: it can orphan logs and skip writing bad blocks so the node keeps running. The option text is already pretty clear, but I’d consider explicitly saying this is an emergency/recovery option and may leave state-history incomplete for affected blocks.

- Add a CLI test for ship-log block-id against a damaged pruned log
The library has good tests for find_block_id, and the Python test covers the CLI broadly. A specific test that asserts the CLI reports damage instead of crashing would lock in the intended operator behavior.

libraries/state_history/log_utils.cpp
heifner: Good catch — fixed in 8a52ba75b0, and it turned out to bite **valid** logs too, not just damaged ones.

A deep fork switch in a pruned log's retained tail legitimately leaves superseded entries whose block number is **above** the head — exactly the library's own documented *"log contains 2345675 (begin=2, end=6, but we see block 7 and 6 when reading)"* case in `state_history_log::check_index_on_init`. So the unguarded `ci.slots[c.block_num - ci.index_first]` could run off the end of the slot vector on a perfectly healthy pruned log, not only on a corrupt one.

Because of that the two sides need different handling, and I ended up deviating from the literal "reject `c.block_num > ci.last_block`" suggestion:

- **Above `last_block` → skip, don't reject.** These are valid fork-superseded entries; the library skips them (`if(read_block_num < _end_block)`) when it regenerates the index. Rejecting them as damage would make the tool refuse logs the node opens fine.
- **Below `index_first` → reject as damage.** Nothing is ever written before the log's first block, so this can only be corruption. The index floor is now passed into `check_entry`, which stops the walk and reports it.

`build_index`, full `check_index`, and `find_block_id` therefore fail cleanly (throw / `log_damaged`) on the below-floor case and produce a correct, in-bounds index on the above-head case. Both are now driven through a single `walk_pruned_chain` helper (also addresses the "centralize the pruned backward-walk" point).

New coverage:
- `pruned_index_skips_superseded_above_head` — asserts our `build_index` output is **byte-identical** to the library's regenerated index on a deep-reorg pruned log.
- `pruned_out_of_range_block_num_is_damage` — a structurally valid entry forged to decode below the first block is reported as damage; verified to SIGSEGV without the floor guard.

libraries/state_history/log_utils.cpp
```diff
@@ -0,0 +1,1251 @@
...
+            if(is_ship(hdr.magic) && is_ship_supported_version(hdr.magic) &&
+               chain::block_header::num_from_id(hdr.block_id) == block_num)
+               return hdr.block_id;
+         }
+      }
+   }
+
+   const computed_index ci = e.pruned ? compute_index_pruned(log, size, e, stem, progress)
+                                      : compute_index_forward(log, size, stem, progress);
+   return log.unpack_from<log_header>(ci.slots[uint64_t(block_num) - ci.index_first]).block_id;
```
huangminghuang: This falls back from a distrusted index to compute_index_pruned()/compute_index_forward(), then unpacks whatever offset is in the computed slot. For pruned logs, the computed index can legitimately contain zero slots below the servable range, and a damaged retained chain with a missing/duplicate block inside the servable range can also leave a requested slot at zero. In that case find_block_id() reads offset 0 and can return the first stub header’s id for the wrong block instead of reporting damage/null. The fast index path already validates “slot points at the requested block”; the fallback should do the same check before returning.

heifner: Good catch — fixed in 8068ad784f.

The index fast path (and `position_lookup::entry_pos_of`) already read the slot's entry back and confirm it decodes to the requested block, but the fallback trusted the recomputed slot blindly. I extracted that check as a shared `slot_holds_block(log, size, pos, block_num)` helper and routed all three lookups through it (fast path, `entry_pos_of`, and the fallback) so they can't drift. The fallback now `SYS_ASSERT`s it before returning, throwing `plugin_exception` ("block N is within its servable X-Y range but has no valid entry") instead of unpacking offset 0's stub header.

On the two sub-cases you raised:

- **Below the servable range** (pruned logs carry zero slots there): `find_block_id` already returns `nullopt` for `block_num < first_servable` up front, so those slots are never reached.
- **Missing/duplicate block inside the servable range → zero slot**: this is the real hole, and it only occurs on the pruned path. The forward walk's continuity check (`check_entry`'s `prev_block`) already rejects a gap as damage before any slot is read; the pruned backward walk intentionally tolerates non-contiguous block numbers (fork-superseded entries), so a damaged retained chain can leave an in-range slot at zero there.

New coverage: `find_block_id_fallback_rejects_orphaned_servable_block` forges a retained entry to decode as its neighbour (orphaning its own slot to zero) while leaving positions intact so the trailer chain still walks cleanly, drops the on-disk index to force the fallback, and asserts `find_block_id` reports damage. Verified to return the stub's (block 2's) id rather than throw without the new guard.

(The branch was well behind master, so I also merged current master in; that's the bulk of the new diff.)

huangminghuang (approved):

PR #393, opened 2026-06-11
Preserve exception type when rethrowing stored exceptions

heifner:
Ports Spring PRs AntelopeIO/spring#1535 and AntelopeIO/spring#1669.

`throw *ptr` on a stored `fc::exception_ptr`/`optional<fc::exception>` slices to `fc::exception`, losing the derived type. This adds a virtual `fc::exception::rethrow()` (overridden by the exception-declaration macros and wrapper classes), switches `transaction_trace` rethrow sites to `std::rethrow_exception(trace->except_ptr)` -- the unsliced original -- and converts remaining `throw *exception_ptr` sites to `rethrow()`. Unit tests now catch specific exception types (e.g. `BOOST_REQUIRE_THROW(..., account_name_exists_exception)`, `interrupt_exception`), which doubles as the regression test that types survive the apply-block path.

Deviations from upstream: `unhandled_exception` also gets the `rethrow()` override (upstream only covered `std_exception_wrapper`), and three wire-only sites are converted for consistency (`subjective_billing_tests.cpp`, a second `wasm_tests.cpp` site, `no_block_extensions_allowed_test`).

Tests: full `unit_test`/`plugin_test`/`contracts_unit_test`/`test_fc` pass (only the known local-contract reference-data failures remain, identical to a clean-tree baseline); new `test_fc --run_test=exception` covers the rethrow round-trip.

### DIFF
--- a/contracts/tests/sysio.system_blockinfo_tests.cpp
+++ b/contracts/tests/sysio.system_blockinfo_tests.cpp
@@ -323,7 +323,7 @@ try {
       });
       if (!result.first.has_value()) {
          const sysio::chain::transaction_trace& trace = *result.second;
-         if (trace.except.has_value()) {
+         if (trace.except) {
             elog("{}", trace.except->to_detail_string());
          }
       }
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -970,7 +970,7 @@ struct controller_impl {
          if( trace->except_ptr )
             std::rethrow_exception(trace->except_ptr);
          if( trace->except)
-            throw *trace->except;
+            assert(trace->except_ptr); // except/except_ptr always set together
          getpeerkeys_res_t res;
          if (!trace->action_traces.empty()) {
             const auto& act_trace = trace->action_traces[0];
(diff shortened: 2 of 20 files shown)


huangminghuang (commented):
deleted

heifner:
Closing out the design question raised in this thread (`fc::exception_ptr` vs `std::exception_ptr`) for the record:

**Verdict: keep both, each for the job the other cannot do.**

- `std::exception_ptr` is used wherever an original throw exists and faithful rethrow is the goal: `transaction_trace::except_ptr` and the controller/tester rethrow sites (`std::rethrow_exception`). It also covers non-fc exceptions.
- `fc::exception_ptr` remains the carrier where errors are *data* that gets inspected (`code()`/`name()`/log for HTTP error bodies, subjective billing classification, logging) and only sometimes rethrown. `std::exception_ptr` is opaque -- every such inspection would need a throw/catch round trip -- and `fc::exception` stays regardless as the serializable error type (`error_results`, `trace->except` in SHiP/RPC output). Since this PR, every `fc::exception_ptr` in the tree is a dynamic copy created via the virtual `dynamic_copy_exception()`, so the channel preserves the derived type end to end; the slicing found in review was the `optional<fc::exception>` trace field, fixed at the source in 08a8aef426.
- The wire formats already follow "errors as data": both HTTP boundaries (clio, the kiod signing client) transport the plain `error_results` shape and mint a base `fc::exception` locally only to throw it.

The one residual hazard of the virtual-rethrow design -- a future hand-written derived exception class forgetting its `rethrow()` override and silently slicing -- is now closed mechanically: every `rethrow()` implementation asserts `typeid(*this) == typeid(T)` in debug builds, so a missing override fails loudly in any test that exercises it.

huangminghuang (approved):

PR #394, opened 2026-06-11
producer: regenerate snapshot when the snapshotted block is forked out

heifner:
Ports Spring PR AntelopeIO/spring#1538 (GH-1489). **Depends on #393 -- merge that first** (this branch is stacked on it and uses `fc::exception::rethrow()`).

Previously, if `/v1/producer/create_snapshot`'s block was forked out before becoming irreversible, the request failed with `snapshot_finalization_exception` (`pending_snapshot.hpp`). Now `create_snapshot` schedules a one-shot snapshot request carrying the caller's response handler; a forked-out pending snapshot is dropped (logged) and the request re-executes on the adopted branch, so the caller gets the regenerated snapshot. In irreversible read-mode the snapshot is created directly (cannot fork out). A duplicate `create_snapshot` for the same block now returns `duplicate_snapshot_request` instead of chaining handlers.

Wire adaptations: explicit variant copy where the scheduler fans results out to the HTTP handler (wire's `next_function::operator()` is rvalue-only), and wire's `threaded_snapshot_writer` plus its improved fork-out validity check in `pending_snapshot::finalize` are retained.

Tests: `plugin_test` snapshot scheduler suite passes; `nodeop_snapshot_forked_test.py` extended per upstream -- it now issues a blocking `create_snapshot` on the fork that loses and verifies the RPC completes with a regenerated snapshot after the network re-merges (verified passing locally).

Deviation from upstream (review finding): scheduling validation errors are thrown synchronously to the caller rather than routed through the stored callback; the stored callback is completion-only, so later execution failures cannot rethrow on the block-start path.


huangminghuang (commented):
-  /schedule_snapshot stores a callback that later rethrows on the producer thread: producer_plugin.cpp (line 1901), snapshot_scheduler.cpp (line 78), snapshot_scheduler.cpp (line 141)
The callback created for producer_plugin::schedule_snapshot is meant to rethrow synchronous validation errors to the HTTP caller. But schedule_snapshot now stores that callback in snapshot_schedule_information, and execute_snapshot invokes it later for actual snapshot execution/finalization results. If a scheduled snapshot later fails, for example a snapshot file already exists or the write/rename fails, this rethrows from the scheduler path instead of being logged/handled as before. At that point there is no HTTP caller to report to, and on_start_block only catches snapshot_execution_exception, so other snapshot/producer exceptions can escape the block-start signal path. The fix is to separate “validation error reporting” from “request completion callback”; normal /schedule_snapshot requests should not persist the rethrowing HTTP validation callback.
-  Stacked exception-preservation change still leaves producer trace failures sliced: producer_plugin.cpp (line 1078), producer_plugin.cpp (line 2744)
These paths still copy from transaction_trace::except, which is stored as std::optional<fc::exception> and has already sliced the original derived type. They should derive the fc::exception_ptr from trace->except_ptr instead, otherwise transaction ack/API failure responses can still lose the derived exception type even though this PR stack adds virtual rethrow() support.

huangminghuang (approved):

PR #396, opened 2026-06-11
transaction_dedup: drive undo via chainbase (fix restart-persistence, remove desync class)

heifner:
Rebuilds the transaction dedup as a chainbase undo participant so the database drives its undo lifecycle in lockstep with the segment indices, instead of the controller hand-pairing every dedup undo call with the database's. That manual
lockstep was the root cause of the dedup desync bugs (fork-switch and restart).

- chaindb: database::add_undo_participant hook (abstract_index participant, revision-range validated like add_index).
- transaction_dedup: unified undo-session stack keyed by chainbase revision; dedup_undo_index adapter; undo stack persisted across clean restart.
- controller / transaction_context: register the participant; remove the manual lockstep sites.
- Data path stays on the heap (O(1) lookup; churn kept out of the shared segment).

Fixes the restart-persistence bug (a fork switch after a HEAD-mode restart was rejecting the canonical fork with tx_duplicate).

Tested: transaction_dedup_tests (incl. a chainbase-database-driven undo test and a randomized oracle fuzz), block/fork/restart/snapshot/replay suites, plugin_test, and the restart + snapshot integration tests. Full unit_test sweep clean except pre-existing locally-rebuilt-WASM reference mismatches.

### DIFF
--- a/unittests/transaction_dedup_tests.cpp
+++ b/unittests/transaction_dedup_tests.cpp
@@ -1,11 +1,22 @@
 #include <sysio/chain/transaction_dedup.hpp>
+#include <sysio/chain/transaction_dedup_undo_index.hpp>
 #include <sysio/chain/exceptions.hpp>
+
+#include <chainbase/chainbase.hpp>
 
 #include <boost/test/unit_test.hpp>
 
 #include <fc/crypto/sha256.hpp>
 #include <fc/filesystem.hpp>
 #include <fc/io/json.hpp>
+
+#include <algorithm>
+#include <deque>
+#include <map>
+#include <optional>
+#include <set>
+#include <utility>
+#include <vector>
 
 using namespace sysio::chain;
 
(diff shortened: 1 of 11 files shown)


unittests/transaction_dedup_tests.cpp
```diff
@@ -149,10 +233,10 @@ BOOST_AUTO_TEST_CASE(nested_sessions) {
 BOOST_AUTO_TEST_CASE(block_commit) {
    transaction_dedup d;
 
-   d.start_block_revision(1);
+   d.add_undo_session();
    d.record(make_id(1), make_exp(100));
    d.record(make_id(2), make_exp(100));
-   d.commit_block_revision();
+
```
huangminghuang, line 239: remove trailing whitespaces

huangminghuang (approved):

huangminghuang (approved):

PR #397, opened 2026-06-12
Re-record dedup entry on sys-vm-oc tier-up interrupt

heifner:
## Summary

Fixes a latent bug in the sys-vm-oc tier-up interrupt path. **This bug also exists in Spring** -- the same `transaction_context::reset()` retry path never re-records the transaction.

When sys-vm-oc finishes compiling a contract while that contract's transaction is executing under the interpreter, it interrupts the transaction and restarts it under the optimized runtime. The restart goes through `transaction_context::reset()`, which:

1. `undo()`s the transaction's undo session -- reverting the dedup record made in `init_for_input_trx`, then
2. re-runs only `exec()`, **not** `init_for_input_trx`.

So the dedup record is never re-created: an interrupting node silently drops the transaction from its dedup set.

## Why it matters

The OC interrupt is only allowed when applying a block with explicit CPU billing (a validator/relay), and only fires if OC compilation happens to finish mid-execution -- so it is validator-only and timing-dependent. When it does fire:

- The dedup set is folded into `calculate_integrity_hash` and snapshots, so the interrupting node diverges on that auxiliary state from nodes that did not interrupt at the same block.
- A still-unexpired resubmission of the same transaction would execute again instead of being rejected `tx_duplicate`.

## Fix

Re-issue `record_transaction(...)` in `reset()` after `initialize()` re-creates the undo session, gated exactly as the original record in `init_for_input_trx` (`is_input && !is_transient()`).

## Test

`sysvmoc_interrupt_tests/oc_interrupt_preserves_dedup_record` pushes a finite-but-long `checktime_failure` action (a `bound*bound` loop) under explicit billing -- the applying-block path where the OC interrupt is allowed -- forcing an OC tier-up interrupt and restart, then asserts the transaction is still known to the dedup set after the block is produced.

The assertion can fail **only** when the bug manifests (the interrupt fired AND the record was dropped). If OC compilation happens not to interrupt on a given run, the record is trivially still present, so the test never fails spuriously (it logs when the interrupt path was not exercised). This keeps the test non-flaky -- a best-effort regression guard, as the interrupt timing is inherently machine-dependent.

## Second fix: sys-vm-oc executor paired with the wrong code cache (`7c99d6d4ec`)

The new test exposed a second latent bug, which this PR also fixes. **This bug also exists in Spring** -- `eosvmoc_tier` has the same `thread_local static` executor/memory shared across instances.

CI's asan and gcc jobs SEGV'd in `oc_interrupt_preserves_dedup_record` with the crash pc inside `code_cache.bin` while the **validating** node applied the block. Root cause: `sysvmoc_tier`'s executor and memory were process-wide `thread_local` **statics** -- one per thread, shared by every tier instance, rebuilt by whichever tier was constructed last. An executor mmaps its own code cache's backing file, and descriptors hold offsets only meaningful within that file. A validating tester runs two controllers (the validating node constructs first, the main node second), so the validating node's tier-up executions ran:

```
apply_func = main node's code_cache.bin mapping + validating node's descriptor offsets
```

The two caches usually compile the same contracts in the same order, so the wrong mapping silently executes byte-identical code. When compile-completion order diverges between the nodes (asan/gcc builds perturb timing), the offsets land on different code and the JIT executes garbage -- a wild read the OC SEGV handler correctly refuses to claim, so the process dies.

`wasm_interrupt_test` never tripped this because its transaction always fails on deadline and is never re-executed by the validating node; the new test is the first to tier-up execute a **committed** transaction on the validating node. **A running nodeop is unaffected**: one controller per process means the pairing could never mismatch (and the base `sysvmoc_runtime`'s main-thread executor/memory were already per-instance members).

Fixed by extracting `sysvmoc::thread_exec_mem`, owned per tier/runtime instance:

- **Main thread**: executor/memory are instance members, so pairing with the instance's code cache is guaranteed by construction.
- **Read-only threads**: one `thread_local` slot per thread, tagged with a monotonic owner id (never reused, so a recycled `this` pointer cannot alias a destroyed instance) and rebuilt whenever a different instance executes on that thread. This also fixes the same latent mismatch in the base `sysvmoc_runtime`'s read-only-thread statics.

Verified against a RelWithDebInfo + ASan build matching CI's asan job: `sysvmoc_interrupt_tests` under `--sys-vm` crashed 5/5 runs before the fix and passes 10/10 after; `--sys-vm-jit`/`--sys-vm-oc` variants and the `sysvmoc_limits`, `checktime`, and `block_tests` OC suites pass unchanged.

### DIFF
--- a/unittests/sysvmoc_interrupt_tests.cpp
+++ b/unittests/sysvmoc_interrupt_tests.cpp
@@ -56,4 +56,76 @@ BOOST_AUTO_TEST_CASE( wasm_interrupt_test ) { try {
 #endif
 } FC_LOG_AND_RETHROW() }
 
+// Regression for the sys-vm-oc interrupt dropping a transaction's dedup record. When oc tier-up
+// completes mid-execution of a transaction applied as part of a block (explicit cpu billing), the
+// transaction is restarted via transaction_context::reset(), whose undo() reverts the dedup record
+// made in init_for_input_trx; the retry re-runs only exec(), so without the fix the record is never
+// re-created and the transaction silently drops out of the dedup set on the interrupting node.
+//
+// The action is finite-but-long: long enough under the interpreter that oc compile completes and
+// interrupts+restarts it, but it completes under sys-vm-oc and the transaction is committed. We then
+// assert the transaction is still known. This assertion can fail ONLY when the bug manifests (the
+// interrupt fired AND the record was dropped); if oc compile happens not to interrupt this run, the
+// record is trivially still present, so the test never fails spuriously.
+BOOST_AUTO_TEST_CASE( oc_interrupt_preserves_dedup_record ) { try {
+#ifdef SYSIO_SYS_VM_OC_RUNTIME_ENABLED
+   fc::temp_directory tempdir;
+   constexpr bool use_genesis = true;
+   savanna_validating_tester t(
+      tempdir,
+      [&](controller::config& cfg) {
+         cfg.sys_vm_oc_whitelist_suffixes.insert("testapi"_n);
+         if (cfg.wasm_runtime != wasm_interface::vm_type::sys_vm_oc)
+            cfg.sysvmoc_tierup = chain::wasm_interface::vm_oc_enable::oc_auto;
+      },
+      use_genesis
+   );
+   if( t.get_config().wasm_runtime == wasm_interface::vm_type::sys_vm_oc ) {
+      return; // eos_vm_oc does not tier-up; no interrupt path (see wasm_interrupt_test)
+   }
+   t.execute_setup_policy( setup_policy::full );
+   t.produce_block();
+
+   t.create_account( "testapi"_n );
+   t.set_code( "testapi"_n, test_contracts::test_api_wasm() );
+   t.produce_block();
+
+   const auto pre_count = t.control->get_wasm_interface().get_sys_vm_oc_compile_interrupt_count();
+
+   // Build a finite checktime action (bound^2 inner iterations) and push it with explicit cpu
+   // billing so it takes the applying-block path where the oc interrupt is allowed.
+   signed_transaction trx;
+   action act;
+   act.account = "testapi"_n;
+   act.name = action_name(WASM_TEST_ACTION("test_checktime", "checktime_failure"));
+   act.authorization = vector<permission_level>{{"testapi"_n, config::active_name}};
+   act.data = fc::raw::pack( (unsigned long long)15000 );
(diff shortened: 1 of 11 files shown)


unittests/sysvmoc_interrupt_tests.cpp
```diff
@@ -56,4 +56,76 @@ BOOST_AUTO_TEST_CASE( wasm_interrupt_test ) { try {
...
+                                                        transaction_metadata::trx_type::input );
+   auto trx_meta = fut.get();
+   cpu_usage_t billed_cpu_us;
+   billed_cpu_us.insert( billed_cpu_us.end(), total_actions, 1000u ); // explicit, within [min, max]
+   auto res = t.control->test_push_transaction( trx_meta, fc::time_point::now() + fc::seconds(60),
+                                                fc::seconds(60), billed_cpu_us, /*explicit_bill*/ true );
+   if( res->except_ptr ) std::rethrow_exception( res->except_ptr );
+   if( res->except ) throw *res->except;
+
+   const auto post_count = t.control->get_wasm_interface().get_sys_vm_oc_compile_interrupt_count();
```
huangminghuang: The regression test can pass without exercising the regression path: unittests/sysvmoc_interrupt_tests.cpp (line 121). If post_count == pre_count, the OC interrupt never fired, so the original dedup record was never undone by transaction_context::reset(). The later is_known_unexpired_transaction() check is then satisfied by the normal init_for_input_trx() path, meaning this test can stay green even if the production fix is removed. I’d make the interrupt a prerequisite, e.g. BOOST_REQUIRE_EQUAL(post_count, pre_count + 1), before checking the dedup record.

heifner: Good catch — adopted. `BOOST_REQUIRE_MESSAGE(fired, ...)` now gates the dedup check (7948630f2c).

A bare REQUIRE on the single push would have been timing-flaky in the other direction: the interrupt only fires if oc compile completes while the action is still interpreting, and `async_compile_complete` has a 500ms floor — a fast release build can finish the fixed-size action before the interrupt is allowed to fire. The test now retries with the action size doubling per attempt (up to 3), re-deploying with a distinct trailing custom section each time (same behavior, fresh code hash) so each attempt opens a fresh compile-interrupt window.

Verified on a RelWithDebInfo+ASan build: fires on attempt 1, 8/8 passes; with the `reset()` re-record reverted, `REQUIRE(fired)` passes and the dedup check fails 3/3 — the test can no longer pass without the production fix.

huangminghuang (approved):

PR #399, opened 2026-06-12
Generate TestHarness WASM capabilities from CMake

huangminghuang:
Summary

Generate TestHarness WASM runtime capabilities from CMake instead of discovering them by parsing `nodeop --help`.

What changed

- Add a generated `TestHarness/build_capabilities.py` module with `supports_oc`, supported runtime choices, and the default runtime.
- Make `Utils` read WASM runtime and SYS VM OC support from that generated capability module.
- Update read-only transaction tests and the performance harness to use the generated default and choices instead of hard-coded `sys-vm-jit`.
- Fail CMake configuration if TestHarness cannot derive a supported default WASM runtime.

Why

The supported runtime set is already known by CMake through `SYSIO_WASM_RUNTIMES`. Generating this capability data makes the Python test harness deterministic and avoids coupling tests to CLI help text formatting.

Testing

- `git diff --check`
- Rendered `build_capabilities.py.in` through a temporary CMake project and compiled the generated Python.
- Imported `testUtils.py` with a temporary generated `build_capabilities.py` and verified `nodeopSupportsSysVmOc()`, `supportedWasmRuntimes()`, and `defaultWasmRuntime()`.


heifner (commented):
Reviewed this by exercising the generation mechanism directly: rendered `build_capabilities.py.in` through the new CMakeLists logic for three configs (x86_64+OC, sys-vm-only, empty runtime list) and imported the result through `testUtils.py` as a package. x86_64+OC renders `supports_oc: True, runtimes: ["sys-vm-jit", "sys-vm"], default_runtime: "sys-vm-jit"` — bit-for-bit the old hard-coded behavior, so CI on x86_64 is unchanged. The sys-vm-only config degrades correctly (this actually fixes non-x86 builds, where the read-only tests previously forced a `sys-vm-jit` that isn't compiled in), and the empty list hits the FATAL_ERROR as designed. The returned runtimes list is a copy (caller mutation doesn't leak), and the missing-module path raises the intended RuntimeError. Also confirmed the import-time hard failure adds no new constraint: package imports from the source tree were already impossible since `__init__.py` imports the generated `core_symbol`.

A few things worth addressing:

**Dead constants.** The three new constants `SysVmRuntime` / `SysVmJitRuntime` / `SysVmOcRuntime` (`tests/TestHarness/testUtils.py:94-96`) aren't referenced anywhere. And `SysVmOcEnableOption` (`testUtils.py:90`) lost its only consumer when the help-parsing was removed — nothing matches option strings anymore. Suggest either wiring the runtime-name constants into use or dropping all four; orphaned constants will send the next reader hunting for option-string matching that no longer exists.

**`defaultReadOnlyWasmRuntime()` is a pure alias** of `defaultWasmRuntime()` (`testUtils.py:132-135`). If it's a deliberate seam for the read-only tests to diverge later, the docstring should say so; otherwise the two call sites could just use `defaultWasmRuntime()` directly — two names for one value invites accidental divergence.

**Preference order is load-bearing but uncommented.** In `tests/TestHarness/CMakeLists.txt:3`, `foreach(runtime IN ITEMS sys-vm-jit sys-vm)` both filters and defines preference order — the first surviving entry becomes the harness default. Worth a one-line comment so nobody alphabetizes it and silently flips the default to `sys-vm`.

Minor:
- `tests/PerformanceHarness/performance_test_basic.py:755` — the `--wasm-runtime` help prose still hard-codes `("sys-vm-jit", "sys-vm")` while `choices`/`default` are now dynamic; on a sys-vm-only build the help advertises a value argparse rejects. Could interpolate `Utils.supportedWasmRuntimes()` into the help text.
- `tests/TestHarness/CMakeLists.txt:21-24` — `foreach`/`set`/`break` to take the first element; `list(GET SYSIO_TEST_WASM_RUNTIMES 0 ...)` is the usual idiom. The FATAL_ERROR message could also include the actual `SYSIO_WASM_RUNTIMES` contents to speed diagnosis on exotic platforms.
- `configuredWasmCapabilities()` (`testUtils.py:27-29`) is a public module-level function whose only consumers are `Utils` statics — could be private to `Utils` or inlined at the three call sites.

heifner (approved):

PR #401, opened 2026-06-12
fix(chain): dedup S-root merkle leaves to one per transaction per (contract, root)

heifner:
## Problem

`root_txn_identification_impl::process_action_traces` records one merkle leaf per **matching action trace**, but the S-root merkle is meant to commit to the set of **transactions** that touched a matched contract. Two common patterns produce several matching traces for one transaction, all carrying the same `trx_id`:

- **`require_recipient` notifications** — every notified receiver gets its own action trace replicating `act.account`/`act.name`. A single matched token `transfer` yields 3 matching traces (receiver = contract, `from`, `to`).
- **Repeated matching actions** — a transaction carrying N matching actions yields N traces.

Each duplicate trace appended another identical leaf, so the per-contract S-root committed to a multiset: external settlement verifiers over-count transactions, and one transaction is provable at multiple leaf positions.

Found in the 2026-06-09 block-validation/trx-execution consensus review (finding #8); it was erroneously believed fixed by #391. Deterministic across nodes (producer and validator run the same code), so no chain-split risk — but the committed root is wrong for any external consumer. Wire-specific machinery; not present in Spring. Chain is pre-launch, so no produced-block compatibility concern.

## Fix

Dedup at insertion in `process_action_traces`: skip the append when the bucket's most recent id already equals `action_trace.trx_id`. All traces of a transaction are delivered in one applied-transaction signal and a transaction id is applied (with a receipt) at most once per block, so a duplicate can only ever be the most recently stored id — an O(1) back-check is exact, and merkle leaf order remains transaction application order.

## Tests (`test_s_root_extension`)

| Case | Asserts | Without fix |
|---|---|---|
| `notification_produces_single_leaf` | token transfer notifying `from`/`to` → exactly one leaf; `current_s_root == calculate_merkle({trx_id})` | **fails** (3 identical leaves) |
| `repeated_matching_actions_produce_single_leaf` | two matching `newaccount` actions in one trx → one leaf | **fails** (2 identical leaves) |
| `distinct_transactions_produce_one_leaf_each` | two separate matched trxs in one block → one leaf each, application order | passes (guards against over-dedup) |

Fail-without/pass-with verified locally; `forked_tests_savanna`, `snapshot_part1/2_tests`, `block_tests`, and the full `test_s_root_extension` suite are green with the fix.

### DIFF
--- a/libraries/chain/root_txn_identification.cpp
+++ b/libraries/chain/root_txn_identification.cpp
@@ -110,9 +110,21 @@ namespace sysio { namespace chain {
             if (root_contract_match.is_contract_match(contract) &&
                 root_contract_match.is_action_match(action)) {
                // We have a match, so we need to store the transaction id
-               // in the storage for this contract
+               // in the storage for this contract.
+               //
+               // A single transaction can carry several traces that match the same
+               // (contract, root): one per matching action plus one per
+               // require_recipient notification, which replicates act.account/act.name
+               // for every notified receiver. The S-root merkle commits to the set of
+               // transactions that touched the contract, so each transaction id must
+               // contribute exactly one leaf per (contract, root). All traces of a
+               // transaction are delivered in a single call and a transaction id is
+               // applied (with a receipt) at most once per block, so a duplicate can
+               // only ever be the most recently stored id.
                auto& contract_storage = this->storage[std::make_pair(contract, root_contract_match.root_name)];
-               contract_storage.push_back(action_trace.trx_id);
+               if (contract_storage.empty() || contract_storage.back() != action_trace.trx_id) {
+                  contract_storage.push_back(action_trace.trx_id);
+               }
                break;
             }
          }
--- a/unittests/test_s_root_extension.cpp
+++ b/unittests/test_s_root_extension.cpp
@@ -232,6 +232,74 @@ BOOST_AUTO_TEST_CASE(distinct_transactions_produce_one_leaf_each) {
    } FC_LOG_AND_RETHROW()
 }
 
+// A transaction that touches TWO watched (contract, root) pairs must contribute
+// one leaf to EACH pair's merkle — each contract's S-root commits to the set of
+// transactions that touched that contract. This pins why the dedup is a
+// per-bucket check rather than a per-transaction early-out: stopping at the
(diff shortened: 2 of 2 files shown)


libraries/chain/root_txn_identification.cpp
```diff
@@ -110,9 +110,21 @@ namespace sysio { namespace chain {
...
+               // require_recipient notification, which replicates act.account/act.name
+               // for every notified receiver. The S-root merkle commits to the set of
+               // transactions that touched the contract, so each transaction id must
+               // contribute exactly one leaf per (contract, root). All traces of a
+               // transaction are delivered in a single call and a transaction id is
+               // applied (with a receipt) at most once per block, so a duplicate can
+               // only ever be the most recently stored id.
                auto& contract_storage = this->storage[std::make_pair(contract, root_contract_match.root_name)];
-               contract_storage.push_back(action_trace.trx_id);
+               if (contract_storage.empty() || contract_storage.back() != action_trace.trx_id) {
```
brianjohnson5972, line 125: This works, but couldn't we skip this and just put a return in `root_txn_identification_impl::process_action_traces`  by replacing the `break` with `return`?

brianjohnson5972 (approved):
Approved, but read my comment.

libraries/chain/root_txn_identification.cpp
heifner: `return` isn't equivalent when a transaction touches more than one watched (contract, root) pair — it stops processing the transaction's remaining traces, so the trx id is only recorded under the first matched pair and disappears from every other matched contract's root. The production config in chain_plugin wires up two rules (`*utl`/batchw+snoop → "s", `*sysio`/regproducer+unregprod → "op"), and suffix matching means even two different `*utl` contracts in one transaction land in different buckets — in both cases the second contract's S-root would no longer commit to a transaction that genuinely touched it.

Added `multi_contract_transaction_records_in_each_root` (6e1cd252ca) to pin this: with `break` + the dedup check it sees two s_root_extensions; with the `return` variant it sees one (verified locally — the existing single-pair tests all still pass with `return`, which is what makes it look safe). Per-bucket dedup is the most we can skip safely.

brianjohnson5972 (approved):

PR #402, opened 2026-06-12
chain: uint128/int128 + checksum256 support in the BE key codec; fix batch_operator byoutepoch packing

heifner:
## Problem

PR #386 widened the msgch `byoutepoch` composite key to **uint128** contract-side (`opp_keys.hpp`: chain_code bits 32-79, epoch bits 0-31), but the host side was never updated:

1. `be_key_codec::encode_field` / `decode_field` (`database_utils.hpp`) stop at 64-bit types — any `get_table_rows` bound or `next_key` on a uint128 KV index asserts **`Unsupported BE key type: uint128`**.
2. `batch_operator_plugin::has_delivered_envelope` still packs the pre-#386 lossy 64-bit `(chain_code << 32) | epoch` key.

Net effect on master — correctness and noise, **not** flow-fatal: `read_table_rows` catches the assert, logs `batch_operator: table read threw sysio.msgch::envelopes -- Assert Exception (10) false: Unsupported BE key type: uint128`, and returns an empty result. `has_delivered_envelope` therefore always answers "not delivered," so the dedup short-circuit is permanently non-functional — every relay poll emits the error line and re-attempts delivery, leaning on the contract's duplicate/late-delivery tolerance. Flows still pass without this fix for exactly that reason. What's actually broken:

1. The delivered-already dedup never works (error spam + redundant `deliver` traffic every poll).
2. Pre-#386, the same query ran with the lossy 64-bit packing — silently capable of cross-wiring outposts (the truncation #386 fixed contract-side, still present host-side until this PR).
3. Any JSON bound / `find` / `next_key` against a uint128 (or checksum256) kv index is broken RPC surface.

(Found while flow-verifying wire-tools-ts #7: the flow-run heartbeat treats `table read threw` plugin errors as a first-occurrence stop signal per the cluster-probing runbook, which is what halted those runs — the circulation itself survives this particular error.)

## Fix

- `database_utils.hpp`: add `uint128` / `int128` branches to `encode_field` and `decode_field` — 16-byte big-endian, high quadword first (BE order == numeric order for the unsigned key); `int128` uses the same high-bit sign bias as the existing `int64` handling; variant round-trip via `fc::uint128` (`to_variant` decimal string / `from_variant` accepting native, uint64, or decimal/hex string).
- `batch_operator_plugin.cpp`: pack the `byoutepoch` bound per `opp_keys.hpp` (uint128) and serialize it into the `find` JSON as a **decimal string**, so the round-trip stays lossless past 2^64.

## checksum256 (second commit)

The kv-key-type census surfaced the identical hole for **checksum256**: four live secondary indexes declare it (`authex links/bypubkey`, `msgch messages/bymsgid`, `roa nodeownerreg/bytrxid`, `system finkeys/byfinkey`), and any bound/`find`/`next_key` against them asserted `Unsupported BE key type: checksum256` — un-hit so far only because nothing host-side bound-queries them yet. Encoded as the raw 32 digest bytes in display order (fixed_bytes packs words big-endian, so CDT `to_key` emits the canonical sequence and memcmp order matches checksum256 ordering); JSON spelling via `fc::sha256`'s canonical variant conversions.

`uint256`/`int256` are deliberately not added: no contract can declare one (no native 256-bit integer in the CDT toolchain; `fc::uint256` also lacks variant conversions today) — checksum256 is the 256-bit key type in actual use. The full census of declared kv secondary key types is now covered: uint64 ×36, uint128 ×8, checksum256 ×4.

## Tests

`get_table_tests` gains four cases on the existing fixtures:
- **(sec-3b)** uint128 `find` bound on `numobjs/bysec2` — the exact query shape `batch_operator_plugin` uses; asserted before this fix.
- **(sec-3c)** uint128 pagination — `next_key` produced by `decode_field` fed back through `encode_field` as `lower_bound`, covering the full decode/encode round-trip and BE ordering.
- **(sec-4b)** checksum256 `find` bound on `hashobjs/bysec1` (bound computed as `sha256("secondinput")`).
- **(sec-4c)** checksum256 pagination round-tripping `next_key` through decode/encode.

## Test sweep

- `plugin_test` — green (`get_table_tests` incl. new cases)
- `contracts_unit_test -- --sys-vm` — green
- `test_fc` — green
- `unit_test -- --sys-vm` — 4 failures, **all pre-existing on stock master** (control-verified with this change stashed; the codec is referenced nowhere in `unittests/`): `api_part2_tests/crypto_tests` ("public key does not match") and `savanna_misc_tests/verify_block_compatibitity` (reference-id mismatch — the documented regenerate-reference-data case after #386 recompiled `sysio.msgch.wasm`).

## End-to-end verification

All 9 wire-tools-ts e2e flows pass with this fix applied (wire-tools-ts #7 verification sweep), with zero `table read threw` signatures in the cluster logs — the relay dedup short-circuit works and no redundant deliveries occur. Flows also pass **without** this fix (see Problem section — the error is caught and degrades to redundant delivery), so this is a correctness/noise fix for the relay and an unbreaking of the uint128/checksum256 RPC bound surface, not a circulation blocker.

### DIFF
--- a/libraries/chain/include/sysio/chain/database_utils.hpp
+++ b/libraries/chain/include/sysio/chain/database_utils.hpp
@@ -164,6 +164,13 @@ struct reader {
    uint32_t read_be32() { return read_be<uint32_t>(); }
    uint64_t read_be64() { return read_be<uint64_t>(); }
 
+   // Copy `n` raw bytes (already order-preserving, e.g. a checksum digest).
+   void read_bytes(char* out, size_t n) {
+      FC_ASSERT(remaining() >= n, "BE key underflow reading {} raw bytes", n);
+      std::memcpy(out, pos, n);
+      pos += n;
+   }
+
    // NUL-escape decoding: 0x00,0x01 = literal NUL byte, 0x00,0x00 = end of string.
    std::string read_nul_escaped_string() {
       std::string s;
@@ -201,6 +208,9 @@ struct writer {
    std::vector<char> buf;
 
    void write_u8(uint8_t v) { buf.push_back(static_cast<char>(v)); }
+
+   // Append `n` raw bytes verbatim (already order-preserving).
+   void write_bytes(const char* p, size_t n) { buf.insert(buf.end(), p, p + n); }
 
    void write_be16(uint16_t v) { write_be(v); }
 
@@ -257,6 +267,17 @@ inline fc::variant decode_field(reader& r, const std::string& type) {
       fc::to_variant(static_cast<fc::int128>(fc::to_uint128(hi, lo)), v);
       return v;
    }
+   if (type == "checksum256") {
+      // Raw 32 digest bytes in display order: fixed_bytes packs its words
+      // big-endian, so CDT's to_key emits the canonical byte sequence and
+      // memcmp order matches checksum256 ordering. Canonical hex spelling
+      // via fc::sha256's variant conversion.
+      fc::sha256 h;
+      r.read_bytes(h.data(), h.data_size());
+      fc::variant v;
+      fc::to_variant(h, v);
+      return v;
+   }
    if (type == "name")          return fc::variant(name(r.read_be64()).to_string());
    if (type == "bool")          return fc::variant(r.read_u8() != 0);
    if (type == "string")        return fc::variant(r.read_nul_escaped_string());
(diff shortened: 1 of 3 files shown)


huangminghuang (approved):

jglanz (approved):
G2G - just wait for CICD tests to pass

PR #405, opened 2026-06-12
chain: ABI-aware BE key codec — struct key expansion (slug_name), float128, fixed_bytes<32>

heifner:
Stacked on #402 (retarget to master when it merges).

## Problem

CDT's `to_key` supports arbitrary key types — typedefs, reflected structs, all arithmetic types — but the host-side bound codec only handled a flat allowlist of builtin leaves. Census of every declared kv key type in-tree after #402:

| key type | declarers | status before this PR |
|---|---|---|
| `slug_name` (ABI struct `{ value: uint64 }`) | **primary key of the v6 registry tables** — `chains:chains`, `tokens:tokens`, `tokens:chaintokens`, `reserv:reserves` (7 key slots) | any JSON bound/`find`, and `next_key` on any paginated read, asserted `Unsupported BE key type: slug_name` |
| `float128` (long double) | test contracts (`get_table_test numobjs/bysec4`, `snapshot_test data/byff`) | no leaf branch (the float128 helpers live in the same header) |
| `fixed_bytes<32>` | `snapshot_test data/byiiii` | abigen emits the raw C++ template spelling for a checksum256-equivalent key |

## Fix

`key_shape` — the per-field encode/decode plan, resolved **once per request** from the `abi_def`:

- typedefs chased to a fixpoint (bounded by the typedef count, so alias cycles can't loop)
- struct key types expanded recursively into declaration-order fields, matching `to_key`'s reflected walk (based structs rejected — they have no defined field order)
- template spellings canonicalized (`fixed_bytes<32>` → `checksum256`)

`encode_key`/`decode_key` now take shapes; the flat `(key_names, key_types)` overloads are removed. `chain_plugin` builds shapes at request setup, `std::optional`-tolerant: hex bounds and hex key output keep working even for a hypothetical unrepresentable key type, while JSON bounds get a clear `contract_table_query_exception` instead of a deep assert.

The `float128` leaf uses the same IEEE sort-order bias as the existing float32/64 branches and byte-matches CDT `kv_multi_index::encode_secondary(long double)`; the JSON spelling is fc's canonical `"0x"` + 16 LE hex bytes.

`uint256`/`int256` remain deliberately unsupported — no producer exists (no native 256-bit integer in the CDT toolchain; `fc::uint256` lacks variant conversions).

## Query syntax

A JSON bound mirrors the key's ABI shape — the existing flat convention `{"primary_key":0}`, except a struct-typed key field becomes a nested object instead of a scalar. The decoded `key` and the `next_key` cursor come back in that same shape, so `next_key` feeds straight back as the next `lower_bound`.

For `sysio.chains/chains` (`key_names: ["code"]`, `key_types: ["slug_name"]`, `slug_name` = `{ value: uint64 }`):

```json
POST /v1/chain/get_table_rows
{
  "json": true,
  "code": "sysio.chains",
  "scope": "sysio.chains",
  "table": "chains",
  "lower_bound": "{\"code\":{\"value\":42}}",
  "upper_bound": "{\"code\":{\"value\":99}}",
  "limit": 10
}
```

Response rows and the pagination cursor carry the same nested shape:

```json
{
  "rows": [ { "key": { "code": { "value": 42 } }, "value": { ... } } ],
  "more": true,
  "next_key": "{\"code\":{\"value\":43}}"
}
```

- **Nested structs** chain the same way, encoded in declaration order: a key field `k` of type `pair_key { code: slug_name, idx: uint32 }` → `{"k":{"code":{"value":5},"idx":9}}`.
- **Secondary-index** bounds are keyed by the index name (single field), e.g. `{"bykind":1}`; a struct-typed secondary nests identically — `{"<index_name>":{...}}`.
- `json:false` (hex bounds) and hex key output are unaffected and still work for any key type, representable or not.

## Tests

- **`unittests/be_key_codec_tests.cpp`** (new, 7 cases): slug_name shape round-trip + byte ordering, two-hop typedef chain, nested struct key, `fixed_bytes<32>` ≡ `checksum256`, float128 ordering across the sign boundary + round-trip, and the rejection paths (unknown type, based struct, missing bound field, nesting depth guard). Key comparisons use `string_view` semantics to match the chain's unsigned byte order.
- **`get_table_tests` (sec-9)**: float128 e2e pagination on `numobjs/bysec4` — `next_key` produced by the codec round-trips back through `lower_bound` on the live RPC path.
- **`get_table_tests` (sec-10)**: live struct-key (`slug_name`) coverage on the `get_table_rows` path. A struct-keyed `kv::table` (`structobjs`) was added to the `get_table_test` contract, keyed by a reflected `slug_name { value: uint64 }` — the regenerated ABI emits `key_types: ["slug_name"]`, mirroring `sysio.chains chains`. The case drives (a) a JSON bound of the documented `{"code":{"value":N}}` form (`encode_key`) and (b) a `next_key` pagination cursor that round-trips the nested struct shape (`decode_key → next_key → encode_key`). Fail-without verified: against the pre-struct codec the case asserts `Unsupported BE key type: slug_name`. The contract change is additive — its other consumers (`table_reader_tests`, `state_history` contract-delta cases) still pass.

Note: the e2e `slug_name` coverage uses a purpose-built test-contract table rather than the real `sysio.chains` (the plugin_test fixture doesn't bootstrap system contracts), but the key shape is identical (`key_types: ["slug_name"]`); the unit tests additionally pin the slug_name encoding byte-for-byte.


huangminghuang (approved):
A few small improvements, but nothing I’d block on:

- `leaf_key_spellings` could derive its array size instead of hard-coding `20`, e.g. `std::to_array<leaf_key_spelling>({...})` if available. That would remove one tiny sync point when adding a leaf spelling.

- The fallback comments in `chain_plugin.cpp` say “JSON keys fall back to hex”, but JSON bounds are rejected while JSON output falls back to hex. The code is correct; the wording could be a touch more precise to avoid future confusion.

- `resolve_key_type()` stops after bounded hops but doesn’t explicitly say “typedef cycle” when it lands on an unresolved alias. The eventual `Unsupported BE key type: ...` is acceptable, but a cycle-specific assertion would make malformed ABI diagnostics sharper.

- There’s room to add a focused test for scoped primary struct keys if those can exist in real ABIs. Current struct coverage is unscoped primary (`structobjs`) and scalar scoped tables. Not urgent unless registry-like scoped struct keys are expected.

PR #408, opened 2026-06-13
test: deflake snapshot_attest_test & snapshot_api_test (regproducer/setrank race)

heifner:
## Problem

`snapshot_attest_test` failed on the **ubuntu24** CI runner (passed on asan/ubsan/asserton/gcc) with:

```
Failed to set rank for defproducerb        # on-chain assert: "producer not found"
```

Root cause: `setrank` reads the on-chain `producers` table and asserts `"producer not found"` if a producer's `regproducer` has not yet been applied. Both tests followed `regproducer` with only `waitForHeadToAdvance()` (one block), which does **not** guarantee those specific transactions were applied — `pushMessage` only confirms speculative execution, and with multiple producing nodes a transaction pushed to node0 can be forwarded into a peer's block, racing the dependent `setrank`. The race is timing-sensitive, hence the single-runner flake.

`snapshot_api_test` has the identical pattern in its setup (latent), so it is fixed here too.

## Fix

Wait for the prerequisite transactions to appear in a block (via the existing `waitForTransactionsInBlock` / `waitForTransactionInBlock` harness helpers) before the dependent action/read:

- **`snapshot_attest_test`** — `regproducer` transactions before `setrank`; `regsnapprov` transactions before the `snapprovs` exact-count assertion.
- **`snapshot_api_test`** — `regproducer` transaction before `setrank`; `regsnapprov`/`setsnpcfg` before the snapshot-attestation flow that depends on them.

Test-harness (Python) only — no production code changes.

## Verification

Both tests run end-to-end locally against a freshly built `nodeop`:

- `snapshot_attest_test`: the previously-racing `Set rank for defproducerb` now passes; **Tests 1–5 PASS**.
- `snapshot_api_test`: **Tests 1–10 PASS**.

### DIFF
--- a/tests/snapshot_attest_test.py
+++ b/tests/snapshot_attest_test.py
@@ -152,16 +152,19 @@ try:
         json.dumps({"min_providers": 1, "threshold_pct": 50}),
         "--permission sysio@active")
     assert success, f"Failed to set snapshot config: {trans}"
-
-    assert node0.waitForHeadToAdvance(), "Head did not advance after config"
+    setCfgTransId = node0.getTransId(trans)
 
     # ---------------------------------------------------------------
     # Verify provider registration via table query
     # ---------------------------------------------------------------
-    # Ensure both regsnapprov registrations are in a block before reading the
-    # table, so the exact-count assertion below cannot race the registrations.
-    assert node0.waitForTransactionsInBlock(regSnapProvTransIds, timeout=60), \
-        "regsnapprov transactions did not make it into a block"
+    # Ensure both provider registrations and the attestation config are applied
+    # in a block before proceeding. The exact-count assertion below depends on
+    # the registrations, and Test 1's single votesnaphash depends on setsnpcfg
+    # making quorum == 1 — if the config is forwarded into a peer's block and not
+    # yet applied, the vote runs against the default config (2 providers @ 67%,
+    # quorum 2), creates no snaprecords entry, and the test fails.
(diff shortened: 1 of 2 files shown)


tests/snapshot_attest_test.py
huangminghuang, line 151: This still follows setsnpcfg with only waitForHeadToAdvance(). The first votesnaphash later relies on that config making quorum equal 1, but if the config transaction was forwarded into a peer’s block and is not yet applied on node0, the vote runs with the default config: 2 providers at 67%, quorum 2. Then the single vote will not create snaprecords, and the test can still fail. This is the same class of race the PR fixes elsewhere; capture setCfgTransId = node0.getTransId(trans) and include it in the exact waitForTransactionsInBlock(...) before the snapshot/attestation flow.

heifner: Good catch — fixed in 20410d8d03.

Captured `setCfgTransId = node0.getTransId(trans)` after `setsnpcfg` and added it to the existing in-block wait — it's now `waitForTransactionsInBlock(regSnapProvTransIds + [setCfgTransId], ...)` before the snapshot/attestation flow — and dropped the now-redundant `waitForHeadToAdvance()` after the config. So Test 1's single `votesnaphash` can no longer run against a stale default config (2 providers @ 67%, quorum 2). This matches the pattern already used in `snapshot_api_test.py`.

Verified end-to-end locally against a freshly built nodeop: Tests 1–5 pass.

huangminghuang (approved):

PR #410, opened 2026-06-16
opp: DEX bootstrap config — input-shape proto, example configs, validation test

heifner:
## Summary

Launch-day chains, tokens, reserves, and swap settings become a single human-authored JSON file validated against a protobuf schema, replacing the dataset hard-coded in the cluster tooling. Three artifacts:

- **`libraries/opp/proto/sysio/opp/bootstrap/bootstrap.proto`** — message `DexConfig` (`ChainSpec` / `TokenSpec` / `ReserveSpec` / `UwritConfig` + a `t5_reserve_allocation` earmark). Auto-discovered by the proto GLOB; both the host `.pb.h` and CDT `.pb.hpp` codegen build clean with no CMake changes.
- **`etc/config/dex/dex-config.launch.example.json`** — mainnet strawman (3 chains / 5 tokens / 4 reserves, placeholder economics, mainnet token addresses marked VERIFY-BEFORE-LAUNCH) and **`dex-config.dev.json`** — the full dev-cluster dataset (3 / 9 / 8) proving the schema carries it 1:1.
- **`docs/dex-bootstrap-config.md`** — engineering doc (schema, action mapping, T5 earmark mechanics, validation).

### Authoring conventions (vs. the registry carriers in `types.proto`)

The spec messages deliberately differ from `Chain`/`Token`/`ChainToken`/`Reserve` (wire carriers with packed-uint64 codes and `bytes` addresses): codes are **strings** (slug-packed by the tool), addresses are **chain-native display strings** (0x-hex / base58, eyeball-verifiable against block explorers), amounts are **uint64 subunits** (quoted in JSON, since launch-scale values exceed 2^53), enums by full name. The `ChainToken` binding is folded into `TokenSpec` (a token binds to exactly one chain); there is no `is_depot` field (the depot is the single `CHAIN_KIND_WIRE` chain).

### T5 reserve earmark (config + arithmetic, no contract change)

`regreserve` drains each reserve's WIRE side from the `sysio` emissions treasury at registration. `t5_reserve_allocation` sits **outside** `emission_config.t5_distributable`: at genesis `t5_distributable = T5_allotment − t5_reserve_allocation`, so the `sysio.epoch` emissions readiness gate never sees a short treasury balance after the drain. Invariant: `Σ reserves[].initial_wire_amount ≤ t5_reserve_allocation`.

## Why stacked on #407

The schema matches the reserve-and-swap-beta surface from #407: `regreserve(..., is_private, owner)` and the ms-based `uwrit::setconfig(fee_bps, collateral_lock_duration_ms, ...)`. Built on #407's branch so the proto/config reference the final signatures. **Merge #407 first, then retarget this onto master.**

## Tests

New Boost suite `dex_bootstrap_config` in `test_opp` (`libraries/opp/test/test_dex_bootstrap_config.cpp`), 4 cases:

- `launch_example_parses_and_validates` / `dev_config_parses_and_validates` — both example configs strict-parse and satisfy invariants V1–V9.
- `strict_parse_rejects_unknown_field` — a typo'd JSON key fails the strict parse (not silently dropped).
- `validator_rejects_mutations` — 7 single-field mutations of the valid dev config, each tripping one invariant (slug length, two-depots, two-natives-on-a-chain, weight bound, earmark sum, private-without-owner, fee splits ≠ 100).

The validator enforces slug validity, unique codes + exactly-one-depot, token bindings + precision + address well-formedness (EVM `0x`+40 hex, SVM base58→32 bytes), one-native-per-chain, reserve uniqueness/binding/weight/amounts, `Σ wire ≤ earmark`, owner rules, and the uwrit fee config.

Result: `test_opp` 5/5 green. Purely additive (new proto message + new test) — no contract or chain-serialization change, so no reference-data regen.

## Follow-ups (after #407 merges)

- Wire the bootstrap tool to consume `DexConfig` (the cluster phases become a loop over the parsed config), injecting per-deployment addresses from deployment artifacts into the dev config.
- Regenerate the TS/Solidity/Solana model bundles so the tool imports the generated `DexConfig` type.
- Fill in launch economics + verified mainnet addresses; freeze `dex-config.mainnet.json`.


jglanz (approved):
Approved, with the caveat that we will have to tweak this before launch I believe

FYI - On the open questions, we wont know the definitive bootstrap reserve values (LIQ (erc20/spl) and NATIVE (ETH/SOL) until we stop the existing outpost

PR #414, opened 2026-06-17
opp/dex: weighted Bancor pricing, WIRE-leg swap fee, exact-fill semantics

heifner:
Stacked on #410. Wires up the DEX swap economics that #407 declared but left inert, and fixes the precision/semantics gaps surfaced in pre-PR review.

## F1 — weighted Bancor (honor `connector_weight_bps`)
New deterministic, integer-only weighted constant-product kernel (`contracts/sysio.opp.common/.../amm_math.hpp`):
- exact constant-product fast path for the 50/50 case (bit-identical to the old `cp_output`),
- Q60 fixed-point `log2`/`exp2` for other weights (validated to sub-subunit accuracy vs `__float128`/Decimal references).

Replaces the two duplicated `cp_output` copies in `sysio.reserv` and `sysio.uwrit`, so quotes and books ride one curve. `connector_weight_bps` is the WIRE-side weight (token side = remainder).

## F2 — swap fee out of the WIRE leg
`sysio.uwrit::fee_bps` is charged on the WIRE leg of every swap (so the ETH/SOL the recipient can receive is reduced). `sysio.reserv` routes the collected fee **50/50** to a new on-chain `rewards_bucket` (WIRE kept in custody for a later distribution action) and the `sysio` emissions treasury. Custody invariant preserved. Removes the unused whole-percent `fee_split_*_pct` config in favor of the fixed bps split.

## F3 — exact-fill-or-revert
`target_amount` is the exact amount delivered (never less); if the live **post-fee** quote drifts outside `target_tolerance_bps` the swap reverts (`SwapRevert`). Proto doc corrected (was "minimum"); the variance check uses the post-fee quote at ingestion, race resolution, and drain.

## Also
- read-only `swapquote` gains WIRE-endpoint parity with the internal quote;
- proto comments fixed (`target_amount` canonical 9-decimal frame; `connector_weight_bps` WIRE-side weight);
- bootstrap proto / example configs / validator updated for the removed `fee_split` fields.

## Tests
- new `amm_math_tests` (kernel vs high-precision references, invariants, 50/50 exactness),
- new fee-routing settlement test (rewards bucket + emissions transfer + custody),
- updated `sysio.reserv` / `sysio.uwrit` suites.

Full `contracts_unit_test` green (380 cases); `test_opp` green (4 cases).


jglanz (approved):
@heifner this is really good work - awesome to see the alignment.  With the `e2e-tests` run from `wire-platform-build-system` complete, we are now just about feature complete give or take some chores, cleanup and bug fixes

GREAT WORK

PR #415, opened 2026-06-17
Pre-launch security hardening: non-contract fixes

heifner:
Non-contract fixes from the 2026-06-17 pre-launch security/bug audit (find → adversarial-verify). The contract-side findings (msgch::queueout authorization, the OPP "never-throw in dispatch" cluster, uwrit chklocks settlement, etc.) are handled separately.

Commits, highest severity first:

**🔴 net_plugin: reject under-length trx/vote frames; bound message_buffer advance**
The early-dedup path computed `advance_read_ptr(message_length - header_bytes)` from the peer-supplied frame length (both `uint32_t`). A peer that declares `message_length < 33` while pipelining ≥33 bytes makes `header_bytes > message_length` and underflows the subtraction to ~4.29e9, over-popping the `message_buffer` deque (`free()`/`pop_front()` on an emptied deque is UB) — a remote, **pre-authentication** memory-corruption primitive on a default-config node (`allowed-connection=any`). Guards both paths, adds a defense-in-depth bound in `advance_read_ptr`, plus a regression test.

**🟠 libfc: overflow-safe bounds checks in the Ethereum ABI decoder**
`offset + N <= data_size` checks wrap `size_t` for offsets/lengths decoded from untrusted ABI words on the inbound OPP path, so a value near 2^64 passes the check and `data + offset` is dereferenced at a wild address (OOB read; the SIGSEGV is not caught there). `decode_static_value` never even received `data_size`. Adds overflow-safe helpers, threads `data_size` through, bounds array element counts, plus a regression test feeding overflowing offset/length words.

**🟡 chain: pin snapshot row_count to hashed section data on load**
A section's `row_count` lives in the section index, which is outside the integrity/root hash. A tampered (decreased) `row_count` silently truncates loaded rows while still matching the on-chain attested hash, loading a node with consensus state that diverges from the canonical chain. Requires exact section consumption after the last row + an empty-section invariant.

**🟢 chain: 64-bit length math in BLS host functions**
`n*96` / `n*192` / `n*32` guards wrapped in `uint32_t`; a crafted `n` could match an undersized span and the `i*96` loop then read past the buffer. Computed in `size_t`.

**🟢 libfc: fix big-endian sha3 OOB write + uninitialized index**
The big-endian branch (constexpr-discarded on Wire's little-endian targets) used `sizeof(words)` as an element count and an uninitialized loop induction variable.

**🟢 underwriter_plugin: harden ETH/SOL source-deposit verification**
ETH: `std::stoul`/`std::stoull` on a malformed RPC field throw `std::` exceptions (not `fc::exception`) that escaped and aborted the whole scan cycle; now caught + blockNumber prefix guarded. SOL: matched the SwapDeposit marker anywhere in `meta.logMessages` without program attribution (a forged marker from any program in a tx that merely references opp-outpost passed); now tracks the `Program invoke/success|failed` bracket stack and trusts the marker only when opp-outpost is top-of-stack.

**🟢 chain: consistent weak_final boundary in QC vote aggregation**
`add_weak_vote`'s `weak_achieved` arm used `>=` vs the strict `>` of the unrestricted/restricted arm, declaring `weak_final` one weight-unit early (`weak_sum == sum - threshold` still allows a strong QC). Local aggregation only; QC verification is independent and unchanged, and the Savanna proof format/verification are unaffected — at the exact boundary it only changes which valid QC variant (strong vs weak) a producer forms, in the safe direction. Confirmed present **verbatim in upstream Spring** — see Upstream below. Covered by a new boundary regression test.

## Testing
Full pre-PR sweep — `unit_test` (`--sys-vm`), `plugin_test`, `contracts_unit_test` (`--sys-vm`), `test_fc` — all green **except three failures that pre-exist on `master`** and are unrelated to this diff (verified byte-identical on a clean-`master` baseline): `deep_mind_tests` and `savanna_misc_tests/verify_block_compatibitity` (stale reference data — need regeneration after the latest contract recompiles) and `api_part2_tests/crypto_tests`. None touch the files in this PR. New tests: `message_buffer` advance underflow; Ethereum ABI malicious offset/length decode; QC `weak_final` boundary (`finality_misc_tests/qc_state_transitions`) — fail-without-fix verified (the new asserts fail with `>=`, pass with `>`).

## Upstream (Spring)
The QC `weak_final` off-by-one (last commit) is **not a Wire regression** — it exists verbatim in upstream `AntelopeIO/spring`, and Spring's own `qc_state_transitions` unit test has the same `max_weak_sum_before_weak_final == 1` blind spot that kept it from being caught. The one-line fix and the new boundary regression test here apply to Spring unchanged and should be filed/upstreamed as a Spring issue.

### DIFF
--- a/libraries/chain/snapshot.cpp
+++ b/libraries/chain/snapshot.cpp
@@ -616,6 +616,11 @@ void threaded_snapshot_reader::set_section(const string& section_name) {
       if(entry.name == section_name) {
          cur_row = 0;
          num_rows = entry.row_count;
+         // An empty section carries no data bytes (writer invariant). Enforce it so a tampered
+         // row_count of 0 cannot hide a non-empty, hash-pinned data range from ever being read.
+         SYS_ASSERT(num_rows != 0 || entry.data_size == 0, snapshot_exception,
+                    "Snapshot section '{}' has row_count 0 but {} bytes of data; file may be tampered",
+                    section_name, entry.data_size);
          ds = fc::datastream<const char*>(mapped_snap_addr + entry.data_offset, entry.data_size);
          return;
       }
@@ -626,7 +631,16 @@ void threaded_snapshot_reader::set_section(const string& section_name) {
 
 bool threaded_snapshot_reader::read_row(detail::abstract_snapshot_row_reader& row_reader) {
    row_reader.provide(ds);
-   return ++cur_row < num_rows;
+   if (++cur_row < num_rows)
+      return true;
+   // Last row consumed. A section's data_size is covered by the integrity/root hash but its
+   // row_count is not, so require the section data to be exactly consumed: a tampered (decreased)
+   // row_count would otherwise silently drop the trailing rows while still matching the attested
+   // hash, loading a node with state that diverges from the canonical chain.
+   SYS_ASSERT(ds.remaining() == 0, snapshot_exception,
+              "Snapshot section row_count does not match section data ({} bytes unread); "
+              "file may be tampered", ds.remaining());
+   return false;
 }
 
 bool threaded_snapshot_reader::empty ( ) {
--- a/libraries/chain/snapshot.cpp
+++ b/libraries/chain/snapshot.cpp
@@ -631,16 +631,7 @@ void threaded_snapshot_reader::set_section(const string& section_name) {
 
 bool threaded_snapshot_reader::read_row(detail::abstract_snapshot_row_reader& row_reader) {
    row_reader.provide(ds);
-   if (++cur_row < num_rows)
-      return true;
-   // Last row consumed. A section's data_size is covered by the integrity/root hash but its
-   // row_count is not, so require the section data to be exactly consumed: a tampered (decreased)
-   // row_count would otherwise silently drop the trailing rows while still matching the attested
-   // hash, loading a node with state that diverges from the canonical chain.
-   SYS_ASSERT(ds.remaining() == 0, snapshot_exception,
-              "Snapshot section row_count does not match section data ({} bytes unread); "
-              "file may be tampered", ds.remaining());
-   return false;
+   return ++cur_row < num_rows;
 }
 
 bool threaded_snapshot_reader::empty ( ) {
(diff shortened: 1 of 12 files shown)


huangminghuang (approved):

libraries/chain/snapshot.cpp
```diff
@@ -626,7 +631,16 @@ void threaded_snapshot_reader::set_section(const string& section_name) {
 
 bool threaded_snapshot_reader::read_row(detail::abstract_snapshot_row_reader& row_reader) {
```
huangminghuang, line 632: The new exact-consumption check only runs when `read_row()` is called for the declared last row. Several callers read singleton sections once and ignore the returned `bool` before `read_section()` calls `clear_section()`; for example `chain_snapshot_header`, `block_state`, and `global_property_object`. If the index row_count is tampered upward from 1 to 2 while the hashed data still contains one row, the first read succeeds, the callback returns, and `clear_section()` resets without checking `cur_row == num_rows` or `ds.remaining() == 0`. That means row_count is still not fully pinned to the hashed section data for these sections. Consider enforcing the invariant in `clear_section()`/`read_section()` as well, so every selected section must be fully consumed exactly, including singleton readers.

huangminghuang (changes requested):

libraries/chain/snapshot.cpp
heifner: Good catch — the `read_row`-only check missed exactly the singleton readers (one `read_row`, return value ignored). Moved enforcement to `clear_section()`, the universal teardown reached after the `read_section` callback for both loop readers and singleton readers. It now requires `cur_row == num_rows` **and** `ds.remaining() == 0`, so an upward-tampered `row_count` for a singleton (e.g. `cur_row=1`, `num_rows=2`) is rejected at teardown — pinning the unhashed `row_count` to the hash-covered section data for every reader.

The check runs before the madvise `seekp(0)` and is guarded by `if(num_rows)`, so `return_to_header` before any read (tester `open` / `snapshot_api` info) is unaffected — verified both call sites have `num_rows == 0` there.

Added a regression test `threaded_snapshot_requires_full_section_consumption`: writes a genuine 2-row section and then reads only one row (the singleton shape), which must be rejected at teardown. Verified it fails without the `clear_section` check and passes with it; `snapshot_part1/part2/attest` all remain green.

Fixed in 6f34b69777.

huangminghuang (approved):

PR #417, opened 2026-06-18
Launch audit: OPP/DEX system-contract security fixes

heifner:
Contract companion to #415 (non-contract fixes, merged). System-contract fixes from
the 2026-06-17 pre-launch security audit, re-evaluated against post-#414 master. Each
fix is scoped to one contract with regression tests; the affected wasm are rebuilt.

## Fixes

**`sysio.msgch` — gate `queueout` to depot system contracts (CRITICAL)**
`queueout` had no `require_auth`, so any account could inject forged attestations into
the depot-signed outbound envelope (drain ETH/SOL outpost funds, forge SLASH, activate
reserves). Now gated to `{epoch, uwrit, opreg, reserv, self}`.

**`sysio.dclaim` — keep `onreward` non-throwing on malformed input**
`onreward` is reachable from the consensus-tipping deliver→evalcons chain. A `name()`
throw on a bad `staker_wire_account`, or an `asset{}` throw on a reward ≥ 2^62, would
abort the tx and stall epoch advancement chain-wide. Now validates the name string and
bounds the amount, parking malformed rewards instead of throwing.

**`sysio.opreg` — handle TERMINATED operator in `flushwtdw`; fix dead re-validation guard**
A terminated operator's matured withdraw hit a `subtract_balance` underflow throw. Adds
a TERMINATED branch (erase without subtract) and replaces the never-firing guard with a
real `find_balance` / sufficiency check.

**`sysio.reserv` — saturating reserve credits + connector-weight cap**
Unchecked `uint64 +=` on operator-relayed external amounts could overflow reserve
balances; all credit sites now saturate. `MAX_CONNECTOR_WEIGHT_BPS` lowered 10000→9999
(cw=10000 zeroed the token-side weight → `out_given_in` returns 0, a dead reserve).

**`sysio.reserv` — reclaimable CANCELLED reserve row (#19)**
`oncrtreserve` inserts a CANCELLED row on a no-authex-link create (refund idempotency),
but the existence guard treated that dead row as immutable — permanently burning the
`(chain, token, reserve_code)` identity for its rightful owner (no path erases a row).
Now a CANCELLED row is reclaimable by a later, properly-linked creator (overwrite in
place, `bystatus` CANCELLED→PENDING); non-CANCELLED rows stay immutable; a re-relay of
the same unlinked create still avoids a second `RESERVE_CREATE_CANCELLED` refund.

**`sysio.uwrit` — non-throwing swap-remit settlement in evalcons (#5-residual)**
`emit_swap_remit` resolved the recipient / destination outpost / winner's destination
authex link with `check()` AFTER the caller's `reserv::applyswap`/`applyfromwire`
mutation, inside the synchronous evalcons → dispatch → try_select_winner chain — a throw
there aborts consensus application for the whole envelope and stalls OPP epoch
advancement chain-wide. Split into a non-throwing `try_build_swap_remit`
(`ok`/`terminal`/`disqualified`) run BEFORE any lock / CONFIRMED / reserve write, plus
`queue_swap_remit` which only ships the pre-built envelope after the books move.
`terminal` (uwreq-wide: stored-request/dst-outpost/chain-kind) → REJECT + refund;
`disqualified` (candidate-specific: missing dst-chain authex link) → skip candidate,
leave the race open. Adds non-throwing `disqualify_candidate` / `reject_and_refund`
(also reused by the variance-drift and invalid-WIRE-recipient paths).

**`sysio.opp.common` — fix `UnderwriteIntentCommit` DataStream field order**
The CDT operator dropped `outpost_id` and double-encoded `chain_code`. Currently dead
code (not instantiated into compiled output), fixed to remove the landmine.

## Tests
`contracts_unit_test`: `sysio_msgch_tests`, `sysio_dclaim_tests`, `sysio_opreg_tests`,
`sysio_reserve_tests` (incl. `oncrtreserve_cancelled_relay_does_not_double_refund` and
`oncrtreserve_cancelled_is_reclaimable_by_linked_creator`), `sysio_uwrit_tests`, and
`sysio_dispatch_tests` — all green. The #5-residual `disqualified` classification is now
covered by `sysio_dispatch_tests::swap_winner_without_dst_authex_link_is_disqualified`:
it drives a full two-leg ETH→SOLANA underwrite race to `try_select_winner` (active-key
signed UICs, opreg bond via `depositinle`, ACTIVE funded reserves via `regreserve`) where
the winner holds an EVM authex link but no SVM link, and asserts the commit is SLASHED
with the "no authex link for the destination" reason while the uwreq stays PENDING — i.e.
the disqualification is reached in `try_build_swap_remit` before any CONFIRMED/reserve
write, so nothing throws in evalcons. (The `terminal`/`reject_and_refund` branches remain
e2e-covered.)

## Artifacts
Rebuilt `sysio.{msgch,dclaim,opreg,reserv,uwrit}.wasm`. The round-2 review fixes add `UNDERWRITE_STATUS_DISQUALIFIED` to `sysio.uwrit.abi` (the only ABI change); all other ABIs are unchanged. `sysio.dclaim`/`sysio.reserv` recompiled byte-identical.
No reference-data regen needed: the deep_mind / consensus / snapshot reference tests
deploy only bios / token / test-contracts, none of which this branch changes.


contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -831,15 +826,81 @@ uwrit::commit_entry* find_or_create_commit(uwrit::uw_request_t& req, name underw
...
+         std::span{req.attestation_inbound_data.data(),
+                    req.attestation_inbound_data.size()},
+         zpp::bits::no_size{}};
+      if (in(sr) == zpp::bits::errc{}) {
+         if (auto src_outpost_opt = find_outpost_id_for_chain(req.src_chain_code)) {
+            emit_swap_revert(self, *src_outpost_opt, req.id, sr,
+                             req.src_chain_code, req.src_reserve_code, revert_reason);
+         }
+      }
+   } else if (auto user = parse_wire_name(req.depositor)) {
```
huangminghuang, line 896: This depends on `parse_wire_name`, which currently rejects every 13-character WIRE name. CDT names can be 13 chars when the final symbol fits the 4-bit final slot, so valid from-WIRE depositors/to-WIRE recipients can be treated as invalid. Please reuse/extract the dclaim-style non-throwing name validation so the full CDT name domain is accepted.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -831,15 +826,81 @@ uwrit::commit_entry* find_or_create_commit(uwrit::uw_request_t& req, name underw
...
 
+/// Mark a single candidate's commit_entry as race-disqualified — the
+/// `SLASHED` status doubles as the "lost on a hard disqualification" sentinel
+/// (vs `RELEASED` for a clean race loser) — and leave the uwreq PENDING so
+/// the race can still resolve for another valid underwriter. Non-throwing.
+void disqualify_candidate(uwrit::uwreqs_t& reqs, const uwrit::id_key& pk,
+                          name candidate, const std::string& reason) {
+   reqs.modify(same_payer, pk, [&](auto& r) {
+      auto* c = find_or_create_commit(r, candidate);
+      c->status = UnderwriteStatus::UNDERWRITE_STATUS_SLASHED;
```
huangminghuang: This helper should also be used for invalid UIC signatures. Right now the invalid-signature branch only logs and returns, leaving the candidate as `INTENT_SUBMITTED`; the same bad commit can keep the uwreq pending/noisy until another underwriter wins. Disqualifying it here would make race state converge like the missing-destination-authex-link path.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -831,15 +826,81 @@ uwrit::commit_entry* find_or_create_commit(uwrit::uw_request_t& req, name underw
    return &req.commits_by.back();
 }
 
+/// Mark a single candidate's commit_entry as race-disqualified — the
+/// `SLASHED` status doubles as the "lost on a hard disqualification" sentinel
```
huangminghuang: This overloads `UNDERWRITE_STATUS_SLASHED` to mean race-disqualified even when no economic slash happened. If ABI churn is still acceptable pre-production, a distinct status such as `UNDERWRITE_STATUS_DISQUALIFIED` would make downstream indexing and audits clearer, and avoid conflating protocol punishment with pre-settlement invalidity.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -367,124 +367,119 @@ std::vector<char> wire_name_bytes(name n) {
...
    opp::attestations::SwapRequest sr;
    {
       auto in = zpp::bits::in{
          std::span{req.attestation_inbound_data.data(),
                     req.attestation_inbound_data.size()},
          zpp::bits::no_size{}};
-      auto rc = in(sr);
-      check(rc == zpp::bits::errc{},
-            "emit_swap_remit: failed to decode stored SwapRequest");
+      if (in(sr) != zpp::bits::errc{}) return swap_remit_disp::terminal;
```
huangminghuang, line 416: Nice: this decode is now non-throwing. The same evalcons rule should probably be applied to `createuwreq()` too: its `SwapRequest` decode still uses `check(rc == ...)`, so a malformed inbound swap can abort the consensus-tipping delivery before a uwreq row even exists. Please soft-handle that path as well, by emitting a revert when possible or logging/skipping without aborting.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -1055,6 +1063,40 @@ void try_select_winner(name self, uint64_t uwreq_id, name candidate) {
       }
    }
 
+   // ── Pre-build the outbound SWAP_REMIT (dst-outpost paths only) ───────
```
huangminghuang, line 1119: [P1] Before this new “no failures past prebuild” boundary, the to-WIRE precheck above still uses unchecked `req.dst_amount + to_wire_fee`, and `req.dst_amount` comes from the cross-chain `SwapRequest.target_amount` as an unbounded `uint64_t`. A maliciously large target can wrap that sum and pass the guard, then `reserv::paywire` can still abort when it builds/sends `asset(static_cast<int64_t>(wire_out), WIRE_SYMBOL)` after the request has moved into the settlement path. Please reject/refund if `dst_amount > asset::max_amount`, use checked/saturating addition for payout + fee before any CONFIRMED/lock/reserve mutation, and add regression tests for oversized target, payout+fee overflow, and target > asset max.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Done — extracted the dclaim-style validator into a shared header `contracts/sysio.opp.common/safe_ops.hpp` (`sysio::opp::safe::is_valid_name_string`), and both `sysio.dclaim` and `sysio.uwrit::parse_wire_name` now delegate to it. `parse_wire_name` no longer caps at `size() > 12`, so the full CDT name domain — including a legitimate 13-byte name whose final symbol fits the 4-bit final slot — is accepted (it still rejects the empty address as a non-principal). Pinned by host unit tests in `contracts/tests/safe_ops_tests.cpp`: `aaaaaaaaaaaaa`/`...j` accepted, `...k`/`wirerecipient` (final `t`)/`...z` rejected.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Done — the invalid-signature branch in `try_select_winner` now calls `disqualify_candidate(...)` (reason: "invalid underwrite-intent-commit signature on one or both required legs") instead of only logging, so the race state converges exactly like the missing-destination-authex-link path, and the candidate can re-arm via a later `rcrdcommit`. Still non-throwing. New regression test `sysio_dispatch_tests/swap_candidate_with_invalid_uic_signature_is_disqualified` drives a source UIC signed by the wrong account → candidate `DISQUALIFIED`, uwreq stays `PENDING`, push succeeds.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Done — added `UNDERWRITE_STATUS_DISQUALIFIED = 11` to `types.proto` (and the host `FC_REFLECT_ENUM` in `libraries/opp/include/sysio/opp/opp.hpp`). `disqualify_candidate` and the `rcrdcommit` re-arm guard now use it. `commit_entry.status` is never set to `SLASHED` anymore — an economic slash is recorded on the lock, not the commit — so `SLASHED`/`RELEASED`/`DISQUALIFIED` are now cleanly separated for indexing/audits. `sysio.uwrit.abi` regenerated with the new value; the existing `swap_winner_without_dst_authex_link_is_disqualified` assertion updated to expect `DISQUALIFIED`.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Done — `createuwreq` now soft-handles a malformed `SwapRequest` (and, for good measure, a non-`SWAP_REQUEST` type): it logs + returns without creating a row, instead of `check()`-aborting the consensus-tipping delivery. A revert is intentionally *not* emitted on a decode failure: `emit_swap_revert` needs the decoded `actor` / `source_amount` to tell the outpost which deposit to reverse — the exact fields that failed to decode — so there is nothing to route a refund to, and no row is created so no state is corrupted. New test `createuwreq_malformed_swaprequest_does_not_abort` (garbage payload → success, no uwreq row).

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Done (P1). In the to-WIRE precheck, before any CONFIRMED/lock/reserve write:

1. **Terminal reject + refund** when `req.dst_amount > asset::max_amount` — the payout is sent as `asset(static_cast<int64_t>(dst_amount), WIRE)` by `reserv::paywire`, so an unrepresentable target can never settle for any winner or liquidity. This also covers the `2^63`-magnitude (negative-as-`int64`) case.
2. The payout+fee guard now uses saturating `sysio::opp::safe::add_sat_u64(dst_amount, to_wire_fee)` (shared with `reserv::add_capped_u64`), so it is provably overflow-free at the settlement boundary.

Regression test `swap_to_wire_oversized_target_is_rejected_not_aborted` drives a full to-WIRE race with `target = asset::max+1`, `2^63`, and `UINT64_MAX`: each ends `REJECTED` (refunded) with the push succeeding — a pre-fix `UINT64_MAX` wraps the guard and aborts paywire's `asset()`, failing the tx. Plus host unit tests for `add_sat_u64` saturation.

huangminghuang (approved):

huangminghuang (approved):

jglanz (approved):
good stuff!

PR #418, opened 2026-06-18
snapshot-attest: enforce a Byzantine quorum floor in votesnaphash

heifner:
## Problem
`votesnaphash` computes its attestation quorum purely from governance-set config:

    quorum = max(min_providers, ceil(provider_count * threshold_pct / 100))

Both knobs are operator-settable and `setsnpcfg` permits `min_providers = 1`,
`threshold_pct = 1`. Misconfigured low, a single Byzantine snapshot provider
(which may be any top-30 producer's provider) can unilaterally attest an
arbitrary `(block_id, snapshot_hash)`. Because reaching quorum also purges all
pending votes at `block_num <= attested`, that single false attestation can
also discard honest providers' in-flight votes — honest nodes that loaded the
real snapshot then see a disagreeing on-chain record and self-shut-down.

## Fix
Add a Byzantine quorum floor enforced independently of the governance config:

    bft_floor = provider_count / 3 + 1
    quorum    = max(max(min_providers, ceil(provider_count*threshold_pct/100)), bft_floor)

Under the standard `< N/3` fault assumption an attestation must now carry more
than `N/3` registered providers, so a Byzantine minority cannot on its own
attest a record — and, combined with the existing disagreement reject, cannot
win the race to quorum. The floor cannot be lowered by governance.

## Operator note (testnets / sizing)
The floor is a **no-op under the default `threshold_pct = 67`**: `ceil(0.67·N)`
is `>= floor(N/3)+1` at every provider count `N`, so a normally-configured chain
sees no change in required quorum. It only raises the bar when an operator sets
`threshold_pct` below ~33%. Minimum to use the feature is **1 provider** (N=1 →
quorum 1, self-attests); zero providers is also valid (attestation is opt-in).
Required quorum (online + agreeing) by provider count, default config:

| Providers N | bft_floor ⌊N/3⌋+1 | 67%-threshold ⌈0.67N⌉ | Required quorum | floor effect |
|---|---|---|---|---|
| 1 | 1 | 1 | 1 | none |
| 2 | 1 | 2 | 2 | none |
| 3 | 2 | 3 | 3 | none |
| 4 | 2 | 3 | 3 | none |
| 5 | 2 | 4 | 4 | none |
| 6 | 3 | 5 | 5 | none |

The floor only changes behavior on an `N >= 3` chain that deliberately runs a
low `threshold_pct` (e.g. `setsnpcfg(1,1)` → 2 votes required instead of 1).
Meaningful Byzantine tolerance (survive >=1 lying provider) needs `N >= 4`.
This interaction is also documented in the `setsnpcfg` doc-comment.


contracts/sysio.system/src/snapshot_attest.cpp
huangminghuang, line 122: The new floor counts registered providers, but votes are de-duplicated only by `snap_account` name and old pending votes are not invalidated when a provider unregisters. A single producer can register snap account A, vote, unregister it via `delsnapprov`, register snap account B for the same producer, and vote again for the same `(block_id, snapshot_hash)`. The `voters` vector now has two distinct names, so `voter_count` can satisfy the new `provider_count / 3 + 1` floor even though only one producer/provider identity participated. That defeats the security property this PR is adding. Consider voting by producer identity, rejecting provider rotation while pending votes exist, or purging/removing that provider's pending votes on unregister.

heifner: Good catch — fixed in 9aceced75c by tracking votes by the delegating **producer identity** instead of the snap_account (your first suggestion).

`votesnaphash` now resolves the producer from the provider row at vote time, stores the producer in `voters`, and de-duplicates/counts by producer. Since `regsnapprov` enforces a 1:1 producer↔provider mapping, a rotated snap_account (`regsnapprov A → vote → delsnapprov A → regsnapprov B → vote`) resolves to the same producer, so the second vote is rejected and `voter_count` stays at 1. This aligns the numerator (distinct voting producers) with the floor's denominator (registered providers). No ABI change — `voters` stays `name[]` and nothing outside the contract reads its content.

On "old pending votes aren't invalidated on unregister": with producer-keyed dedup a lingering vote from an unregistered provider is bounded to one per producer and represents a real past attestation, so it can't be weaponized into a false attestation or a Sybil — the security property holds without purging. Happy to also add purge-on-`delsnapprov` if you'd prefer the extra hygiene (keeping `voter_count ≤ provider_count` exactly).

Regression test `votesnaphash_rejects_snap_account_rotation_sybil` reproduces the exact rotation (4 providers, `setsnpcfg(1,1)` → floor 2): the rotated re-vote is rejected, the snapshot stays unattested, and only a genuinely distinct producer reaches quorum. Full suite 24/24, unit_test `snapshot_attest_tests` 5/5.

huangminghuang (approved):

PR #420, opened 2026-06-18
Fix compiler warnings in wire-sysio

huangminghuang:
## Summary

Clean up compiler warnings reported by the ubuntu24 GitHub Actions build.

## What changed

- Avoid Clang's self-move warning in the `variant` self-move assignment test while preserving coverage of assigning through an alias.
- Remove an unused lambda capture in producer schedule validation.
- Pass the batch operator status string directly to the error log instead of using an accidental comma expression.
- Remove unused constants from contract tests while keeping the relevant value documented.

## Why

The referenced CI job completed successfully but emitted source warnings in these files. Removing them keeps Release builds cleaner and makes future warnings easier to spot.

## Testing

- `git diff --check`

Not run locally: full `wire-sysio` build/test suite.

### DIFF
--- a/contracts/tests/emissions_tests.cpp
+++ b/contracts/tests/emissions_tests.cpp
@@ -212,7 +212,8 @@ static constexpr uint16_t GOV_BPS                = 1000;
 // lazily via sysio.dclaim::onreward -> sysio.system::fundclaim, not paid at
 // payepoch -- so it doesn't appear in t5state.total_distributed until the
 // underlying fundclaim transfer happens.
-static constexpr uint16_t IMPLICIT_CAPITAL_BPS   = 10000 - COMPUTE_BPS - CAPEX_BPS - GOV_BPS;
+// The implicit capital reserve at these defaults is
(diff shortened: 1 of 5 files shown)


contracts/tests/emissions_tests.cpp
```diff
@@ -212,7 +212,8 @@ static constexpr uint16_t GOV_BPS                = 1000;
 // lazily via sysio.dclaim::onreward -> sysio.system::fundclaim, not paid at
 // payepoch -- so it doesn't appear in t5state.total_distributed until the
 // underlying fundclaim transfer happens.
-static constexpr uint16_t IMPLICIT_CAPITAL_BPS   = 10000 - COMPUTE_BPS - CAPEX_BPS - GOV_BPS;
+// The implicit capital reserve at these defaults is
+// 10000 - COMPUTE_BPS - CAPEX_BPS - GOV_BPS = 3000 bps.
 static constexpr uint16_t PRODUCER_BPS           = 7000;
```
heifner, line 217: Why this change?

heifner (approved):

PR #424, opened 2026-06-19
clio: add `system setcode` / `system setabi` for sysio.roa deployment

heifner:
## Summary

Deploying a system contract goes through `sysio.roa` (`setsyscode`/`setsysabi`) rather than the chain's native `setcode`/`setabi`: the privileged roa contract performs the inline `setcode`/`setabi`, makes the target privileged, and gifts exactly the RAM the code/abi consumes out of sysio's pool (a conserving transfer, reclaimed on redeploy of smaller code).

Until now, driving these from clio meant a raw `push action` — fine for the WASM (`code` is just hex) but painful for the ABI, since `abi` must be a **packed `abi_def`** and clio only produces those bytes internally inside `set abi` (there is no one-shot "abi-file → packed-hex"). This adds first-class subcommands that mirror `set code` / `set abi` mechanically, but live under the **`system`** command group — because, like `system newaccount` / `system regproducer`, they depend on a deployed system contract (`sysio.roa`), unlike the chain-native `set code` / `set abi` which work on any account:

```
clio system setcode <account> <contract.wasm> -p sysio@active
clio system setabi  <account> <contract.abi>  -p sysio@active
```

## Implementation

- `create_setsyscode` / `create_setsysabi` build the `sysio.roa::setsyscode`/`setsysabi` actions, default-authorized by `sysio@active` (the actions `require_auth("sysio")`). Action data is packed against the live `sysio.roa` ABI via the same `variant_to_bin` path `create_transfer` uses, so the bytes match the chain serializer exactly. (The contract action names are unchanged — only the clio verb is new.)
- The `system setcode` / `system setabi` subcommands reuse the existing `set code` / `set abi` callbacks — same file reading, `fc::raw::pack(abi_def)` packing, duplicate check, and send path — selecting the roa-routed action via a small `sys_variant` flag. The shared callbacks and option variables are declared in the `set` block and remain in scope where the `system` group is built. Default permission is `sysio@active`.
- No combined `system setcontract` form (not needed).

## Testing

- **Offline:** the action-data packing round-trips byte-exact against the real `sysio.roa.abi` (`convert pack_action_data` / `unpack_action_data --abi-file`); field order matches the ABI.
- **`tests/clio_set_syscode_test.py`** (`add_np_test`): launches a ROA-active cluster, seeds a finite-RAM target via `forcereg` (giftram rejects an unlimited target), then drives the real commands end-to-end and verifies:
  - `system setcode` deploys code **and makes the account privileged** (the tell that the roa path ran, not a plain `setcode`),
  - `system setabi` deploys the ABI,
  - a redeploy replaces the code.

  Passes locally.

## Verification of placement

`clio system --help` lists `setcode` / `setabi`; `clio set --help` no longer does (only `code` / `abi` / `contract`). clio builds clean.


programs/clio/main.cpp
```diff
@@ -2865,7 +2923,8 @@ int main( int argc, char** argv ) {
       }
 
       if (!duplicate) {
-         actions.emplace_back( create_setcode(name(account), code_bytes ) );
+         actions.emplace_back( sys_variant ? create_setsyscode(name(account), code_bytes)
```
huangminghuang, line 2930: For `system setcode`, a byte-identical WASM is not necessarily a no-op: `sysio.roa::setsyscode` also sets the target privileged and runs RAM reconciliation. Reusing the native duplicate check means an account that already has the same code through native `set code` or a partial/manual setup will print “Skipping” and never get the ROA side effects. Please bypass this duplicate skip for `sys_variant`, or include the ROA side-effect state in the duplicate predicate.

programs/clio/main.cpp
```diff
@@ -2919,7 +2978,8 @@ int main( int argc, char** argv ) {
 
       if (!duplicate) {
          try {
-            actions.emplace_back( create_setabi(name(account), abi_bytes) );
+            actions.emplace_back( sys_variant ? create_setsysabi(name(account), abi_bytes)
```
huangminghuang, line 2988: For `system setabi`, matching raw ABI bytes still may need `sysio.roa::setsysabi` to reconcile gifted RAM. The inherited native duplicate check can skip the action entirely, so a target with the same ABI but unreconciled ROA accounting will not be fixed unless the user knows to pass `--suppress-duplicate-check`. Please disable this skip for the ROA variant or make it check the ROA accounting state too.

programs/clio/main.cpp
heifner: Fixed in 51584b14fc. The duplicate computation is now gated on `!sys_variant`, so the ROA path (`system setcode`) always emits the action even on byte-identical code — the native `set code` skip behavior is unchanged (`duplicate` defaults to false). This guarantees the `setsyscode` setpriv + RAM-reconcile side effects run. Added Test 2b to the integration test: re-runs `system setcode` with identical bytes and asserts the transaction is still pushed (via `-j` stdout).

programs/clio/main.cpp
heifner: Fixed in 51584b14fc — same `!sys_variant` gate on the abi duplicate check, so `system setabi` always emits `setsysabi` (RAM reconciliation runs) even on identical ABI bytes. Test 2b covers the abi path too.

programs/clio/main.cpp
heifner: Good catch on the CI failure. Resolution (per maintainer): a duplicate deploy is allowed to just fail rather than adding contract-side no-op logic. The clio duplicate-skip bypass stays — so an identical `system setcode` now *sends* the action and fails loudly at the chain ("contract is already running this version of code") instead of silently skipping (the original concern). The success-assertion I had added (Test 2b) was wrong and is removed in 23311ab6ce; Tests 1–3 still cover the real deploy paths. No contract change.

huangminghuang (approved):

PR #425, opened 2026-06-22
emissions: distribute swap-fee rewards bucket via payepoch; regctok precision_override

heifner:
## Summary

Finishes the fee-emissions work: distributes the swap-fee **rewards bucket** each epoch, and fixes the missing **precision input** on the chain-token registration action.

The per-epoch emissions distribution to active batch operators + producers already exists on master (`sysio.system::payepoch`, driven inline by `sysio.epoch::advance`). This PR adds the **fee** side and the precision field.

## Fee-bucket distribution

- `sysio.reserv`'s `rewards_bucket` (the 50% rewards share of collected swap fees — the other 50% already transfers to the `sysio` treasury at swap time) previously only ever accrued, with no payout path ("pending a future distribution action").
- New **`sysio.reserv::drainrewards(amount)`** (auth = `sysio`) sweeps the bucket WIRE to the treasury, preserving the custody invariant and leaving `lifetime_accrued` (the audit total) untouched.
- **`payepoch`** reads the bucket, queues the drain **first**, then folds the fee into the compute distribution — paid to producers + batch operators using the **same `producer_bps` / `batch_op_bps` split and the same weighting** as emissions.

### Consensus-critical: drain-before-payout ordering

The fee WIRE lives in `sysio.reserv`'s custody, but `payepoch` pays out of `sysio`'s balance. Inline actions execute **depth-first** (`apply_context::exec`), so queueing `drainrewards` ahead of the payout transfers guarantees the `reserv → sysio` sweep (and its nested transfer) completes before any payout — no overdraw. No `sysio.epoch` change is required (payepoch reads the bucket itself; the readiness gate is unaffected since the fee is extra inflow).

### Accounting

Fees are funded by the sweep, **not** the T5 treasury, so they are **excluded from `total_distributed`** — the emission decay curve is unchanged. `epoch_log` gains a `fee_distributed` field for forensics. Any fee not distributed (producer round-scaling, skipped slashed/terminated recipients, integer-division remainders) stays in the treasury, exactly as undistributed emission does.

## Precision input

`regctok` / `regchaintoken` gained the documented-but-missing **`precision_override`** input; `chain_token_row.precision_override` defaults to the canonical **9**-decimal depot frame (the header doc already promised this field).

## Testing

- **405/405** contract test cases pass (`contracts_unit_test`).
- New coverage:
  - `sysio_tokens_tests/regctok_records_precision_override` (new `sysio.tokens_tests.cpp`)
  - `sysio_reserve_tests::drainrewards_sweeps_bucket_to_treasury` + `drainrewards_auth_and_overdrain_guarded`
  - `t5_emissions_tests/payepoch_folds_swap_fee_rewards` — end-to-end: deploy reserv, seed the bucket via a real swap, advance to the pay-epoch, and assert the producer is paid `producer_pool + fee_share`, the bucket is swept to 0, the fee is excluded from `total_distributed`, and there is no overdraw.
- Rebuilt `wasm`/`abi` for `sysio.system`, `sysio.reserv`, `sysio.tokens`. No reference-data regeneration needed — `sysio.roa` / `sysio.msgch` / `sysio.uwrit` wasms are byte-identical after rebuild.

## Notes

Branches off the latest `master` (which already includes #417). Design choices (fold into `payepoch`, reuse the compute split) were confirmed before implementation.


contracts/sysio.system/src/emissions.cpp
```diff
@@ -135,6 +139,29 @@ int64_t get_wire_balance(name account) {
    return acct_tbl.get(key).balance.amount;
 }
 
+// Local mirror of sysio.reserv's rewards_bucket singleton. The kv row is keyed
+// by table name ("rewardbkt") + the reserv account scope, so a mirror with the
+// same table name and field layout reads the exact bytes reserv wrote. Mirrored
+// rather than #include'd so sysio.system does not pull sysio.reserv's AMM /
+// attestation headers into its translation unit. Layout MUST stay in lockstep
+// with sysio.reserv.hpp's rewards_bucket (balance, lifetime_accrued).
+struct reserv_rewards_bucket {
```
huangminghuang, line 150: This local mirror needs to stay serialization-compatible with sysio.reserv's rewards_bucket. Even with the warning comment, a future change to either side could silently break payepoch's view of the reserve bucket. Consider moving the bucket value type into a small shared header so both contracts serialize the same definition.

contracts/sysio.reserv/src/sysio.reserv.cpp
```diff
@@ -537,6 +537,34 @@ uint64_t reserve::rewardbal() {
    return bkt.get_or_default(rewards_bucket{}).balance;
 }
 
+void reserve::drainrewards(int64_t amount) {
+   // Only the system treasury (where sysio.system::payepoch runs) may sweep the
+   // rewards bucket. The swept WIRE is redistributed to producers + batch
+   // operators at the next pay-epoch.
+   require_auth(TREASURY_ACCOUNT);
+
+   if (amount <= 0) return; // defensive no-op (payepoch only calls with a positive amount)
```
huangminghuang: Because drainrewards is an internal treasury sweep action, a non-positive amount likely indicates caller or integration logic is wrong. Consider replacing this silent return with a check(amount > 0, ...) so bad calls fail loudly instead of looking successful while doing nothing.

contracts/sysio.system/src/emissions.cpp
```diff
@@ -135,6 +139,29 @@ int64_t get_wire_balance(name account) {
...
+using reserv_rewardbkt_t = sysio::kv::global<"rewardbkt"_n, reserv_rewards_bucket>;
+
+// Read the live swap-fee rewards balance held in sysio.reserv's custody.
+// Returns 0 when never accrued. Clamps to int64 max defensively (reserv caps
+// the bucket at uint64 max via saturating add).
+int64_t get_reserv_rewards_balance() {
+   reserv_rewardbkt_t bkt(RESERV_CONTRACT);
+   const uint64_t bal = bkt.get_or_default(reserv_rewards_bucket{}).balance;
+   constexpr uint64_t int64_max = static_cast<uint64_t>(std::numeric_limits<int64_t>::max());
+   return bal > int64_max ? std::numeric_limits<int64_t>::max() : static_cast<int64_t>(bal);
```
huangminghuang: Since this value is later drained as a sysio::asset, consider clamping to asset::max_amount rather than int64_t::max(). The reserve bucket should be backed by real WIRE and therefore probably cannot exceed the asset limit, but using the same bound here would make that invariant explicit and avoid constructing an out-of-range asset if accounting ever drifts.

contracts/sysio.tokens/src/sysio.tokens.cpp
```diff
@@ -107,6 +108,7 @@ void tokens::regctok(sysio::slug_name   chain_code,
       .chain_code         = chain_code,
       .token_code         = token_code,
       .contract_addr      = std::move(contract_addr),
+      .precision_override = precision_override,
```
huangminghuang, line 122: If there is a meaningful maximum precision this contract supports, consider validating precision_override before storing it. Accepting any uint32_t leaves room for bindings that downstream display/math code may not handle safely or consistently.

contracts/sysio.system/src/emissions.cpp
heifner: Good instinct, but this one runs into a CDT limitation. I tried it — moved the value type into a shared header in `sysio.opp.common`. A `[[sysio::table]]`-attributed struct pulled into **sysio.system's** translation unit corrupts that contract's read-only-action return-value codegen: `getpeerkeys` began failing with `out_of_range unpacking std::vector<...>`. Dropping the `[[sysio::table]]` attribute fixes sysio.system, but then **sysio.reserv's** ABI loses the `rewards_bucket` struct (the `rewardbkt` table references an undefined type, and reserv's own `get_rewardbkt` decode fails). So the two contracts genuinely need separate declarations.

I kept the local layout-compatible view, but hardened the drift risk you flagged:
- both definitions now carry a NOTE cross-referencing each other and this constraint, and
- the cross-contract read is covered end-to-end by `t5_emissions_tests/payepoch_folds_swap_fee_rewards`, which writes the bucket from sysio.reserv via a real swap and reads it back through sysio.system's view in `payepoch` — any field-layout divergence fails that test.

Happy to revisit if there's a CDT-friendly way to share an attributed table type across contracts. (3ebe99e72f)

contracts/sysio.reserv/src/sysio.reserv.cpp
heifner: Done — replaced the silent return with `sysio::check(amount > 0, "drainrewards: amount must be positive")`. `payepoch` only ever calls it with a positive amount, so a non-positive value is a caller bug and now fails loudly. Updated `drainrewards_auth_and_overdrain_guarded` to assert the rejection for both 0 and a negative amount. (3ebe99e72f)

contracts/sysio.system/src/emissions.cpp
heifner: Done — `get_reserv_rewards_balance` now clamps to `sysio::asset::max_amount` instead of `int64_t::max()`, so the returned value is always a constructible asset amount. As you noted it's a defensive guard (the bucket is backed by real WIRE), but it makes the invariant explicit. (3ebe99e72f)

contracts/sysio.tokens/src/sysio.tokens.cpp
heifner: Added a `MAX_TOKEN_PRECISION = 18` bound (10^18 is the largest power of ten that fits in int64, the asset amount type) and validate it in `regctok` (`precision_override`) — and, for consistency, in `regtoken` (`precision`) too, since the same concern applies to the base token concept. New test `regctok_rejects_precision_over_max` covers the boundary (19 rejected, 18 accepted). (3ebe99e72f)

huangminghuang (approved):

PR #426, opened 2026-06-23
sysio.roa: reconcile pre-existing reslimit at node-owner registration (SEC-087)

heifner:
### Problem
Any node owner can provision a `reslimit` row on an arbitrary account via `addpolicy` (it requires only the **issuer's** auth — no target consent). `nodeownreg`'s check (6) then rejected the claim with `OWNER_HAS_RESLIMIT` whenever such a row already existed, because `regnodeowner` created the row from scratch and `set_reslimit` asserts it is absent. Result: a planted policy could **block a victim's legitimate registration indefinitely** (a denial-of-registration grief).

### Fix — reconcile instead of reject
`regnodeowner` now stacks the node-owner allocation onto whatever the account already holds, via the existing `increase_reslimit` helper, and syncs the on-chain quota to the resulting row totals:

- **Fresh path is unchanged.** `increase_reslimit`'s create branch folds in the one-time `newaccount_ram` gift, so a brand-new node owner still gets exactly `newaccount_ram + personal_ram_bytes` — byte-for-byte identical to the old `set_reslimit` + absolute `set_resource_limits`. The `verify_ram` parity anchor stays green unchanged.
- **Pre-existing row is reconciled.** The modify branch adds only the personal delta, so the gift is never double-counted (the row was already created — with the gift — by the planting `addpolicy`, which routes through the same helper).
- The planted policy is left in place: it stays backed by its issuer's reserve and remains reclaimable by the issuer via `reducepolicy`. No value is stolen, and the claim reaches `CONFIRMED`.

The root-cause vector (`addpolicy` without target consent) is intentionally **not** changed: that is Wire's sponsor-pays model (a node owner debits its own reserve to gift resources to users who hold no SYS), and post-reconcile a planted policy is harmless — a gift at the attacker's own expense that can no longer block registration, can't be yanked while in use (`validate_ram_usage`), and blocks no other issuer or the victim's own registration.

### Overflow hardening (folded in)
`increase_reslimit` did raw `+=` on `asset.amount` (signed-overflow **UB**) and `ram_bytes` (silent `uint64` wrap). This adds a saturating signed add `add_sat_i64` beside the existing `add_sat_u64` in `sysio.opp.common/safe_ops.hpp` and uses saturating adds in both branches; the `ram_bytes >= 0` guard is hoisted to cover the create branch too. These **saturate rather than `check()`-throw** on purpose: `increase_reslimit` sits on the OPP never-throw dispatch path (`nodeownreg → regnodeowner`, and `newnameduser`), where an abort would roll back the consensus-tipping delivery and stall epoch advancement. The saturation cap is unreachable for any weight bounded by the SYS supply, so behaviour is unchanged for all realistic inputs.

### Pre-launch cleanup
- Removed the now-unused `OWNER_HAS_RESLIMIT` reject reason from the enum and the matching test constant (pre-launch — no persisted rows to migrate).
- Fixed a stale "would throw 'Resource limit already exist'" comment in `sysio.dispatch_tests`.

### Tests
- Inverted `nodeownreg_owner_has_reslimit` → `nodeownreg_reconciles_existing_reslimit`: a planted RAM-only policy now yields `CONFIRMED`, the reslimit reconciles to `{0.0500, 0.0500, 1,049,464}` (planted + gift + personal), the on-chain quota is synced to those totals, the planted policy is intact, and the issuer's allocations are unchanged.
- Added `nodeownreg_reconciles_existing_reslimit_net_cpu`: a net/cpu-planted policy also stacks correctly.
- Added `add_sat_i64` coverage in `safe_ops_tests`.

### Verification
- `contracts_unit_test` (owns the change): green across `safe_ops_tests`, `sysio_roa_tests` (incl. `verify_ram` parity anchor + both new reconcile tests), `sysio_dispatch_tests`, `sysio_emissions_tests` (forcereg).
- `deep_mind_tests`, `snapshot*`, `plugin_test`: green. `sysio.roa.wasm` rebuilt and committed; **ABI unchanged** (no action/table signature change), so no reference-data regen required.

### Notes
- `forcereg` over a pre-existing row now reconciles too (previously threw); intended, and no test relied on the throw.
- For a `sysio.*`-prefixed owner the new path writes row and chain quota consistently (the old code left them divergent); unreachable by real registration flows.


contracts/sysio.roa/sysio.roa.cpp
```diff
@@ -866,20 +869,21 @@ namespace sysio {
...
+        // aborting, and never double-counting the gift. Then sync the on-chain quota to the resulting
+        // row totals so the reslimit row and the resource limits stay equal on BOTH the fresh and
+        // reconcile paths.
+        //   Conservation: production (newnameduser -> nodeownreg) funds newaccount_ram at creation, so
+        //   the create path stays conservation-exact; the reconcile path adds only the node-owner delta.
+        //   forcereg over a pre-existing row now reconciles too (previously threw in set_reslimit) --
+        //   acceptable for the governance/bootstrap path.
+        roa::resources_t owner_res = increase_reslimit(owner, net_cpu_weight, net_cpu_weight,
+                                                       (int64_t)personal_ram_bytes, /*require_to_exist=*/false);
+        set_resource_limits(owner, (int64_t)owner_res.ram_bytes, owner_res.net.amount, owner_res.cpu.amount);
```
huangminghuang, line 889: Now that increase_reslimit uses saturating helpers for accumulator updates, it may be worth applying the same convention to the resource-limit quota arithmetic as a follow-up. The existing activation/allocation bounds appear to keep those additions reachable-safe, so this is not a blocker, but using the same safe arithmetic for both reslimit rows and quota sync would make the resource accounting easier to audit.

heifner: Good call — folded it in rather than deferring (e083009160). `add_system_resources` now uses `add_sat_i64` for the three `current + delta` quota additions, and the one reslimit-row accumulator that doesn't route through `increase_reslimit` (the sysio reslimit `+=` in `regnodeowner`) now uses `add_sat_u64`. So the reslimit rows and the chain-quota sync are hardened with the same saturating, never-throw convention. No behavior change for realistic values (the cap is unreachable under the activation/allocation bounds); contracts_unit_test green against the rebuilt wasm. Left the SYS *allocation* accounting (`nodeowners.allocated_*`, policy weights) as-is since it's a separate concern from resource-limit quota.

huangminghuang (approved):

PR #430, opened 2026-06-24
SEC-58: Fix P2P session and transaction generator handling

huangminghuang:
## Problem

SEC-58 / WSA-133 identified that `net_plugin` accepted some stateful P2P traffic before a peer had completed the initial session handshake and, when configured, peer authentication. In that window a peer could send sync-oriented messages before its chain/protocol identity was established, which could make nodeop start expensive block-log streaming work for a connection that had not passed the session gate.

The regression surfaced in the long-running startup/catchup path because `trx_generator`'s P2P provider did not behave like a normal peer: it connected and immediately streamed `transaction_message` frames without first sending a handshake or answering nodeop heartbeat/time messages. Once non-control traffic is gated until the session is established, the generator must participate in the small amount of P2P session traffic nodeop expects.

While fixing that path, the P2P provider also had a transport accounting bug: failed socket writes could be treated as successful transaction acknowledgements. That made the generator capable of reporting 100% delivery even when the peer had closed the socket and no transaction traffic was delivered.

## Resolution

- Added a protocol-level helper for pre-session control messages and shared the wire protocol base version through `protocol.hpp`.
- Added `peer_session_established` tracking in `net_plugin` and gate all non-control messages until both the initial handshake and any required peer auth have completed.
- Deferred sync negotiation until the peer has passed the session gate, then re-ran handshake handling after peer auth succeeds.
- Updated `trx_generator` P2P mode to send an initial handshake, maintain a peer read loop, answer heartbeat handshakes and time pings, and use the latest peer LIB as the TAPOS reference block.
- Hardened P2P write handling so handshake/time/transaction write failures mark the transport unhealthy, close the socket, stop the read loop, and are no longer counted as successful transaction acknowledgements.
- Fixed handshake generation wraparound so the next generation after `int16_t::max()` avoids nodeop's special generation-1 initial-handshake path.
- Moved shared trx_generator wire constants into named `constexpr` values and added regression coverage for the new handshake and failed-write behavior.

## Tests

- `git diff --check`
- `ninja -C build/sec-58 -j$(nproc) trx_generator trx_generator_tests test_net_plugin`
- `./build/sec-58/plugins/net_plugin/test/test_net_plugin --run_test=net_message_session_gate`
- `./build/sec-58/tests/trx_generator/trx_generator_tests --run_test=trx_generator_tests/p2p_provider_sends_handshake_on_setup`
- `./build/sec-58/tests/trx_generator/trx_generator_tests --run_test=trx_generator_tests/p2p_provider_write_failure_does_not_ack_transaction`
- `./build/sec-58/tests/trx_generator/trx_generator_tests`
- `ctest -R '^nodeop_startup_catchup_lr_test$' -j$(nproc) --output-on-failure --timeout 1800` passed in 561.92s


heifner (approved):
Approving — reviewed the initial commit and the `615579c8` follow-up.

**The security fix is sound and fail-closed.** Non-control messages are dropped until both the initial handshake and (when configured) peer auth complete, with sync negotiation deferred and correctly re-run after `peer_auth`. This is all plugin-zone — it gates *which* P2P messages get processed and never touches block/transaction validation — so there's no consensus-determinism risk. The new connection flags are strand-confined, and the `is_pre_session_control_message` helper is exhaustively unit-tested across all message types.

The follow-up cleanly addressed the review feedback:
- Initial `gossip_bp_peers` is now posted to the strand behind the handshake/auth frames, so it's no longer dropped by the peer's own session gate.
- TAPOS now references the peer's irreversible fork-db root instead of the reversible head — fresher than the static startup LIB, but fork-proof.
- `_transport_failed` moved to acquire/release ordering; the read-loop error paths are consolidated through `mark_transport_failed`; trailing-byte and length bounds are hardened.
- Handshake generation wraparound now matches nodeop (skips the special generation-1 path).

The transport-accounting fix (a failed socket write is no longer counted as a successful ack) is correct and well covered by `p2p_provider_write_failure_does_not_ack_transaction`.

PR #431, opened 2026-06-24
SEC-56: Bound state-history status request queues

huangminghuang:
## Problem

SEC-56 / WSA-124 identified that each state-history WebSocket session could append every `get_status_request` into an unbounded per-session queue. A client that sent status requests faster than the write loop drained them could grow `queued_status_requests` without limit, consuming memory and forcing the node to serialize and send one status response per queued request.

That made the status endpoint vulnerable to request-queue amplification: the client did not need to request block data or hold many sessions open; one session could accumulate work and memory between write-loop drains.

## Solution

This PR adds a bounded `status_request_queue` utility for SHiP status requests:

- caps pending status requests at `max_status_request_queue_depth` (`1024`) per session queue drain window
- tracks pending v0/v1 status requests as counts, giving O(1) queue memory and avoiding per-request deque allocation
- preserves v0/v1 request identity with a typed `status_request_version` append API
- returns `false` from `try_append()` when the queue is full, with `[[nodiscard]]` so future call sites cannot accidentally ignore the security gate
- drains via `pop_all()` into `pending_status_request_counts`, leaving the queue reusable for later requests
- closes the session by throwing through the existing coroutine error path when a client exceeds the queue-depth cap

The cap is intentionally a queue-depth limit, not a lifetime counter: normal clients can continue sending status requests after the write loop drains pending work, while abusive bursts cannot grow memory without bound.

The counted drain intentionally does not retain interleaved v0/v1 ordering within one drain. That is safe for SHiP status responses because every response in a drain carries the same status snapshot, and a session uses one status protocol version.

## Tests

Latest validation after review follow-up:

- `ninja -C build/sec-56 -j6 unit_test`
- `./build/sec-56/unittests/unit_test --run_test=test_state_history/status_request_queue_is_bounded -- --sys-vm`
- `./build/sec-56/unittests/unit_test --run_test=test_state_history -- --sys-vm`
- `ninja -C build/sec-56 -j6 state_history_plugin nodeop`
- `ctest -j "$(nproc)" -R '^test_state_history_unit_test_' --output-on-failure --timeout 1000`
- `ctest -j "$(nproc)" -R '^ship_test$' --output-on-failure --timeout 1000`

Earlier broader validation on the initial patch before the review follow-up:

- `ninja -C build/sec-56 -j"$(nproc)"`
- `ctest -j "$(nproc)" -LE '(nonparallelizable_tests|long_running_tests|wasm_spec_tests)' --output-on-failure --timeout 1000`

That broad CTest sweep passed 320/326 tests. The six failures were local build/configuration checks unrelated to SEC-56:

- `native_overlay_unit_test_sys-vm-oc`, `native_overlay_unit_test_sys-vm`, `native_overlay_unit_test_sys-vm-jit`: this Debug worktree was configured without system contract native `.so` artifacts because CDT was unavailable
- `release-build-test`: expected failure for a Debug build
- `version-label-test`, `full-version-label-test`: local worktree reports version `unknown`

## Review

Addressed the optional reviewer suggestion to replace `std::deque<bool>` with counted v0/v1 pending requests. Also ran a read-only Claude review of the original patch; it found no correctness or behavioral regressions, and its hardening suggestions (`[[nodiscard]]` and clearer queue-depth naming) are included.

### DIFF
--- a/libraries/state_history/include/sysio/state_history/status_request_queue.hpp
+++ b/libraries/state_history/include/sysio/state_history/status_request_queue.hpp
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <cstddef>
-#include <deque>
 #include <string_view>
 
 namespace sysio::state_history {
@@ -13,21 +12,52 @@ inline constexpr std::size_t max_status_request_queue_depth = 1024;
 inline constexpr std::string_view status_request_queue_limit_exceeded =
       "state history status request queue limit exceeded";
 
-/** Bounded FIFO for pending SHiP status requests. */
+/// SHiP status request protocol version.
+enum class status_request_version {
+   v0,
+   v1,
+};
+
+/** Counts pending SHiP status requests by protocol version for one queue drain. */
+struct pending_status_request_counts {
+   std::size_t v0 = 0; ///< Number of pending v0 status requests.
+   std::size_t v1 = 0; ///< Number of pending v1 status requests.
+
+   /// Returns the total number of pending status requests.
+   std::size_t size() const {
+      return v0 + v1;
+   }
+
+   /// Returns true when no status requests are pending.
+   bool empty() const {
+      return size() == 0;
+   }
+};
+
+/** Bounded counter queue for pending SHiP status requests. */
 class status_request_queue {
 public:
    /// Attempts to append one status request, returning false when the queue depth limit has been reached.
-   [[nodiscard]] bool try_append(bool is_v1_request) {
+   [[nodiscard]] bool try_append(status_request_version version) {
       if(requests.size() >= max_status_request_queue_depth)
          return false;
-      requests.emplace_back(is_v1_request);
+
+      switch(version) {
+      case status_request_version::v0:
+         ++requests.v0;
+         break;
+      case status_request_version::v1:
+         ++requests.v1;
+         break;
+      }
+
       return true;
    }
 
    /// Removes and returns all pending status requests while leaving the queue reusable.
-   std::deque<bool> pop_all() {
-      std::deque<bool> pending;
-      pending.swap(requests);
+   pending_status_request_counts pop_all() {
+      const pending_status_request_counts pending = requests;
+      requests = {};
       return pending;
    }
 
@@ -42,7 +72,7 @@ public:
    }
 
 private:
-   std::deque<bool> requests; ///< false for v0 requests, true for v1 requests.
+   pending_status_request_counts requests; ///< Counted pending requests since the last drain.
 };
 
 } // namespace sysio::state_history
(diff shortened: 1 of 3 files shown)


libraries/state_history/include/sysio/state_history/status_request_queue.hpp
```diff
@@ -0,0 +1,48 @@
...
+      return requests.size();
+   }
+
+   /// Returns true when no status requests are queued.
+   bool empty() const {
+      return requests.empty();
+   }
+
+private:
+   std::deque<bool> requests; ///< false for v0 requests, true for v1 requests.
```
heifner: **Optional: the status queue could be two counters instead of `std::deque<bool>`**

Nice fix — the depth bound and the `[[nodiscard]]` gate are the right shape. One non-blocking simplification to consider:

The queue doesn't actually need to retain a *sequence*. In `write_loop` the status snapshot is computed once per drain and replayed for every pending request — the only per-entry difference is the v0 vs v1 wire format. Since a SHiP session speaks a single protocol version, all that's needed is the count of each:

```cpp
struct pending_counts {
   std::size_t v0 = 0, v1 = 0;
   std::size_t total() const { return v0 + v1; }
   bool        empty() const { return total() == 0; }
};

[[nodiscard]] bool try_append(bool is_v1_request) {
   if(counts.total() >= max_status_request_queue_depth) return false;
   if(is_v1_request) ++counts.v1; else ++counts.v0;
   return true;
}
pending_counts pop_all() { pending_counts c = counts; counts = {}; return c; }
```

and the drain consumer becomes two count loops instead of iterating bools:

```cpp
for(std::size_t i = 0; i < status_requests.v0; ++i)
   co_await stream.async_write(/* get_status_result_v0 */);
for(std::size_t i = 0; i < status_requests.v1; ++i)
   co_await stream.async_write(/* get_status_result_v1 */);
```

Same bound, same `[[nodiscard]]` gate, but O(1) memory, zero allocation, and it drops the `<deque>` include.

The one behavioral change is that interleaved v0/v1 ordering *within a single drain* is no longer preserved. That's safe here: every result in a drain carries the same snapshot, and a connection only ever uses one version. I checked Wire-Hyperion to be sure — it sends a single `get_status_request_v0` per session at bootstrap (`state-reader.ts:612`) and then streams blocks, so it's all-v0 and never queues more than one anyway.

huangminghuang: Addressed in 1ab8c0ce05: `status_request_queue` now stores counted v0/v1 pending requests instead of `std::deque<bool>`, and the writer drains those counts in version batches. I also updated the focused queue test to assert the counted drain behavior.

heifner (approved):

PR #434, opened 2026-06-25
Uwrit: fail closed on a zero AMM quote from an ACTIVE reserve (WSA-041)

heifner:
## WSA-041 — fail closed on a zero AMM quote from an ACTIVE reserve

**P1 High** · sysio.uwrit / reserves / from-WIRE settlement

### Bug
`sysio.uwrit::swap_quote` returns 0 both when a required reserve is missing/not-ACTIVE **and** when an ACTIVE reserve is degenerate — a side drained to zero, `connector_weight_bps` at the boundary, or a src/WIRE leg so small the weighted-Bancor kernel floors the output to 0. Every ingestion/race site treated *any* zero quote as "no LP provisioned, skip the variance check", so the caller-supplied `target_amount` flowed straight into the uwreq and on to settlement. `sysio.reserv::applyfromwire` debits `dst_amount` directly after only a `reserve_chain_amount >= dst_amount` check — so a negligible WIRE escrow could drain a reserve's token side at a caller-chosen price.

`connector_weight_bps == 10000` is already rejected at creation (`MAX_CONNECTOR_WEIGHT_BPS = 9999`), but a zero quote is **not** unique to that boundary — a small enough `wire_in` floors the quote to 0 on any large reserve — so the cap alone does not close the hole.

### Fix
Distinguish "unprovisioned (skip)" from "ACTIVE but unpriceable (fail closed)" via `required_reserves_active()`:
- **createuwreq** — emit SWAP_REVERT, create no uwreq.
- **try_select_winner** — reject + refund at race resolution.
- **drainfwq** — refund (the target reserve is already confirmed ACTIVE there, so a zero quote is unconditionally fail-closed).

Dev/smoke clusters without provisioned LPs still skip (the reserve is not ACTIVE). `connector_weight_bps` proto docs tightened to `1..9999`.

### Test
`sysio_dispatch_tests/swap_zero_quote_from_active_reserve_fails_closed` — a `src_amount = 1` swap against ACTIVE 1e12/1e12 reserves yields a zero quote and must create no uwreq.

### Scoping
Settlement-level repricing in `applyfromwire` (a chokepoint defense-in-depth) is left as a follow-up — the ingestion/race fail-closed prevents an unpriceable target from ever reaching settlement.


libraries/opp/proto/sysio/opp/bootstrap/bootstrap.proto
```diff
@@ -110,8 +110,10 @@ message ReserveSpec {
   uint64 initial_chain_amount  = 6;
   // WIRE-side seed, in WIRE subunits, drained from the earmark (quote in JSON).
   uint64 initial_wire_amount   = 7;
-  // Bancor connector weight (WIRE-side) in basis points: 1..10000; 5000 =
-  // 50/50 pure constant product. Honored by the weighted swap curve.
+  // Bancor connector weight (WIRE-side) in basis points: 1..9999; 5000 =
```
huangminghuang, line 113: The bootstrap proto/test now reject `10000`, but `docs/platform-bootstrap-config.md` still lists the V6 invariant as `0 < connector_weight_bps <= 10000`. Please update that doc to `<= 9999` too so bootstrap authors do not follow stale launch guidance.

heifner: Fixed in 0d3436297f — updated the V6 invariant in `docs/platform-bootstrap-config.md` from `connector_weight_bps ≤ 10000` to `≤ 9999`, matching the proto and the `BootstrapPlatformConfig` validator.

(The V9 `fee_bps ≤ 10000` line on the same table is the separate uwrit bound — that one is tightened, validator + doc, in #436.)

huangminghuang (approved):
Looks good. The follow-up aligns the bootstrap docs with the corrected connector_weight_bps bound, and the focused contract suites passed locally.

PR #436, opened 2026-06-25
Uwrit/reserv: reject 100% swap fee + fail-closed settlement (SEC-26 / WSA-042)

heifner:
## SEC-26 / WSA-042 — reject 100% swap fee + fail-closed reserve settlement

**P1 High** · `sysio.uwrit` config + `sysio.reserv` settlement + platform bootstrap config

### Bug
`sysio.uwrit::setconfig` accepted `fee_bps == 10000` (`fee_bps <= 10000`). A 100% fee makes `opp::amm::split_wire_fee` return `net == 0` — the whole WIRE leg is consumed as fee. Reserve settlement then credited only `fee.net` (== 0) WIRE to the destination reserve while debiting the full caller-supplied target:

- `sysio.reserv::applyfromwire` — `reserve_wire_amount += fee.net` (0), `reserve_chain_amount -= dst_amount` after only a `reserve_chain_amount >= dst_amount` check.
- `sysio.reserv::applyswap` — same, for token-to-token.

So a from-WIRE swap with a negligible WIRE input and a large `target_amount` drained the destination reserve's token side at an arbitrary price. The platform bootstrap config validator independently accepted the same `fee_bps == 10000`, so a genesis manifest could ship a 100% fee.

### Fix
This PR adds the items the WSA-041 fix (#434) explicitly left as follow-ups, hardening every fee-bps entry point:

1. **On-chain config** — `setconfig` now rejects `fee_bps >= 10000` via `MAX_FEE_BPS = 9999` (mirrors `sysio.reserv::MAX_CONNECTOR_WEIGHT_BPS`). Any sub-100% fee leaves a positive post-fee remainder for **every** positive input (`net = amt − floor(amt·bps/10000) ≥ 1` for `bps ≤ 9999`), so `net == 0` becomes unconfigurable.
2. **Platform bootstrap config** — the bootstrap manifest validator's V9 check now rejects `fee_bps > 9999` (was `> 10000`), so a genesis manifest cannot ship a 100% fee either. The `bootstrap.proto` field doc is updated to match, and the bound mirrors the on-chain cap.
3. **Settlement layer (defense-in-depth)** — `applyfromwire` / `applyswap` assert `fee.net > 0` before debiting destination reserve liquidity. Given the config caps this is unreachable-by-construction, so it never fires in the consensus dispatch chain (respects the OPP-handlers-never-throw invariant) — it is a backstop, matching the existing pre-validated `check()`s in those functions.

### Relationship to #434 (WSA-041 / SEC-25)
WSA-041 and WSA-042 are the same underlying defect: a zero/degenerate quote bypassing the variance check, letting a caller-controlled target reach settlement. #434 fixes the shared root cause — fail-closed on a zero AMM quote from an ACTIVE reserve at `createuwreq` / `try_select_winner` / `drainfwq` — which already closes the WSA-042 drain (a 100% fee yields a zero quote). This PR is complementary and independent: it touches `sysio.reserv.cpp` (untouched by #434), the `setconfig` function, and the bootstrap validator (all far from #434's hunks), and it hardens the config and settlement chokepoints #434 deferred.

### Tests
- `sysio_uwrit_tests`: `setconfig` rejects `10000` and `10001`, accepts `9999`.
- `amm_math_tests/split_wire_fee_boundaries`: `net == 0` only at 100% (and the over-100% clamp); `net > 0` and `net + fee == amount` for every sub-100% fee across dust-to-large inputs.
- `test_bootstrap_platform_config`: V9 rejects `fee_bps == 10000`.

### Notes
- No ABI change (action signatures and tables unchanged) — `.abi` files are byte-identical; no `SystemContractTypes` regeneration or companion `wire-libraries-ts` branch needed. The committed `.wasm` for both contracts is refreshed.


docs/platform-bootstrap-config.md
```diff
@@ -140,7 +140,7 @@ because the file is hand-authored and drives irreversible actions. A validator
 | V6 | reserve `(chain, token, code)` unique; references a declared binding; not on the depot; `0 < connector_weight_bps ≤ 10000`; amounts > 0 |
 | V7 | Σ `initial_wire_amount` ≤ `t5_reserve_allocation` > 0 |
 | V8 | `is_private` ⇒ `owner` a valid account name; else `owner` empty |
-| V9 | `uwrit` present; `fee_bps ≤ 10000`; lock duration > 0 |
+| V9 | `uwrit` present; `fee_bps ≤ 9999`; lock duration > 0 |
```
huangminghuang, line 143: This PR now rejects `fee_bps == 10000` in `setconfig` and in the bootstrap validator/docs, but `libraries/opp/proto/sysio/opp/bootstrap/bootstrap.proto` still documents `UwritConfig.fee_bps` as `<= 10000`. Please update that proto comment to `<= 9999` as well so generated/bootstrap config authors do not follow stale schema guidance.

heifner: Done in 87fdf0ea14 — updated the `UwritConfig.fee_bps` comment in `bootstrap.proto` to `<= 9999` (noting a 100% fee is rejected), matching the `setconfig`/bootstrap-validator behavior. Thanks for catching the stale schema guidance.

huangminghuang (approved):
Re-reviewed current head 478ad934. No findings; focused OPP bootstrap config and contract suites pass locally.

PR #437, opened 2026-06-25
Opp-bundler: fix protoc-gen-ts resolution and propagate publish failures

heifner:
The OPP model packages (@wireio/opp-solidity-models, @wireio/opp-typescript-models) have not been republished since 2026-05-27, predating the entire DEX precision feature, despite multiple merges touching libraries/opp/proto. The opp-bundles workflow reported success on every run while publishing nothing.

There are four causes in wire-protobuf-bundler, the last three exposed in sequence as each preceding one is fixed:

- protoc-gen-ts could not be resolved. generate-opp-bundles.fish runs the globally-linked CLI from the repo root, so the bundler's CWD-relative node_modules/.bin lookups missed. protoc-gen-solana/-solidity are linked onto the global PATH and survived; @protobuf-ts/plugin's protoc-gen-ts is not, so the typescript target threw and the solidity target never ran.

- bundleCommand logged the failure but did not rethrow, so the process exited 0 and CI reported green even though no package was generated or published.

- protoc was invoked as bare `npx protoc`, which from the repo-root CWD ignores the pinned protoc@^34 and downloads the latest protoc instead.

- @protobuf-ts/plugin@2.11.1 (the latest published) targets the TypeScript 3.9 factory API, but the tools workspace pins typescript@6.0.2 globally via package.json "resolutions", which pnpm forces onto the plugin as well, crashing protoc-gen-ts (ts.createTypeReferenceNode was removed in 5.0; the createInterfaceDeclaration signature changed in 4.8).

Changes: resolvePluginBin and a new resolveProtocBin resolve plugin and protoc binaries from the bundler's own install tree first, anchored on __dirname and walking up to the nearest node_modules/.bin/<name>, independent of the working directory. bundleCommand re-throws after logging so a failed generate/publish exits non-zero. A scoped resolution "@protobuf-ts/plugin>typescript": "3.9.10" (the plugin's declared range, and the version the known-good install resolves) keeps the plugin on a compatible compiler while the bundler itself keeps typescript@6.

With these, wire-protobuf-bundler generates all three targets (solana, typescript, solidity) from a clean install; verified locally and in the PR's opp-bundles run. The PR run validates generation without publishing (publish is gated on push to master); on merge it republishes the synchronized 1.0.28 models carrying source_token_precision and NodeOwnerTier.

### DIFF
--- a/libraries/opp/tools/package.json
+++ b/libraries/opp/tools/package.json
@@ -24,7 +24,8 @@
   "resolutions": {
     "lodash": "4.17.21",
     "prettier": "3.8.1",
-    "typescript": "6.0.2"
+    "typescript": "6.0.2",
+    "@protobuf-ts/plugin>typescript": "3.9.10"
   },
   "packageManager": "pnpm@10.32.1",
   "engines": {
--- a/libraries/opp/tools/protobuf-bundler/src/commands/bundle.ts
+++ b/libraries/opp/tools/protobuf-bundler/src/commands/bundle.ts
@@ -84,6 +84,10 @@ export async function bundleCommand(args: BundleArgs): Promise<void> {
   } catch (err: any) {
     skipCleanup = true
     log.error(`Bundle failed: ${err.message}`, err)
+    // Re-throw so the failure propagates to `main().catch` → process.exit(1).
+    // Without this the CLI exits 0 on a failed generate/publish and CI reports
+    // a green run even though no package was produced or published.
+    throw err
   } finally {
     if (!skipCleanup) {
       try {
--- a/libraries/opp/tools/protobuf-bundler/src/steps/generateTypescript.ts
+++ b/libraries/opp/tools/protobuf-bundler/src/steps/generateTypescript.ts
@@ -3,7 +3,7 @@ import Fs from "node:fs"
 import Path from "node:path"
 import { log } from "../util/logger.js"
 import { renderTemplate } from "../util/templates.js"
-import { resolvePluginBin, findProtoRoot } from "./runProtoc.js"
+import { resolvePluginBin, resolveProtocBin, findProtoRoot } from "./runProtoc.js"
 import { Target, TemplatePath, TemplateFile } from "../constants.js"
 
 export interface GenerateTypescriptOptions {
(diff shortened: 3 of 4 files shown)


libraries/opp/tools/package.json
```diff
@@ -24,7 +24,8 @@
   "resolutions": {
     "lodash": "4.17.21",
     "prettier": "3.8.1",
-    "typescript": "6.0.2"
+    "typescript": "6.0.2",
+    "@protobuf-ts/plugin>typescript": "3.9.10"
```
huangminghuang, line 28: This new override needs the matching `pnpm-lock.yaml` update committed. As checked in, `libraries/opp/tools/pnpm-lock.yaml` still has only the old overrides, so `pnpm install --frozen-lockfile` fails with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`. It also means a local bundle run still resolves `@protobuf-ts/plugin@2.11.1` against TypeScript 6 and fails generation with `--ts_out: TypeError: ts.createTypeReferenceNode is not a function`. Regenerating the lockfile adds this override and pins the plugin's nested `typescript` to `3.9.10`, so please commit that lockfile delta with this manifest change.

heifner: Thanks for the close look — but `pnpm-lock.yaml` isn't actually checked in here, so there's no stale lockfile to update. It's gitignored at `libraries/opp/tools/.gitignore:8` (added back in `d835bdebe3` when these tools were moved into wire-sysio), and there's no committed lockfile on `master` or in this PR. The `resolutions` entry in `package.json` is the source of truth for the pin.

Both consumers run a plain `pnpm install` — never `--frozen-lockfile`:
- `libraries/opp/tools/scripts/generate-opp-bundles.fish:31`
- `.github/workflows/opp-bundles.yaml:45`

A non-frozen `pnpm install` reconciles the lockfile to `package.json`, so the new `@protobuf-ts/plugin>typescript: 3.9.10` override is applied on a clean checkout and the plugin resolves TS 3.9.10 — i.e. the `--ts_out: TypeError: ts.createTypeReferenceNode is not a function` path doesn't happen in our flow. `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` only fires under `--frozen-lockfile`; if you hit it locally it's from a pre-existing (gitignored) lockfile that predates the override, and a plain `pnpm install` regenerates it in sync.

I verified locally: with the current `package.json`, `pnpm install --frozen-lockfile` passes (no `CONFIG_MISMATCH`), confirming a freshly-generated lockfile is already in sync with the new override.

Separately — your deeper point is fair: keeping the lockfile gitignored means transitive deps float within semver on a clean install, which is at odds with the "deterministic generation" goal. Committing the lockfile (and dropping the `.gitignore` line) would lock that down and let CI use `--frozen-lockfile`. That's a policy change beyond this PR's scope, so I'd rather do it as a focused follow-up if we want it — happy to open one. For this PR the `package.json` override is sufficient for the flows we actually run.

huangminghuang (approved):

PR #439, opened 2026-06-25
Contracts: enforce inbound OPP source-chain binding (WSA-005)

heifner:
## SEC-3 / WSA-005 (P1)

Value-bearing inbound OPP dispatch trusted payload-level chain identifiers without checking them against the proven source outpost.

### Problem

A consensus-reached inbound envelope is delivered for exactly one source outpost — the `deliver()` `chain_code`, which `sysio.msgch` validates as an active, non-depot `sysio.chains` row. But the value-bearing dispatch paths decoded each attestation's own chain identifier from the payload and acted on that, casting the proven `from_chain` to void. Batch-operator consensus proves only that the delivering outpost's operators agree on the bytes — not that the user acted on the chain the payload names. A compromised or malicious operator quorum for outpost A could therefore deliver a payload claiming chain B and drive deposits, withdrawals, swaps, underwrite commits, reserve mutations, or staking credits against an unrelated chain B's ledger.

### Fix

Bind every value-bearing payload's embedded chain identifier to the proven source outpost before any inline state-changing action. The invariant — `payload.chain_code == proven chain_code` — holds because an OPP attestation about chain X is always relayed by outpost X (the proto documents `ReserveCreate.external_amount.chain_code` as "the outpost's own chain", and each swap leg's UIC is delivered by its own outpost).

- `sysio.msgch` — `OPERATOR_ACTION` (deposit/withdraw), `UNDERWRITE_INTENT_COMMIT`, `STAKING_REWARD`, `RESERVE_CREATE`, `RESERVE_CREATE_CANCEL`: drop the offending attestation on mismatch (logged, never thrown — a throw inside the evalcons dispatch chain stalls consensus chain-wide).
- `sysio.uwrit::createuwreq` — `SwapRequest.source_chain_code`: refund via `SwapRevert` to the proven outpost rather than settling against the payload-named source reserve.

The now-redundant `from_chain` ChainKind argument is removed from the dispatch path (the exact `chain_code` match subsumes the VM-family check), along with its per-envelope `sysio.chains` read.

### Tests

New `sysio_dispatch_tests`: deposit/withdraw, swap, and underwrite-commit cross-chain provenance mismatch, each with a matched-chain control. ABIs are byte-identical (changes are internal logic only); the `.wasm` artifacts are recompiled.

### Follow-up (not in this PR)

The finding also recommends validating envelope semantic fields (message checksum, payload hash, attestation data sizes) before applying consensus — a separate envelope-integrity layer that overlaps WSA-051; worth its own analysis rather than bundling here.


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -722,16 +777,10 @@ void apply_consensus(name self, uint64_t chain_code, uint32_t epoch_index,
    // consumer of inbound message rows ever materialises, the emplace can be reintroduced (and the
    // matching erase left out).
 
-   // Extract + dispatch each attestation in every message of the envelope.
+   // Extract + dispatch each attestation in every message of the envelope. The proven source
+   // outpost is `chain_code` (validated by `deliver`); the per-attestation dispatch binds every
+   // value-bearing payload's own chain identifier to it (WSA-005, `source_chain_binding_ok`).
```
huangminghuang: This says every value-bearing payload is bound to the proven source outpost, but `ATTESTATION_TYPE_NODE_OWNER_REG` still calls `dispatch_node_owner_reg(data)` without passing or checking this `chain_code`. That path is value-bearing: it creates/registers a node owner and records the ETH link, while `roa::nodeownreg` / `authex::recordlink` treat the OPP attestation as the proof. Because `NodeOwnerRegistration` only carries `ChainAddress` fields and no exact chain code, a quorum for another active outpost can still deliver this payload and drive Wire-side node-owner state. Please bind this attestation to the proven source outpost too, either by adding/passing an exact chain-code-bearing claim field or by gating it to the expected ETH outpost before sending `newnameduser` / `nodeownreg`.

heifner: Good catch — you're right this path is value-bearing; I'd mis-filed it as identity-only. Fixed in 6711e3cd42.

`NodeOwnerRegistration` carries no exact chain code (only `ChainAddress` fields), so I took option (b): bind by VM family. `dispatch_node_owner_reg` now resolves the proven source outpost's kind from `sysio.chains` and drops unless it's EVM — the claim is an ERC1155/ETH deposit proven by the depositor's EVM secp256k1 actor key, so a non-EVM (SVM/WIRE) quorum can no longer drive `newnameduser` / `nodeownreg` / `recordlink`.

If multiple EVM outposts ever coexist and registration must pin to one specific chain, that needs option (a) — a chain-code field on the `NodeOwnerRegistration` proto plus the outpost-side populate, a coordinated cross-repo change. Happy to open a follow-up for the exact-chain binding if you'd prefer it now. Added a regression test (non-EVM delivery → no node-owner state).

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -488,7 +550,23 @@ std::optional<sysio::public_key> em_pubkey_from_eth_bytes(const std::vector<char
...
+/// owner, and records an ETH authex link, with the consensus-reached attestation as the sole proof),
+/// but `NodeOwnerRegistration` carries no exact chain code to match against the proven source outpost
+/// -- the claim is an ERC1155 NFT deposit identified only by `ChainAddress` fields. So it is bound by
+/// VM family instead: the registration is an Ethereum-family claim (the depositor's `actor_pub_key`
+/// is an EVM secp256k1 key), and is accepted only when the proven source outpost (`chain_code`) is an
+/// EVM chain. This stops a quorum for a non-EVM (SVM/WIRE) outpost from driving Wire-side node-owner
+/// state off a forged claim. Exact single-chain binding would require a chain-code field on the
+/// claim proto (a coordinated outpost change); the family gate closes the cross-VM-family hole today.
+void dispatch_node_owner_reg(const std::vector<char>& data, uint64_t chain_code) {
+   if (chain_kind_for(chain_code) != ChainKind::CHAIN_KIND_EVM) {
```
huangminghuang: This closes the non-EVM forgery path, but it still binds `NODE_OWNER_REG` only to the VM family rather than the proven source chain. `ChainKind::CHAIN_KIND_EVM` covers multiple chain instances (Ethereum, Polygon, Arbitrum, Base, etc.), while WSA-005 is about binding effects to the exact delivering outpost. Since `NodeOwnerRegistration` still carries no exact `chain_code`, a second active EVM outpost would pass this check and could drive node-owner state for a claim from the wrong EVM chain. Please either add an exact chain-code field and compare it to the proven `chain_code`, or if node-owner registration is Ethereum-mainnet-only, gate on the specific registered ETH outpost code rather than `CHAIN_KIND_EVM`.

heifner: Good point — tightened to bind to the exact Ethereum outpost instead of the EVM family, in 176bd949fe.

`dispatch_node_owner_reg` now drops unless the proven source `chain_code` equals `NODE_OWNER_SRC_CHAIN` (`"ETHEREUM"` — the `ChainSpec.code` the launch and dev bootstrap configs register for the Ethereum outpost; stable across environments, only `external_chain_id` differs: 1 on mainnet vs 31337 on the anvil devnet). It reuses the same `source_chain_binding_ok` exact-match as the other value-bearing paths — the constant just stands in for the chain code the `NodeOwnerRegistration` proto doesn't carry. A second active EVM outpost (Polygon/Base/Arbitrum/…) now mismatches and is dropped, closing the hole the family gate left, and the redundant `chain_kind_for` read of `sysio.chains` is gone.

No OPP proto change: the existing `ChainSpec` already identifies the Ethereum outpost by code, and node-owner NFT registration is Ethereum-mainnet-only, so binding to that single outpost is the intended policy. If multiple Ethereum-family outposts ever need to source registrations, that's the proto chain-code field (the option-(a) follow-up).

Tests (`sysio_dispatch_tests`): the happy path now delivers from the `"ETHEREUM"` source; added `node_owner_reg_from_other_evm_outpost_is_dropped` (a claim proven-delivered from a second active EVM outpost → dropped, with no node-owner state); and `node_owner_reg_from_non_evm_outpost_is_dropped` now delivers from a *scheduled* SVM outpost so the drop is exercised at the source binding (it reaches consensus first) rather than merely the consensus gate.

huangminghuang (approved):
Reviewed the current PR head and found no blocking issues. The exact-source binding for node-owner registration addresses the prior review concern.

PR #443, opened 2026-06-26
sysio.uwrit/opreg: aggregate underwriter collateral per bucket (SEC-15)

heifner:
### Problem
Underwriter collateral is held and rolled up per `(underwriter, chain_code, token_code)` — **not** per `reserve_code`. A same-`(chain, token)` swap between two reserves (distinct `reserve_code`, a shape `rcrdcommit` explicitly routes) therefore draws **both** required legs against the **same** collateral bucket.

`sysio.uwrit::try_select_winner` checked each leg's availability **independently**, so an underwriter whose balance covers each single leg but **not their sum** could win and write two locks that overcommit the bucket — e.g. balance **150** wins `src_amount=100` + `dst_amount=100` because `150 ≥ 100` holds per leg, yet the aggregate **200** is uncovered.

Downstream impact: on a later slash/terminate, `sysio.uwrit::chklocks` calls `opreg::releaselock` **once per lock**, and the second `subtract_balance` underflows the shared balance. `releaselock` runs **inline inside `sysio.epoch::advance`**, so the underflow `check()` aborts the advance and **permanently stalls epoch advancement chain-wide**.

### Fix
- **`uwrit::try_select_winner`** — when both required legs share a `(chain, token)` bucket, require availability to cover the **aggregate** (`src + dst`); saturating add so a near-`UINT64_MAX` total is rejected rather than wrapping to a small passing value. Distinct-bucket and single-leg swaps are unchanged.
- **`uwrit::available_via_mirrors` / `opreg::available_inline`** — saturating `locked + pending` so a wrap can't overstate availability.
- **`opreg::releaselock`** (defence-in-depth) — clamp the deferred slash/remit subtraction to the live balance, so the consensus-inline `chklocks` cleanup path can never underflow + abort even on a residual over-committed lock set.

No ABI change.

### Tests (`contracts_unit_test` → `sysio_dispatch_tests`)
- `swap_same_token_legs_overcommit_is_disqualified` — balance 150, 100+100 same bucket → disqualified, no locks.
- `swap_same_token_legs_exact_balance_wins` — balance 200 → wins, two locks total 200; a fresh same-bucket swap is then disqualified (existing locks reserve the bucket).
- `swap_same_token_legs_aggregate_saturates` — `src=100` + `dst=UINT64_MAX-50` saturates and is rejected (no wrap).
- `releaselock_clamps_overdrain_without_aborting` — over-drain releases clamp to the live balance instead of aborting.

Full `contracts_unit_test` suite: **422/422 pass**.


contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -1034,16 +1037,48 @@ void try_select_winner(name self, uint64_t uwreq_id, name candidate) {
...
+   // aggregate saturates and is rejected, never wraps to a small passing
+   // value; this resolver is non-throwing (`feedback_opp_handlers_never_throw`),
+   // so saturate-then-compare is the correct guard here, not a checked throw.
+   const bool same_bucket = src_needed && dst_needed
+                            && req.src_chain_code == req.dst_chain_code
+                            && req.src_token_code == req.dst_token_code;
+   bool insufficient_bond = false;
+   if (same_bucket) {
+      // One opreg balance funds both legs — it must cover their sum.
+      const uint64_t need  = opp::safe::add_sat_u64(req.src_amount, req.dst_amount);
```
huangminghuang: This still lets an overflowing aggregate pass when the underwriter has `UINT64_MAX` available. Example: `src_amount = UINT64_MAX`, `dst_amount = 1`, and `avail = UINT64_MAX` gives `need = add_sat_u64(...) == UINT64_MAX`, so `avail < need` is false and two locks can be written against one `(chain, token)` bucket whose real requirement exceeds the bucket. Since inbound collateral is stored as uncapped `uint64_t`, please treat `src + dst` overflow itself as insufficient, and also saturate the lock/pending subtotal loops so wrapped active locks cannot later overstate availability.

heifner: Good catch — fixed in 151a326d25.

- The aggregate is now computed in `uint128` and compared against the (`uint64`) available balance, so an overflowing `src+dst` is insufficient by construction. Your exact case (`src=UINT64_MAX`, `dst=1`, `avail=UINT64_MAX`) now disqualifies — the prior `add_sat_u64` capped `need` at `UINT64_MAX` which a max balance spuriously satisfied.
- The lock/pending subtotal loops in both `sysio.uwrit` (`sum_locks_inline`, `opreg_pending_withdraws`) and `sysio.opreg` (`sum_locks_inline`, `sum_pending_withdraws`) now accumulate with `add_sat_u64`, so a wrapped subtotal can't understate `reserved`/overstate availability. (`available_via_mirrors`/`available_inline` already saturate the outer `locked+pending`.)

Added a regression test (`swap_same_token_legs_aggregate_overflow_insufficient`) for the overflow corner. Full `contracts_unit_test` suite passes.

huangminghuang (approved):
Approved. I re-reviewed head 151a326d after the overflow follow-up; no actionable code findings remain. Note: the required check is still red because the GCC job's long-running shard failed in ship_streamer_test, which appears unrelated to this PR's touched surface.

PR #448, opened 2026-06-26
Fix: reject signed TokenAmount values at depot ingress (WSA-028)

heifner:
SEC-14 / WSA-028 (P1). OPP TokenAmount.amount is signed on the wire (int64 / vint64_t), but several value-bearing ingress paths cast it straight to uint64_t, so a negative amount such as -1 wrapped to 18446744073709551615 — an impossible balance that defeats the downstream zero-only guards and inflates operator collateral, reserve settlement, and reward accounting.

This adds a shared fail-closed gate sysio::opp::safe::to_depot_amount, which accepts only amounts in (0, asset::max_amount] and returns nullopt otherwise, and routes every inbound TokenAmount through it. Each call site keeps the never-throw dispatch contract — a rejected amount is dropped, refund-reverted, or clamped, never check()-aborted inside evalcons:

- dispatch_operator_action (DEPOSIT_REQUEST / WITHDRAW_REQUEST) drops the action; an impossible deposit/withdraw has no escrow to refund.
- createuwreq (SWAP_REQUEST source_amount) emits SWAP_REVERT and creates no uwreq.
- STAKING_REWARD drops before crediting the dclaim ledger.
- dispatch_reserve_create folds the existing negative clamp into the gate, so an oversized positive also collapses to 0 (oncrtreserve cancels/refunds) rather than slipping through the prior > 0 clamp.

sysio.dclaim::onreward already rejected > asset::max_amount, so it was the one path already closed; the gate makes the rule uniform across every ingress.

depot_amount_max mirrors asset::max_amount locally so the shared header stays free of <sysio/asset.hpp> for the native-tester host build, with a static_assert keeping the two in lockstep.


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -410,19 +420,18 @@ void dispatch_reserve_create(name self, const std::vector<char>& data, uint64_t
...
 
    // WSA-005: `external_amount.chain_code` is documented in the proto as "the outpost's own chain".
    // Bind it to the proven delivering outpost so an envelope proven from outpost A cannot register a
    // reserve whose external custody is claimed against a different chain B.
    if (!source_chain_binding_ok(chain_code, ext.chain_code, "dispatch_reserve_create")) return;
 
-   const uint64_t ext_amount = ext.amount.amount > 0
-                                  ? static_cast<uint64_t>(ext.amount.amount)
-                                  : 0;
+   const uint64_t ext_amount =
```
huangminghuang, line 435: Invalid RESERVE_CREATE amounts are skipped, not refunded. Here an out-of-range positive external_amount.amount becomes 0 via value_or(0), but sysio.reserv::oncrtreserve returns immediately on external_token_amount == 0 before inserting a CANCELLED row or queueing RESERVE_CREATE_CANCELLED. So 2^62 no longer inflates the reserve, but the creator’s outpost escrow gets no cancel/refund signal. This path should explicitly route invalid reserve-create amounts into the cancel/refund flow, with a regression test that observes the queued cancel.

heifner: Good catch — fixed in d1e1efcd29. `oncrtreserve` now folds the invalid-amount case into the same cancel/refund branch as an unlinked creator: it inserts a CANCELLED row and queues `RESERVE_CREATE_CANCELLED` so the outpost releases the creator's escrow, idempotently (a re-relay over the CANCELLED row doesn't double-refund; a later linked creator with a valid amount still reclaims the identity). Added `oncrtreserve_invalid_amount_is_cancelled` to observe it.

I left the structural-malformation skips (bad `connector_weight_bps` / `source_token_precision`) as silent drops rather than refunds — those are malformed/untrusted attestations rather than a well-formed create with a bad amount. Happy to route those through refund too if you'd prefer.

huangminghuang (approved):
Re-reviewed the follow-up commits. The reserve-create refund path issue is fixed, and I don’t see remaining blocking issues.

PR #449, opened 2026-06-26
Fix: cap operator collateral credit at asset::max_amount (SEC-103, WSA-028 follow-up)

heifner:
## Summary

Follow-up to SEC-14 / WSA-028 (PR #448). That fix gates each **single** inbound `TokenAmount` to `(0, 2^62-1]` at the `sysio.msgch` ingress, closing the signed→unsigned wrap. This PR closes the remaining **accumulation** path it does not cover.

`sysio.opreg::add_balance` credits operator collateral with a raw `b.balance += amount` — the lone amount accumulator that neither saturates nor caps (`sysio.reserv` settlement uses `add_capped_u64`; the opreg availability rollups and `sysio.dclaim` credit use `add_sat_u64`). Repeated legitimate, escrow-backed deposits whose **sum** exceeds `asset::max_amount` (`2^62-1`) push the stored balance out of asset range. The WIRE-direct withdraw/terminate remit path then builds `asset(static_cast<int64_t>(balance), CORE_SYM)`, which `check()`-aborts past the limit — and on the never-throw `depositinle` path that abort would roll back the consensus-tipping delivery and **stall epoch advancement chain-wide**.

Likelihood is low (it needs multiple real escrow-backed deposits near `2^62` each), but it is the one collateral accumulator not covered by the checked-arithmetic remediation WSA-028 calls for.

## Change

Gate both collateral entrypoints on the **running sum**, fail-closed in each path's idiom:

- **`depositinle`** (OPP-inbound, never-throw): refund via `DEPOSIT_REVERT` + append a failed action log + return when `balance + amount` would exceed `asset::max_amount`.
- **`deposit`** (user action): `check()`-reject before the inline WIRE transfer, so no tokens move.

A `balance_of()` helper and a `MAX_COLLATERAL_AMOUNT` constant (`= asset::max_amount`) are added; `add_balance` itself stays uncapped — callers precheck, mirroring the unchecked `subtract_balance`. The overflow check is underflow-safe (`amount > MAX || balance > MAX - amount`).

## Test

`depositinle_credit_over_max_collateral_is_reverted`: credit to **exactly** `asset::max_amount` (accepted — inclusive boundary), then a `+1` credit is refunded via `DEPOSIT_REVERT` with the stored balance unchanged and no throw.


contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -903,6 +929,16 @@ void opreg::deposit(name account, uint64_t amount) {
          op.status != OperatorStatus::OPERATOR_STATUS_TERMINATED,
          "operator not in a deposit-eligible state");
 
+   // Collateral must stay within the asset magnitude range: the WIRE-direct
+   // withdraw / terminate remit path builds `asset(balance, CORE_SYM)`, which
+   // aborts above `asset::max_amount`. Reject before the inline WIRE transfer
+   // moves anything, so the running sum can never exceed the cap. A direct user
+   // deposit may legitimately `check()`-throw here — unlike the never-throw OPP
+   // `depositinle` — and the whole transaction reverts, so no tokens move (SEC-103).
+   check(amount <= MAX_COLLATERAL_AMOUNT &&
```
huangminghuang: [P1] This still leaves a cap bypass for direct WIRE deposits if the operator account has contract code. `sysio.token::transfer` notifies both `from` and `to` before `opreg` credits the balance, so during the `from` notification an operator contract can inline-call `opreg::deposit` again with the same auth while the outer call has not yet mutated the balance. Both deposits pass this precheck against the same old balance, then both `add_balance` calls run and can push the WIRE row over `asset::max_amount`, recreating the withdraw/terminate abort risk this PR is closing.

Please make the cap check atomic with the mutation, for example by re-checking inside the `ops.modify` path before adding, or credit before the token transfer and rely on transaction rollback if the transfer fails. A regression test with an operator notification handler reentering `deposit` would lock this down.

heifner: Addressed in ba34423 (fix) + 801046d (test).

**Fix:** moved the cap check **inside** the `ops.modify` (reading the live row) and credit **before** the transfer, so check+credit is one indivisible step and the balance commits before any `from` notification fires. A re-entrant `deposit` therefore observes the updated balance and its own check accounts for it — two credits can no longer pass against the same stale balance. If the transfer aborts (operator lacks the WIRE) the whole transaction, credit included, rolls back.

On the ordering specifically: the `transfer` is a deferred inline action, so it (and its `from`/`to` notifications) run only after `deposit` returns — i.e. after `add_balance` had already committed — so even pre-fix the re-entry observed the updated balance. The in-modify check makes it correct by construction regardless.

**Test:** added `deposit_reentrancy_cannot_exceed_max_collateral` plus a `reenter_deposit` test contract (`contracts/test_contracts/reenter_deposit`) that re-enters `deposit(+1)` on the operator's outgoing transfer notification, exactly the vector you described. It deploys the SYS core token, funds the operator with exactly `asset::max_amount`, deposits the full cap, and asserts the `+1` re-entry trips the in-modify cap check and aborts the whole transaction with no over-credit (the operator's balance row stays empty). Green.

huangminghuang (approved):
Approved after re-review. The direct-deposit cap check is now atomic with the credit, and the reentrancy regression test covers the token-transfer notification path raised earlier.

PR #356, opened 2026-05-26
Add macOS arm64 developer support

huangminghuang:
## Summary

Adds macOS Apple Silicon support for Wire Sysio, including the Release CI path, Apple arm64 dependency registry updates, macOS-safe CMake/runtime gates, native-debug host integration, and `sys-vm-jit` support on macOS arm64 through `wire-sys-vm` 1.2.0.

## What changed

- Adds a `macos-26` Apple Silicon workflow (`macos_arm64_build.yaml`) that configures a Release build, builds `all`, runs the parallelizable CTest set for PRs, pushes to `master`, and a nightly scheduled run on `origin/master`, and runs the sharded NP/LR sweep only on nightly schedule or explicit manual opt-in.
- Shares `.github/scripts/build-sysio.sh` between Linux and macOS CI by making the build directory, vcpkg triplets, compiler settings, and cache paths environment-driven while keeping Linux workspace ownership repair in the Linux workflow.
- Adds an `arm64-osx-release` vcpkg overlay triplet matching the Linux release-triplet pattern, including release-only vcpkg dependencies and ccache launchers.
- Pins `wire-vcpkg-registry` to a `master` commit containing the macOS-compatible Boost, SoftFloat, secp256k1, and `wire-sys-vm` 1.2.0 port updates, with `wire-sys-vm` pinned to eos-vm master including Wire-Network/eos-vm#13.
- Enables `sys-vm-jit` by default for Apple arm64 builds while keeping x86-only profiling JIT support gated to x86_64.
- Gates Linux/x86-only behavior away from macOS, including linker flags, LLVM/OC assumptions, allocator defaults, and unsupported runtime paths.
- Defaults jemalloc off on macOS while preserving Linux defaults and sanitizer auto-disable behavior.
- Updates SoftFloat type handling and related serialization/precondition paths for Apple arm64 compatibility.
- Fixes AppleClang/libc++ gaps that remain on macOS 26, including missing `std::views::enumerate`, missing `std::move_only_function`, `std::chrono::utc_clock` usage, and `time_t`/`fc::variant` integer typing.
- Skips sys-vm-oc-only chain plugin tests when OC is not enabled.
- Enables macOS Debug native-module host support by allowing `native-module` on Apple Debug builds and generating Apple `-exported_symbols_list` files for native intrinsic exports.
- Documents macOS developer setup and native-debug behavior in `README.md`, `AGENTS.md`, `CLAUDE.md`, and `contracts/native-debug.md`.

## Why

[...]

PR #450, opened 2026-06-26
Fix interrupted transaction recovery catchup

huangminghuang:
## Summary

Fixes the interrupted transaction recovery path that could leave a restarted producer stuck in LIB catchup after it had already reached the peer's known fork DB root.

## Failure

Failure CI run: https://github.com/Wire-Network/wire-sysio/actions/runs/28244239953/job/83678838341

The failing job timed out in `interrupt_trx_test.py` after the producer was relaunched. During recovery, the restarted node could receive duplicate or already-present blocks at/above the peer's known fork DB root. Because those blocks were not newly applied, the old sync path did not emit the accepted-block transition that normally advances LIB catchup to head catchup. The node stayed in `lib_catchup`, kept handshaking/waiting, and the test never reached the recovered producer round.

## Fix

- Add a `sync_manager` helper that completes LIB catchup when the local fork DB root already satisfies the known peer root.
- Cancel any outstanding sync wait before clearing the sync source so an old timer cannot later close a healthy peer.
- When this recovery path promotes to head catchup, request from the fork DB root instead of reusing a fork DB head id from the stale/poisoned transition.
- Assert the post-relaunch producer waits in `interrupt_trx_test.py` so the recovery failure is reported directly instead of silently hanging until the outer CI timeout.

## Branch Base

This branch was rewritten to include only the delta from PR #356 to this fix, then replayed onto current `origin/master`. The PR now contains one commit, `5a4d34ecf1`, and only these files:

- `plugins/net_plugin/src/net_plugin.cpp`
- `tests/interrupt_trx_test.py`

## Review

A subagent review found the missing timer cancellation in the first version of the patch. That was fixed, then re-reviewed with no remaining findings.

## Validation

- `git diff --check origin/master...HEAD`
- `cmake --build build/codex-interrupt-trx --target net_plugin nodeop clio sys-util kiod sysio_testing_contracts_copy -- -j6`
- `timeout 600 python3 tests/interrupt_trx_test.py -v`

Note: local validation used `-DBUILD_OPP_BUNDLES=OFF` because `wire-protobuf-bundler` is not installed on this machine.

### DIFF
--- a/plugins/net_plugin/src/net_plugin.cpp
+++ b/plugins/net_plugin/src/net_plugin.cpp
@@ -185,6 +185,8 @@ namespace sysio {
       connection_ptr find_next_sync_node(); // call with locked mutex
       void start_sync( const connection_ptr& c, uint32_t target ); // locks mutex
       bool sync_recently_active() const;
+      /// Complete LIB catchup when the local fork DB root has already reached the known peer root.
+      bool complete_lib_catchup_if_root_reached( uint32_t fork_db_root_num ) REQUIRES(sync_mtx);
       bool verify_catchup( const connection_ptr& c, uint32_t num, const block_id_type& id ); // locks mutex
    public:
       enum class closing_mode {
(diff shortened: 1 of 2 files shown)


heifner (changes requested):
Targeting the wrong branch or needs to be rebased.

heifner (approved):

PR #451, opened 2026-06-26
Uwrit: require ACTIVE underwriter to win the settlement race (SEC-88)

heifner:
## Problem

`sysio.uwrit::try_select_winner` chose the settlement-race winner after verifying only the candidate's underwrite-intent-commit (UIC) signature and its mirrored per-leg bond. It never checked that the candidate's `sysio.opreg` row is `OPERATOR_TYPE_UNDERWRITER` with `OPERATOR_STATUS_ACTIVE`. The balance mirror (`available_via_mirrors`) zeroes only `SLASHED`/`TERMINATED` and ignores `op.type`, so a non-underwriter (producer / batch / challenger), or an underwriter still `UNKNOWN` because it had not cleared its activation threshold, could win the race with enough mirrored balance and a valid signature -- consuming lock capacity and driving reserve/WIRE settlement while bypassing the role and activation gate.

Finding WSA-199 / OBS-0299.

## Fix

- `is_active_underwriter(candidate)` reads the opreg `operators` row and requires `type == UNDERWRITER && status == ACTIVE`.
- `try_select_winner` gates on it before any signature work, lock, `CONFIRMED` write, or reserve mutation. Ineligible candidates are disqualified (reclaimable, non-throwing -- the resolver runs inside the evalcons dispatch chain and must not stall consensus). This is the single settlement chokepoint; the swap-from-WIRE drain path routes through it too.

## Tests

- The dispatch fixture registered the underwriter but left `req_uw_collat` empty, so it stayed `UNKNOWN` and won regardless -- the bug. A one-unit ETH minimum now promotes it to `ACTIVE` via `processuw` on its existing ETH deposit (adds no balance row, so deposit-routing assertions are unaffected).
- Two regression tests: a funded producer and a funded-but-inactive underwriter, each with a valid self-signed UIC, both disqualified with the race left `PENDING` and no lock written. The existing `swap_same_token_legs_exact_balance_wins` is the positive control.
- The aggregate-overflow test now deposits the collateral cap (its `UINT64_MAX` deposit is refused by the cap, so it never activated).

Recompiled `sysio.uwrit.wasm`; no genesis/snapshot reference data affected.

## Stacking

Stacked on #449 (`fix/sec-103-opreg-collateral-cap`) -- the aggregate-overflow test relies on that branch's collateral cap. Re-target `master` once #449 merges.


contracts/tests/sysio.dispatch_tests.cpp
```diff
@@ -1304,6 +1334,132 @@ BOOST_FIXTURE_TEST_CASE(swap_candidate_with_invalid_uic_signature_is_disqualifie
...
+// control — an ACTIVE underwriter that wins — is `swap_same_token_legs_exact_balance_wins`.
+
+// A non-underwriter operator (PRODUCER), registered and funded with ETH bond,
+// cannot win an underwriting race even with a valid self-signed UIC.
+BOOST_FIXTURE_TEST_CASE(swap_winner_non_underwriter_type_is_disqualified,
+                        sysio_dispatch_tester) { try {
+   bootstrap_for_dispatch();          // ETH source outpost + UWRIT_OP
+   register_wire_depot();             // to-WIRE: a single (source) required leg
+
+   // A second operator registered as a PRODUCER (NOT an underwriter). Privileged
```
huangminghuang, line 1367: This test does not actually isolate the `op.type == UNDERWRITER` half of the new gate. `PRODOP` is registered as a non-bootstrapped producer while `req_prod_collat` stays empty, so it remains `OPERATOR_STATUS_UNKNOWN`; a status-only production check would still disqualify it and this regression would pass. Please make the candidate an ACTIVE non-underwriter (for example by configuring/funding producer collateral, or using a bootstrapped active non-underwriter where valid) and assert `OPERATOR_STATUS_ACTIVE` before `rcrdcommit`, so the test fails if the type check is removed.

heifner: Good catch — addressed. The producer is now promoted to ACTIVE (configured `req_prod_collat` + funded, so `processprod` activates it), and the test asserts `OPERATOR_STATUS_ACTIVE` before `rcrdcommit`. Its only disqualifying attribute is now `op.type`, so the test fails if the type check is removed. Generalized the setconfig helper to take `req_prod_collat`.

huangminghuang (approved):
Approved. The code review looks good at `2f9d76a`, and the active-producer regression now isolates the `op.type` gate correctly.

Non-blocking cleanup: the PR description is stale now that #449 has merged and this PR is based on `master`. The “Stacking” section still says to re-target to `master`, and the tests section still uses the old aggregate-overflow wording even though the current test is `swap_oversized_source_reverts_at_ingress`.

PR #452, opened 2026-06-27
Net: validate connect() peer before retaining in supplied_peers

heifner:
The API connect handler inserted the caller-supplied endpoint into supplied_peers before any validation. resolve_and_connect then rejected unparseable addresses with "invalid peer address", but the entry already lived in supplied_peers permanently; only an exact-match /v1/net/disconnect erases it, and connection cleanup never does. Every later reconnect scan and per-accept inbound check (any_of_supplied_peers) then iterated those dead entries.

Validate the endpoint with the same split_host_port_type gate resolve_and_connect already uses, before inserting, so an address the node refuses is no longer retained. The return value for invalid input is unchanged ("invalid peer address"); the only behavioral difference is that the rejected address is not added to supplied_peers.

Related to SEC-46

### DIFF
--- a/plugins/net_plugin/src/net_plugin.cpp
+++ b/plugins/net_plugin/src/net_plugin.cpp
@@ -375,6 +375,10 @@ namespace sysio {
                             timer_type which);
 
       void add(connection_ptr c);
+      /// Validate and retain an operator-supplied peer, then attempt to connect. Unparseable endpoints are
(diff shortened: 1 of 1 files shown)


huangminghuang (approved):
Reviewed the single net_plugin change and found no blocking issues. The new validation mirrors the existing resolve_and_connect invalid-address gate and prevents rejected API-supplied peers from being retained in supplied_peers.

PR #453, opened 2026-06-27
Trace_api: bound action-range response cardinality

heifner:
## Summary

`/v1/trace_api/get_actions` and `/v1/trace_api/get_token_transfers` clamp the block window they scan (`trace-max-block-range`, default 1000 / max 10000) but do not bound the number of matching actions returned from that window. `get_actions_impl()` appends every match -- ABI-decoded and hex-serialized -- to the response array with no ceiling, so a single small request over a transfer-heavy or unfiltered range can make the node materialize a very large `actions`/`transfers` array (memory + HTTP-worker CPU), degrading the trace/debug service.

## Change

- Add `max_actions_per_response`, a **hard-coded** soft ceiling on the actions materialized per response. It is deliberately a compile-time constant and **not** a nodeop configuration option -- there is no setting to raise or disable it, so the bound holds on every node regardless of operator configuration.
- The ceiling is enforced at block boundaries: the scan stops once a fully scanned block brings the running total to it. A block is never returned partially, so no action within the reported range is omitted; worst-case overshoot is one block's matches (itself bounded by consensus block limits).
- `actions_result` carries `last_block_num` (the highest block fully scanned). Both handlers report it as `block_num_end`, so clients resume at `block_num_end + 1` with no gap or duplicate -- the same block-range pagination the block-window clamp already relies on.
- Unit tests cover truncation at a block boundary, single-block-over-ceiling forward progress, gap/duplicate-free resume, and get_token_transfers parity.

## Notes

- No wall-clock deadline is used: the handlers run on the HTTP thread pool and never block the chain thread, and the block-range cap together with the action ceiling already bounds blocks scanned, actions decoded, and bytes serialized.
- Response shape is unchanged; `block_num_end` differs from prior behavior only when the ceiling truncates a scan.

Jira: https://wire-network.atlassian.net/browse/SEC-64 (WSA-145)

### DIFF
--- a/plugins/trace_api_plugin/include/sysio/trace_api/request_handler.hpp
+++ b/plugins/trace_api_plugin/include/sysio/trace_api/request_handler.hpp
@@ -68,10 +68,31 @@ namespace sysio::trace_api {
    };
 
    /**
+    * Soft ceiling on the number of action results materialized into a single get_actions /
+    * get_token_transfers response.  Complements the per-request block-window cap
+    * (trace-max-block-range): the window bounds how many blocks are scanned, this bounds how many
+    * matching actions are decoded, hex-serialized, and held in memory from that window, so a broad
+    * or unfiltered query over a busy range cannot materialize an unbounded response set.
+    *
+    * Enforced at block boundaries (see get_actions_impl): the scan stops once a fully scanned block
+    * brings the running total to this value, so a single response may exceed it by at most one
+    * block's worth of matching actions (itself bounded by consensus block limits).  Clients page
+    * past the ceiling by resuming the scan at actions_result::last_block_num + 1.
+    *
+    * Intentionally a hard-coded constant, NOT a nodeop configuration option: there is no setting to
+    * raise or disable it, so the bound holds on every node regardless of operator configuration.
+    */
+   inline constexpr uint32_t max_actions_per_response = 10'000;
+
+   /**
     * Result returned by get_actions.
     */
    struct actions_result {
       fc::variants actions;
+      /// Highest block number fully scanned by the request.  Equals the clamped block_num_end on a
+      /// complete scan; lower when max_actions_per_response stopped the scan early.  Clients resume
+      /// at last_block_num + 1.
+      uint32_t     last_block_num = 0;
    };
 
    /**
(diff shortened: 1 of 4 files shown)


plugins/trace_api_plugin/include/sysio/trace_api/request_handler.hpp
```diff
@@ -67,11 +67,32 @@ namespace sysio::trace_api {
...
+    *
+    * Enforced at block boundaries (see get_actions_impl): the scan stops once a fully scanned block
+    * brings the running total to this value, so a single response may exceed it by at most one
+    * block's worth of matching actions (itself bounded by consensus block limits).  Clients page
+    * past the ceiling by resuming the scan at actions_result::last_block_num + 1.
+    *
+    * Intentionally a hard-coded constant, NOT a nodeop configuration option: there is no setting to
+    * raise or disable it, so the bound holds on every node regardless of operator configuration.
+    */
+   inline constexpr uint32_t max_actions_per_response = 10'000;
```
huangminghuang, line 85: This introduces a public pagination behavior change, but `plugins/trace_api_plugin/trace_api_plugin.md` still says “ALL matching actions are returned” and that there is “no per-result limit.” Please update the pagination guide / response field docs for both `get_actions` and `get_token_transfers` to describe `max_actions_per_response`, block-boundary truncation, and resuming from the returned `block_num_end + 1`.

heifner: Done in 9065da05cb. Updated `trace_api_plugin.md`: the Pagination guide now describes the `max_actions_per_response` (10,000) ceiling, block-boundary truncation (a response overshoots by at most one block's worth and is never split mid-block), and resuming at `block_num_end + 1` — replacing the stale "ALL matching actions / no per-result limit / no per-response row cap" claims. The `get_actions` response-field table and the `get_token_transfers` section now both note the ceiling and the cursor.

Also corrected the `trace-max-block-range` `--help` text, which still advised advancing `block_num_start` by the full window — now unsafe once a response truncates early; it points to `block_num_end + 1` like the doc.

huangminghuang (approved):
Re-reviewed current head c37da9f. The earlier documentation gap is addressed: the public docs and help text now describe max_actions_per_response, block-boundary truncation, and resuming from returned block_num_end + 1. No remaining findings.

PR #454, opened 2026-06-27
Http: count request bodies against the in-flight byte budget

heifner:
App-thread HTTP handlers checked `verify_max_bytes_in_flight(b.size())` but never reserved the
request body before moving it into the unbounded app-thread exec queue. `bytes_in_flight` only ever
tracked response and file payloads, so concurrent request bodies accumulated uncounted and the
`http-max-bytes-in-flight` budget did not bound them — a set of near-`max-body-size` POSTs against a
slow handler could pin request-body memory well past the configured budget. The factory documented
`@pre b.size() has been added to bytes_in_flight by caller`, a precondition no caller satisfied.

This reserves the body against `bytes_in_flight` on admission and releases it exactly once via a
`shared_ptr`-held `scoped_exit` captured into the posted lambda, so the release fires whether the work
runs, throws, returns early on shutdown, or is discarded un-run when the queue is cleared. The guard
is `shared_ptr`-wrapped to keep the posted lambda copyable, as `boost::asio::post` requires.
`increment_bytes_in_flight` / `decrement_bytes_in_flight` are promoted onto the `abstract_conn`
interface (sole implementer `beast_http_session`) so the static handler factory can reach them.

The inline http-thread handler path runs synchronously with no queueing, so it has no accumulation
window and is left unchanged; its stale precondition comment is corrected.

A regression test registers an app-thread handler that samples `bytes_in_flight` while the request
body is in flight, asserting the budget reflects the body and returns to zero afterward. It fails on
the prior code (request body uncounted) and passes with the fix.

`bytes_in_flight` is a process-local atomic for HTTP admission control, never part of consensus or
serialization.

Resolves SEC-82.

### DIFF
--- a/plugins/http_plugin/src/http_plugin.cpp
+++ b/plugins/http_plugin/src/http_plugin.cpp
@@ -7,9 +7,11 @@
 #include <fc/log/logger_config.hpp>
 #include <fc/reflect/variant.hpp>
 #include <fc/network/listener.hpp>
+#include <fc/scoped_exit.hpp>
 
 #include <boost/asio.hpp>
 
+#include <functional>
 #include <memory>
 #include <regex>
 
@@ -132,7 +134,8 @@ namespace sysio {
           * Make an internal_url_handler that will run the url_handler on the app() thread and then
           * return to the http thread pool for response processing
           *
-          * @pre b.size() has been added to bytes_in_flight by caller
+          * The returned handler reserves the request body against bytes_in_flight on admission and releases
+          * it (via a shared scoped_exit) when the posted app-thread work completes or is discarded.
           * @param priority - priority to post to the app thread at
           * @param to_queue - execution queue to post to
           * @param next - the next handler for responses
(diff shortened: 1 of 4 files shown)


plugins/http_plugin/src/http_plugin.cpp
```diff
@@ -148,17 +151,30 @@ namespace sysio {
...
+               // The request body is about to be moved into the (unbounded) app-thread queue, where it can
+               // sit until a possibly-slow handler runs. Reserve its bytes against bytes_in_flight now that
+               // admission passed so that the memory is visible to concurrent verify_max_bytes_in_flight()
+               // checks instead of accumulating uncounted. The reservation is released exactly once when the
+               // posted work below is destroyed -- whether it runs to completion, throws, returns early on
+               // shutdown, or is discarded unrun when the queue is cleared. The guard is held through a
+               // shared_ptr so the posted lambda stays copyable (as boost::asio::post requires) and so the
+               // release fires only when the final copy of the lambda is destroyed.
+               const size_t body_size = b.size();
+               conn->increment_bytes_in_flight(body_size);
```
huangminghuang: `increment_bytes_in_flight(body_size)` runs before the `shared_ptr<scoped_exit>` release guard is allocated. If that allocation throws, the outer handler will convert the failure into an HTTP error, but this reservation is never released, permanently inflating `bytes_in_flight`. Since this path is admission control under memory pressure, please create/arm the RAII guard before reserving bytes, or otherwise ensure there is no throwing work between the increment and the guard that guarantees the decrement.

heifner: Good catch. Reworked it in 50815a8717: the increment now lives in a small `bytes_in_flight_reservation` RAII type (ctor reserves, dtor releases), held via `make_shared`. `make_shared` allocates *before* it constructs, so if the allocation throws (`bad_alloc`) the constructor — and therefore the increment — never runs, and a successful construction guarantees the matching decrement exactly once. As a bonus it drops the `std::function` and one allocation per app-thread request.

huangminghuang (approved):
Approved. Re-reviewed at `50815a8`; the request-body accounting fix looks good, and the follow-up RAII reservation closes the allocation-after-increment leak noted in review.

PR #455, opened 2026-06-29
Msgch: gate OPP delivery on current operator status

heifner:
The depot's inbound delivery gate `is_batch_operator_active()` checks only the caller's signature and membership in the current resident batch-operator group. That group is a snapshot taken at schedule time (`advance` / `schbatchgps`), so an operator that loses ACTIVE status mid-window -- slashed, terminated, or dropped below its collateral minimum to UNKNOWN (e.g. after a withdrawal flush) -- remains in `batch_op_groups[current]` until the next reschedule and could still `deliver()` and be counted toward OPP consensus by `evalcons`.

`deliver()` now also reads the operator's `sysio.opreg` row and requires `status == OPERATOR_STATUS_ACTIVE`, failing closed on a missing row. The check is on the operator's own delivery transaction -- not the `evalcons` -> `apply_consensus` -> `advance` inline chain -- so it rejects only that delivery and never stalls epoch advancement. `sysio.opreg` is added to msgch's include path; `is_batch_operator_active` is documented as group-membership-only. The msgch wasm is recompiled; the ABI is unchanged.

Regression tests (`sysio_dispute_tests`): an ACTIVE group member delivers successfully; slashed, terminated, and de-collateralized (UNKNOWN) members are rejected while still group-resident.

Addresses SEC-8 / WSA-017.


huangminghuang (approved):
LGTM. No blocking findings from code review.

PR #456, opened 2026-06-29
Opp: refresh STAKING_REWARD routing docs and drop dead reserv::onreward

heifner:
## Summary

The v6 inbound STAKING_REWARD path credits `sysio.dclaim` directly: the per-staker reward arrives already WIRE-denominated, so `sysio.msgch` forwards it to `sysio.dclaim::onreward` with no re-split. The pre-v6 dual-route that also credited the outpost-side reserve was dropped from the dispatcher, but several comments still described that two-way routing and still named the pre-rename `sysio.cap::onreward`. This corrects the stale documentation and removes the now-orphaned reserve leg.

## Documentation

- `attestations.proto` and `opp_table_types.hpp`: describe the single `sysio.dclaim::onreward` dispatch instead of the obsolete two-way reserve and claim-ledger route, and drop the `sysio.cap::onreward` name.
- `OVERVIEW.md`: the Staking row listed removed/renamed attestation types (`PRETOKEN_STAKE_CHANGE`, `NATIVE_YIELD_REWARD`), replaced with the live `STAKING_REWARD`; `BAR.sol`'s operator-collateral `OPERATOR_ACTION` emission has moved to `OperatorRegistry.sol`.

## Dead code

`sysio.reserv::onreward` has no remaining caller: `sysio.msgch` routes `STAKING_REWARD` only to `sysio.dclaim`, and no e2e flow or outpost contract invokes it. Removed the action and its three unit tests, regenerated `sysio.reserv.wasm` / `.abi`, and left a removal breadcrumb matching the existing `onreject` note.


huangminghuang (commented):
Two non-blocking documentation consistency notes:

- `libraries/opp/proto/sysio/opp/attestations/attestations.proto`: the `StakingReward` block now says the amount arrives already WIRE-denominated, but the `reward_amount` field comment still says "native-denominated". Since this proto comment feeds downstream model/docs, please update it so the denomination story is consistent.

- `contracts/sysio.msgch/src/sysio.msgch.cpp`: the dispatch header comment still says `STAKING_REWARD` routes to `sysio.reserve` / Task 5. The actual dispatch routes to `sysio.dclaim::onreward`, and this PR removes the reserve-side reward leg, so that comment should be refreshed too.

huangminghuang (approved):
LGTM. The follow-up addresses the documentation consistency notes, and CI is green.

PR #459, opened 2026-06-30
Net: Confine P2P frame parsers to the declared message length

heifner:
The native P2P read loop buffers at least one declared frame before dispatching a parser, but the buffer may already hold the next pipelined frame. The generic `net_message`, signed-block, and transaction-notice parsers read from an `mb_datastream` over the entire pending buffer with no per-frame bound, so a peer sending an under-length frame could have missing fields satisfied from the following frame's bytes — desynchronizing the stream and letting a handler act on bytes outside the declared payload (a wrong block id from a short signed-block header, a notice id sourced from the next frame, sync/nack state from a truncated generic message). The transaction and vote paths already rejected frames shorter than their fixed header but still unpacked their variable bodies unbounded.

This adds `fc::bounded_datastream`, a datastream adapter that throws as soon as a read or skip would cross a byte limit, and routes every frame parse through it bounded to the frame's declared length: the routing-tag peek, the generic `net_message`, the signed-block id peek and full unpack, the transaction-notice, and the transaction/vote bodies. After parsing, `connection::advance_to_frame_end` advances the read pointer to the exact end of the frame, skipping any declared bytes the parser did not consume so the next pipelined frame stays aligned.

The signed-block path captures `packed_block` for relay via `datastream_mirror::extract_mirror()`; the bounded stream is nested inside the mirror so that capture is preserved while reads can no longer spill past the frame. New `bounded_datastream` unit tests cover the in/out-of-bound reads, the cross-frame overread, an intact trailing frame, and the mirror-over-bounded composition the block path relies on.

### DIFF
--- a/plugins/net_plugin/src/net_plugin.cpp
+++ b/plugins/net_plugin/src/net_plugin.cpp
@@ -809,6 +809,15 @@ namespace sysio {
       bool process_next_trx_message(uint32_t message_length);
       bool process_next_trx_notice_message(uint32_t message_length);
       bool process_next_vote_message(uint32_t message_length);
+
+      /// Advance the read pointer to the exact end of the frame currently being parsed.
+      /// @param bytes_before pending_message_buffer.bytes_to_read() captured at the frame body start
+      /// @param message_length the peer-declared payload length of the frame
+      /// Skips any declared payload bytes the parser did not consume so the next pipelined frame
+      /// stays aligned. Parsers must bound their reads to message_length (see fc::bounded_datastream),
+      /// so the consumed count can never exceed message_length.
+      void advance_to_frame_end(uint32_t bytes_before, uint32_t message_length);
+
       void update_endpoints(const tcp::endpoint& endpoint = tcp::endpoint());
 
       void send_gossip_bp_peers_initial_message();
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/src/net_plugin.cpp
```diff
@@ -3294,7 +3329,11 @@ namespace sysio {
       auto now = fc::time_point::now();
       // shared_ptr<packed_transaction> needed here because packed_transaction_ptr is shared_ptr<const packed_transaction>
       std::shared_ptr<packed_transaction> ptr = std::make_shared<packed_transaction>();
-      fc::raw::unpack( ds, *ptr );
+      // Confine the body to the remaining declared frame bytes, then advance to the frame boundary,
+      // so the transaction cannot be unpacked from bytes belonging to a following pipelined frame.
+      fc::bounded_datastream body_ds( ds, message_length - header_bytes );
```
huangminghuang: The transaction/vote header read above is still using the unbounded mb_datastream before this bounded body stream is created. A frame declared shorter than which+trx_id/vote_id can still consume bytes from the next pipelined frame, then close() is posted asynchronously and this path returns true, so the read loop can continue on a desynchronized buffer before _close runs. Please bound the fixed-header parse to message_length as well, or stop the read loop synchronously on the short-header path.

heifner: Good catch. Both the transaction and vote paths now wrap the whole parse — `which` + id and the body — in a `bounded_datastream` sized to `message_length`, instead of reading the fixed header through the unbounded `mb_datastream`. A frame too short to hold `which` + id throws during the header read, before any pipelined byte is consumed; `process_next_message` catches it, closes the peer, and returns false, so the read loop stops synchronously rather than continuing on a desynchronized buffer. The explicit `header_bytes > message_length` check is gone since the bound enforces it.

Addressed in ac1de6ceec.

huangminghuang (approved):
Looks good. The follow-up commit bounds the transaction/vote fixed-header parse before reading IDs, so the prior cross-frame overread/desync concern is resolved.

PR #460, opened 2026-06-30
SEC-94: Bound Solana terminal transaction budget

huangminghuang:
## Summary
- bound Solana-bound `sysio.msgch::buildenv` packing with a terminal transaction estimator so READY spillover happens before the Solana relay builds an unrepresentable final transaction
- add legacy Solana transaction packet/key/index validation in libfc
- update the Solana outpost relay path for zero-data terminal finalization and remaining-account extraction from pinned reserve metadata
- add a shared SEC-94 Solana terminal-budget fixture and C++ parity coverage so the sysio estimator is checked against the Solana terminal transaction layout
- restore the Solana data chunk size to 672 bytes now data uploads and terminal finalization are separate transactions, with fixture parity guarding future layout drift
- merge current `origin/master` into the PR branch so the SEC-94 branch includes latest master updates before final e2e
- address review hardening: distinguish absent reserve accounts from RPC/decode failures, require registered SVM chain metadata before queueing, fixture-measure full data chunks, soft-skip unregistered reserve cancellation destinations before `queueout`, and regenerate committed `sysio.msgch`/`sysio.reserv` WASM artifacts

## Companion PR
- Wire Solana: https://github.com/Wire-Network/wire-solana/pull/296 at `f9bf1b2ec`

## Focused local validation
- `contracts_project`
- `test_fc --run_test=solana_client_tests`
- `test_outpost_solana_client_plugin --run_test=outpost_solana_client_plugin`
- `contracts_unit_test --run_test='sysio_msgch_tests,sysio_msgch_envlog_tests' -- --sys-vm`
- `contracts_unit_test --run_test=sysio_reserve_tests -- --sys-vm`
- `unit_test --run_test=deep_mind_tests` across `sys-vm`, `sys-vm-jit`, and `sys-vm-oc`
- `unit_test -t savanna_misc_tests/verify_block_compatibitity` across `sys-vm`, `sys-vm-jit`, and `sys-vm-oc`
- Review follow-up checks: SEC-94 Solana budget fixture parity, Solana chunk tests, sysio outpost Solana plugin test, and `git diff --check`

## Gates
- Latest pushed sysio head: `c69ca846f5`
- GitHub PR CI: passed for latest pushed head
- Wire platform `E2E Flow Tests`: passed at https://github.com/Wire-Network/wire-platform-build-system/actions/runs/28814072876 with `BRANCH_WIRE_SYSIO=fix/sec-94`, `BRANCH_WIRE_SOLANA=fix/sec-94-solana-terminal-finalization`, and `WIRE_BUILD_TYPE=Release`
- Companion Wire Solana PR CI: no checks are reported for the `next`-targeted PR because `wire-solana` PR unit tests currently auto-run only for pull requests targeting `develop`; covered by the shared e2e run above

## Follow-Ups
- Optional relay parsing/batching/walker cleanup is tracked in [SEC-104](https://wire-network.atlassian.net/browse/SEC-104)
- Optional Solana terminal-finalization CU/structure cleanup is tracked in [SEC-105](https://wire-network.atlassian.net/browse/SEC-105)

[SEC-104]: https://wire-network.atlassian.net/browse/SEC-104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


heifner (commented):
Reviewed the full stack against the wire-solana companion (#296): the msgch estimator and budget header, the queueout/buildenv gates, the libfc legacy-transaction validation, the relay's terminal-manifest construction, and the shared fixture. The core SEC-94 mechanism is sound, and I verified the load-bearing invariants rather than assuming them:

- Every attestation type the depot contracts actually queue outbound has an estimator entry (epoch: OPERATORS / BATCH_OPERATOR_GROUPS / EMISSIONS_BLOCKED; opreg: OPERATOR_ACTION SLASH + WITHDRAW_REMIT, DEPOSIT_REVERT; uwrit: SWAP_REMIT / SWAP_REVERT; reserv: RESERVE_READY / RESERVE_CREATE_CANCELLED), and the per-type estimates match the relay's real manifest contribution exactly (the fixture's 5/8 dynamic-account counts reproduce from the relay code).
- Dropping the custody mint from the SPL remit manifest is correct: the on-chain handler uses the pinned `custody_mint` pubkey only to derive the recipient ATA address and pays with a plain `token::transfer`, so no mint `AccountInfo` is needed. The old relay passed it because of a stale on-chain doc comment that the #296 pair fixes.
- The queueout enqueue gate rejects nothing any current caller can queue; it is pure fail-fast against a future uncovered outbound type.
- Reference data on this head is byte-identical to master, so the WASM change doesn't perturb the deep-mind / consensus roots.

Two findings I'd fix before merge (inline): the relay swallowing reserve-lookup failures while still submitting the terminal call, and the ~2-byte data-chunk packet margin under the new serialize assert. The rest is hardening and cleanup discussion. Nice work on the estimator extraction, the measurement fixture with C++ parity coverage, and the uint8_t index hardening in libfc.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -255,63 +388,47 @@ sysio::opp::types::ChainKind outpost_solana_client::chain_kind() const {
...
+      FC_ASSERT(reserve.contains("creator"), "Reserve account missing creator field");
+      FC_ASSERT(reserve.contains("custody_mint"), "Reserve account missing custody_mint field");
+      FC_ASSERT(reserve.contains("custody_decimals"), "Reserve account missing custody_decimals field");
+
+      return reserve_terminal_info{
+         fc::network::solana::solana_public_key::from_base58_string(reserve["creator"].as_string()),
+         fc::network::solana::solana_public_key::from_base58_string(reserve["custody_mint"].as_string()),
+         static_cast<uint8_t>(reserve["custody_decimals"].as_uint64())
+      };
+   } catch (const fc::exception& e) {
```
heifner: This catch-all conflates "reserve genuinely absent" with "RPC transport failure", and `deliver_outbound_envelope` still submits the terminal call with the stripped manifest. On the consensus-tipping operator that turns a transient RPC blip into permanently skipped SPL remit payouts, SPL revert refunds, and reserve-cancel refunds: effects run exactly once in the tipping call, `next_epoch_index` advances in the same transaction, later terminal calls are explicit no-ops, and no instruction can replay a finalized epoch.

The layering below already carries the distinction: `solana_client::get_account_info` returns `nullopt` for account-not-found and throws for transport errors; `get_account_data` then collapses not-found into a throw. Suggest keeping log-and-continue only for genuine absence and letting transport exceptions propagate. `run_outbound` already wlogs and leaves `_last_outbound_epoch` unset on failure, so the next cranker tick re-drives the delivery, and `epoch_in`'s duplicate-chunk no-op makes the re-drive safe on-chain.

plugins/outpost_solana_client_plugin/include/sysio/outpost_solana_client_plugin/outpost_solana_client.hpp
```diff
@@ -23,14 +21,10 @@ inline constexpr size_t SOLANA_MAX_ENVELOPE_BYTES = OPP_MAX_ENVELOPE_BYTES;
...
 /// Solana side. Solana's tx-packet MTU is 1 232 B raw. Tx overhead at the
-/// current 12-account / 1-ix shape is ~492 B. The FINAL chunk tx also
-/// carries one ComputeBudget `request_heap_frame(256_000)` pre-ix (~40 B
-/// for the ComputeBudget program key + ix wrapper) so the consensus-reach
-/// finalize path gets a 256 KiB BPF heap budget instead of the 32 KiB
-/// default that OOM'd at epoch 13. Budget:
-///     492 (overhead) + 40 (last-chunk pre-ix) + chunk_size ≤ 1232
-///     → chunk_size ≤ 700; rounded down to 672 for a comfortable
-///     32-byte safety margin against future ABI / varint slop.
+/// current data-chunk shape remains safely below the cap. Finalization now
```
heifner: "remains safely below the cap" overstates the margin. A full 672-byte chunk under the current 13-key `epoch_in` account list serializes to exactly 1230 bytes -- 2 bytes under the `PACKET_DATA_SIZE` (1232) assert this PR adds to `transaction::serialize`. One added account key (+33 bytes) would make every envelope over ~641 bytes undeliverable client-side. Since the terminal call is now fixture-measured, consider adding a data-chunk case to `sec-94-solana-terminal-budget.json` and asserting this constant against the measured chunk overhead with real margin, so an account-list change fails a test instead of production delivery.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1283,6 +1352,15 @@ void msgch::queueout(uint64_t chain_code,
          has_auth(RESERV_ACCOUNT) || has_auth(get_self()),
          "queueout: caller not authorized to queue outbound attestations");
 
+   {
+      sysio::chains::chains_t chains_tbl(CHAINS_ACCOUNT);
+      auto chain_it = chains_tbl.find(sysio::chains::chain_key{sysio::slug_name{chain_code}});
+      if (chain_it != chains_tbl.end() && chain_it->kind == ChainKind::CHAIN_KIND_SVM) {
+         check(estimate_svm_dynamic_accounts(attest_type, data).has_value(),
```
heifner, line 1363: Discussion, not a blocker: `buildenv`'s fail-closed `check` on an unestimable READY row, combined with `epoch::advance`'s inline `buildenv` fan-out, means one such row would freeze epoch advance for the whole depot -- and there's no evict/expire path for a READY attestation, so `setcode` is the only remedy. Today it's unreachable (every queueout call site queues a covered type, and chains rows can't change kind after registration), so this is belt-and-braces. Options: make this gate unconditional (drop the `find()`-skip so an unregistered-chain queueout of an uncovered type is also rejected at the source), or add a governance evict for READY rows.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1412,6 +1512,12 @@ void msgch::buildenv(uint64_t chain_code) {
             "MAX_ENVELOPE_BYTES -- cannot pack into an envelope");
       entries.pop_back();
       included_ids.pop_back();
+      if (is_svm_destination) {
+         estimated_svm_dynamic_accounts -= included_svm_dynamic_costs.back();
+         included_svm_dynamic_costs.pop_back();
+         check(svm_terminal_budget_fits(estimated_svm_dynamic_accounts),
```
heifner: `svm_terminal_budget_fits` is monotone and the estimator loop already verified every included prefix, so this post-trim "drifted after envelope trim" check can never fire, and `included_svm_dynamic_costs` exists only to feed it. (Likewise the last three conjuncts of `fits()` are implied by the first given the static baselines -- `static_assert`s next to the constants would pin them at compile time.) Fine to keep as belt-and-braces; just flagging none of it is a live guard.

contracts/sysio.msgch/include/sysio.msgch/solana_terminal_budget.hpp
```diff
@@ -0,0 +1,98 @@
+#pragma once
+
+#include <cstddef>
+
+namespace sysio::msgch_svm_terminal_budget {
+
+/// Solana raw transaction packet ceiling used by legacy transactions.
+inline constexpr size_t SVM_TERMINAL_PACKET_LIMIT_BYTES = 1'232;
```
heifner, line 8: Two convention notes. (1) `SVM_TERMINAL_PACKET_LIMIT_BYTES` / `SVM_TERMINAL_ACCOUNT_KEY_LIMIT` duplicate `fc::network::solana::limits::PACKET_DATA_SIZE` / `LEGACY_ACCOUNT_KEY_LIMIT`; the plugin test already includes both headers but never asserts them equal -- one `BOOST_REQUIRE_EQUAL` would pin the two copies together (a versioned-tx bump would otherwise drift them). (2) The per-type costs (8/1/0) returned by `estimate_svm_dynamic_accounts` in the .cpp are consensus constants that hand-mirror the relay manifest; per the repo's no-magic-literals rule they'd read better as named `constexpr`s here beside the other budget constants. Also minor: the fixture's `legacy_packet_limit_bytes` / `runtime_account_limit` / `legacy_account_key_limit` keys are currently read by nothing.

libraries/libfc/src/network/solana/solana_client.cpp
```diff
@@ -1444,6 +1474,8 @@ transaction solana_client::create_transaction(const std::vector<instruction>& in
    // Initialize empty signatures
    tx.signatures.resize(tx.msg.header.num_required_signatures);
 
+   validate_legacy_transaction(tx);
```
heifner: `validate_legacy_transaction` is `(void)tx.serialize()`, and `serialize()` already validates internally, so the create-then-send path serializes the transaction twice (three times counting `sign_transaction`'s message serialize). Dropping this tail call (serialize-time validation is the real enforcement point) or reusing the validated bytes removes the redundant pass.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -333,115 +450,146 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+      if (it != reserve_info_cache.end()) return it->second;
+
+      auto info = reserve_info_for_codes(token_code, reserve_code);
+      return reserve_info_cache.emplace(cache_key, std::move(info)).first->second;
+   };
+   auto is_native_custody = [](const fc::network::solana::solana_public_key& mint) {
+      return mint == fc::network::solana::system::program_ids::SYSTEM_PROGRAM;
+   };
+
+   const auto recipient_pubkeys =
```
heifner, line 469: Efficiency, optional: the five extractors each `ParseFromArray` the same envelope bytes -- parsing once and passing `const Envelope&` (keeping one thin bytes wrapper for the decode-failure wlog) removes four full protobuf decodes per delivery. Same pass: reserve facts are fetched one `get_account_data` round trip per (token, reserve) pair inside the deadline-bounded delivery, and `solana_client` already has `get_multiple_accounts` to batch them; the reserve/vault PDA derivations for identical seeds could ride the same per-pass cache as `reserve_info`. The two new extractors also copy the decode/wlog/dedup scaffolding verbatim from the swap-remit extractor -- a shared walker taking a type filter would leave one copy.

tests/fixtures/sec-94-solana-terminal-budget.json
```diff
@@ -0,0 +1,73 @@
...
+  "terminal_pre_instructions": [
+    {
+      "name": "request_heap_frame",
+      "bytes": 256000
+    }
+  ],
+  "terminal_args": {
+    "epoch_index": 7,
+    "chunk_index": 3,
+    "total_chunks": 3,
```
heifner: Nit: `total_chunks: 3` with `total_bytes: 2526` is an impossible split -- 3 x 672 = 2016, and the relay's own `chunks_for(2526)` is 4. Harmless to the measurements (only the fixed-width scalar sizes matter), but worth correcting to 4 so the dev-026 scenario stays accurate. The companion stale `MAX_CHUNK_BYTES = 768` comment in wire-solana `epoch_in.rs` is the other half.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
huangminghuang: Agreed this is worth doing, but I kept it out of the SEC-94 correctness follow-up to avoid mixing a behavior fix with a relay efficiency refactor. Tracked as follow-up: [SEC-104](https://wire-network.atlassian.net/browse/SEC-104).

heifner (approved):

heifner (approved):

PR #461, opened 2026-06-30
underwriter: route by exact chain_code, not ChainKind (SEC-13/WSA-027)

heifner:
### Problem
The v6 depot identifies every value-bearing underwrite leg by exact slug codes (`chain_code`, `token_code`, `reserve_code`), and an underwriter's collateral is posted per exact `(chain_code, token_code)` on every registered outpost. The underwriter plugin preserved those exact codes inside signed UIC bytes but **selected outpost clients, the UIC `outpost_id`, credit buckets, source-deposit-verification endpoints, and local commit de-dup using only `ChainKind`/`TokenKind`**.

With two active chains of the same VM family (e.g. two EVM outposts — expected immediately after launch) it could not tell them apart: their collateral collapsed into one bucket, a confirmed commit on one suppressed the still-needed commit on the other, and legs were submitted/verified against the wrong chain's endpoint.

> The depot already drops *misrouted* commits via the WSA-005 source-chain binding (#439); this PR fixes the **plugin** so a second same-kind chain is served correctly rather than stalled.

### Changes

**Plugin — key every routing/accounting surface by exact `chain_code`:**
- `commit_key` and the credit-bucket key now carry the exact slug triple, lifted into a new testable `routing_detail.hpp` (`commit_key`, `bucket_key`, `try_debit_buckets`) — two same-VM-family chains hold independent credit, and a confirmed commit on one never suppresses the other.
- `outpost_by_chain` is keyed by `chain_code` and built **one client per active `sysio.chains` row** (real `chain_code`/`external_chain_id`, no more `=0` placeholders) from new repeatable **`--underwriter-{eth,sol}-outpost <chain_code>,<client_id>,<addr…>`** config that replaces the single `--underwriter-eth-client-id`/`-sol-client-id`/`-eth-opreg-addr`.
- `verify_source_deposit_{eth,sol}` resolve the RPC client and the contract/program address per the leg's `chain_code`; preflight now resolves only the chain-agnostic deposit selector / instruction discriminator from the ABI/IDL (the per-chain address comes from config).

**Proto — remove `UnderwriteIntentCommit.outpost_id` (field 4):** a v5 vestige, fully redundant with `chain_code` (field 7) and read by nobody — the depot binds `uic.chain_code` to the proven inbound outpost (WSA-005), and `StakingReward.outpost_id` was already renamed to `chain_code`. The UIC is relayed as raw bytes (not an ABI struct), so the contract ABIs are **byte-identical**; `find_outpost_id` (whose sole purpose was stamping `outpost_id`) is deleted.

### Testing
- **`contracts_unit_test`: 433/433** — incl. new `underwrite_commit_two_evm_chains_route_per_chain` (a commit for a second EVM chain's leg delivered from the first EVM outpost is dropped; the matched delivery lands).
- **`underwriter_plugin` tests: 25/25** — incl. `test_underwriter_routing.cpp` (two same-kind chains hold independent credit; same-chain/different-reserve legs stay distinct).

### Notes
- Contract ABIs are byte-identical → **no `SystemContractTypes.ts` regen / TS rebuild** needed. `.wasm` refreshed for `msgch`/`uwrit` (UIC decode) plus `opreg`/`reserv` (incidental recompile from the regenerated proto header; same size, no behaviour change).
- The `attestations.proto` change requires republishing `@wireio/opp-*-models` for the e2e/outpost consumers; the cross-repo trace confirmed **zero consumer code changes** (ETH/SOL outposts relay the UIC opaquely; TS decodes for display only).
- **Operator-facing config change:** the single ETH/SOL client-id + opreg-addr options are replaced by the per-chain `--underwriter-{eth,sol}-outpost` options.
- No flow-test (e2e) in this PR — deferred per the current e2e infra state.


plugins/underwriter_plugin/include/sysio/underwriter_plugin/routing_detail.hpp
```diff
@@ -0,0 +1,93 @@
...
+inline bool try_debit_buckets(credit_buckets& remaining,
+                              const leg_bond& src, const leg_bond& dst) {
+   const uint64_t src_req = src.require;
+   const uint64_t dst_req = dst.require;
+   if (src_req == 0 && dst_req == 0) return false;
+
+   // Same bucket on both legs: a single row must cover the combined draw.
+   if (src_req > 0 && dst_req > 0 && src.bucket == dst.bucket) {
+      auto it = remaining.find(src.bucket);
+      if (it == remaining.end() || it->second < src_req + dst_req) return false;
```
huangminghuang: [P2] For same-bucket legs this adds two user-controlled `uint64_t` amounts before the sufficiency check, so a pair such as `UINT64_MAX + 1` wraps to 0. That makes an impossible request appear coverable and mutates the remaining credit incorrectly; please use `uint128_t` or guard `src_req > UINT64_MAX - dst_req`, and add a routing regression for the overflow case.

plugins/underwriter_plugin/src/underwriter_plugin.cpp
```diff
@@ -2324,46 +2384,48 @@ void underwriter_plugin::plugin_startup() {
...
-                                                   /*chain_code=*/0,
-                                                   /*chain_id=*/0,
-                                                   _impl->sol_program_id);
-         ilog("underwriter_plugin: wired SOL outpost_client (program={})",
-              _impl->sol_program_id);
-      } else {
-         wlog("underwriter_plugin: SOL outpost_client NOT wired "
-              "(sol_client_id='{}', sol_program_id='{}')",
-              _impl->sol_client_id, _impl->sol_program_id);
+      for (const auto& [chain_code, ep] : _impl->outpost_endpoints) {
```
huangminghuang, line 2464: [P2] This only wires the endpoints that were configured, but preflight/availability still derive the served set from every registered non-depot chain in `sysio.chains`. If one registered chain is missing or has the wrong family in `--underwriter-*-outpost`, the cron can still start and `select_coverable` can pick requests for that chain; submission then fails after potentially landing the other leg. Preflight should require one endpoint per registered outpost chain and verify `ep.kind` matches the registry kind before scheduling the scan job.

huangminghuang (commented):
One non-inline note: `plugins/underwriter_plugin/README.md` still documents the removed scalar options (`--underwriter-eth-client-id`, `--underwriter-sol-client-id`, `--underwriter-eth-opreg-addr`, `--underwriter-sol-program-id`) instead of the new per-chain `--underwriter-eth-outpost` / `--underwriter-sol-outpost` configuration, so operators following the README will configure options the plugin no longer accepts.

plugins/underwriter_plugin/include/sysio/underwriter_plugin/routing_detail.hpp
heifner: Fixed in cbbddd40e2. `try_debit_buckets` now sums the same-bucket dual-leg requirements in `__uint128_t` before the sufficiency check, so a pair past `UINT64_MAX` (e.g. `UINT64_MAX + 1`) can't wrap to a small coverable-looking value; the cast back to `uint64` only runs on the path where the combined draw provably fits. Added `same_bucket_dual_leg_sum_cannot_overflow`, covering both the wrap case and a max-balance bucket.

plugins/underwriter_plugin/src/underwriter_plugin.cpp
heifner: Fixed in cbbddd40e2. Added preflight Check 2: it requires a configured `--underwriter-{eth,sol}-outpost` endpoint for every registered non-depot chain and verifies `ep.kind` matches the registry kind, returning false (no cron scheduled) on a missing or wrong-family entry — so the scan loop can no longer select a request it can't fully commit. The comparison is lifted into a pure `find_endpoint_coverage_gap` helper in `routing_detail.hpp` with unit tests for the missing-endpoint, wrong-family, extra-config, and all-covered cases.

plugins/underwriter_plugin/src/underwriter_plugin.cpp
```diff
@@ -760,6 +835,9 @@ struct underwriter_plugin::impl {
          // round-trip — the variant carries the symbolic name and `.as<T>()`
          // recovers the typed value without a string switch.
          outpost_chain_kinds[chain_code] = obj["kind"].as<ChainKind>();
```
huangminghuang, line 852: `read_outpost_registry()` is treating every non-depot chain row as a live outpost, but `sysio.chains::regchain` creates post-bootstrap rows with `active=false` until `activchain` runs. With the new endpoint coverage and exact-chain `is_available()` checks, registering a future/inactive chain can make the underwriter refuse startup or stop serving the currently active chains until that future chain also has endpoint config and collateral. This should skip `!active` rows, matching `sysio.opreg` and `batch_operator_plugin`.

heifner: Good catch — fixed in 5539ec4d80. `read_outpost_registry` now skips `active == false` rows (mirroring `batch_operator_plugin`'s `is_depot || !active` filter), so inactive/future `regchain` rows never enter `outpost_chain_kinds` — keeping them out of the endpoint-coverage preflight, `is_available()`, and the authex/balance checks. Registering a not-yet-activated chain no longer blocks startup or halts the active chains. Preflight docs/messages and the README now say "active" accordingly.

`read_outpost_registry` reads live chain state, so it's exercised by the flow tests rather than the unit suite (consistent with the existing preflight coverage); happy to add a flow case with an inactive row if you'd like one.

huangminghuang (approved):
Reviewed current head 1ff31f830416d409399922df2cd4dedfe634fe3b; no actionable findings.

PR #464, opened 2026-06-30
SEC-84: Harden CLI secret file outputs

huangminghuang:
## Motivation / Why

SEC-84 covers secret-bearing CLI outputs that were written with regular stream file creation, allowing the caller's umask to decide the final file mode. Under a permissive umask, generated private keys, BLS keys, or wallet passwords could become group/world-readable. This PR centralizes those writes behind a POSIX helper that creates owner-only outputs and avoids publishing incomplete replacements.

## What Changed

- Added `fc::secure_output_file` / `fc::write_secure_file` for secret file writes.
- On POSIX, the helper writes to an owner-only temporary file in the target directory, syncs and closes it, atomically publishes it with `rename`, then syncs the containing directory.
- Existing final symlinks and non-regular final targets are rejected; uncommitted temporary files are discarded on destruction.
- Temporary file names are bounded to the target directory's filename limit so near-limit destination names still work.
- Internal fd/temp-file cleanup uses `fc::make_scoped_exit` while preserving explicit close-error reporting.
- Updated affected `clio` key/wallet password and `sys-util bls` `--file` paths to use the secure helper instead of `std::ofstream`.
- Added regression coverage for permissive umask behavior, existing-file replacement without disturbing already-open readers, move operations, destructor discard before close, near-limit filenames, non-regular targets, and final symlink targets.

## Reviewer Notes

- POSIX hardening guarantees are POSIX-only; Windows currently keeps the fallback `std::ofstream` behavior.
- `wallet create --file` uses the same secure helper, but does not have a separate service-backed wallet integration test.
- System-contract e2e validation is not applicable; this PR does not modify `contracts/sysio.*`.

## Validation

- Focused local Release validation passed: `fc`, `test_fc`, `clio`, and `sys-util` build targets; `secure_file_test_suite` (`7` cases); `clio_em_key_test.py`; `sysio_util_bls_test.py`; `git diff --check`.
- GitHub PR CI passed for latest head `5f5b0ac7c7ec2db9061377fb96d0c2476ac1b162`: Linux amd64 matrix and Apple Silicon.

### DIFF
--- a/libraries/libfc/include/fc/io/secure_file.hpp
+++ b/libraries/libfc/include/fc/io/secure_file.hpp
@@ -12,15 +12,14 @@ namespace fc {
 /**
  * Owner-only output file for newly generated secret material.
  *
- * On POSIX systems the file is opened without following the final path component
- * when the platform exposes that flag, restricted to owner read/write
- * permissions, truncated, and then written through the opened descriptor.
+ * On POSIX systems content is written to an owner-only temporary file in the
+ * target directory and then committed with rename after a successful close. The
+ * final target is rejected when it is already a symlink or non-regular file.
  */
 class secure_output_file {
 public:
    /**
-    * Opens a secret output file and applies owner-only permissions before any
-    * caller-provided content is written.
+    * Opens an owner-only temporary file for a secret output.
     *
     * @param file_path path to the secret file to create or replace
     * @throws std::ios_base::failure when the file cannot be opened securely
@@ -59,13 +58,15 @@ public:
 private:
    void close_noexcept() noexcept;
 
+   std::filesystem::path file_path_;
+
 #ifndef _WIN32
    static constexpr int invalid_file_descriptor = -1;
    int                  file_descriptor_        = invalid_file_descriptor;
+   std::filesystem::path temp_file_path_;
 #else
    std::ofstream file_;
 #endif
-   std::filesystem::path file_path_;
 };
 
 /**
--- a/libraries/libfc/include/fc/io/secure_file.hpp
+++ b/libraries/libfc/include/fc/io/secure_file.hpp
@@ -10,11 +10,17 @@
 namespace fc {
 
 /**
- * Owner-only output file for newly generated secret material.
+ * Output file for newly generated secret material.
  *
- * On POSIX systems content is written to an owner-only temporary file in the
- * target directory and then committed with rename after a successful close. The
- * final target is rejected when it is already a symlink or non-regular file.
+ * On POSIX systems content is written to an owner-only temporary file in the target directory. Calling close()
+ * publishes the file by syncing the temporary file, renaming it into place, and syncing the target directory.
+ * Destroying the object before the rename commit point discards the temporary file without publishing it. If close()
+ * reports a directory sync failure after the rename succeeds, the final file may already be visible but its directory
+ * entry durability could not be confirmed. Existing final symlinks and non-regular final targets are rejected, but
+ * intermediate directory symlinks are still resolved by the platform.
+ *
+ * On Windows this helper currently falls back to plain std::ofstream semantics; owner-only permissions, final-target
+ * rejection, and temporary-file rename commit guarantees are POSIX-only.
  */
 class secure_output_file {
 public:
@@ -30,6 +36,9 @@ public:
    secure_output_file& operator=(const secure_output_file&) = delete;
    secure_output_file(secure_output_file&& other) noexcept;
    secure_output_file& operator=(secure_output_file&& other) noexcept;
+   /**
+    * Destroys the output file, discarding any uncommitted POSIX temporary file.
+    */
    ~secure_output_file();
 
    /**
@@ -41,10 +50,11 @@ public:
    void write(std::string_view content);
 
    /**
-    * Closes the opened secret file and reports close errors.
+    * Closes the opened secret file, publishes it on POSIX, and reports close errors.
     *
-    * Destruction also closes the file, but close errors are intentionally not
-    * thrown from the destructor.
+    * Destruction before the POSIX rename commit point discards pending output. Once the rename succeeds, a later
+    * directory sync failure can still be reported even though the final file may already be visible. Close errors are
+    * intentionally not thrown from the destructor.
     *
     * @throws std::ios_base::failure when close fails
     */
--- a/libraries/libfc/include/fc/io/secure_file.hpp
+++ b/libraries/libfc/include/fc/io/secure_file.hpp
@@ -1,0 +1,80 @@
+#pragma once
+
+#include <filesystem>
+#include <string_view>
+
+#ifdef _WIN32
+#include <fstream>
+#endif
+
+namespace fc {
+
+/**
+ * Owner-only output file for newly generated secret material.
+ *
+ * On POSIX systems the file is opened without following the final path component
+ * when the platform exposes that flag, restricted to owner read/write
+ * permissions, truncated, and then written through the opened descriptor.
+ */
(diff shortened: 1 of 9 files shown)


heifner (commented):
Reviewed the secret-file hardening end to end: the core helper, both call sites, exception handling in clio and sys-util, and the tests. Solid work. The atomic-secure-write core is textbook-correct (`O_EXCL|O_NOFOLLOW|O_CLOEXEC` temp -> `fchmod 0600` -> `fsync` -> same-directory `rename`), errno discipline and move semantics are right, and both CLIs surface the new `std::ios_base::failure` cleanly (clio's outer `catch(std::exception)`; sys-util's `err_guard` `catch(...)` -> `CLI::RuntimeError`, which derives from `ParseError` and so is caught by `CLI11_PARSE` -- no `terminate` on a bad `--file`). No consensus-zone code, and no critical issues.

A few refinements worth discussing, inline. The headline one: the destructor rolls back (unlinks the temp with no `rename`) while the header doc implies it commits, so a direct user of `secure_output_file` who forgets `close()` silently gets no file. `write_secure_file` always closes, so the shipped paths are fine, but the contract deserves a doc fix plus a test. Also worth a look: fsyncing the parent directory after the rename for full crash-durability, and the Windows path being an unhardened plain stream.

Nice touches: `create_wallet` opens the temp before the service `call()` and rolls it back if the call throws (better than the old empty-file-left-behind), the other remaining `ofstream` sites are all non-secret outputs, and the tests are umask-scoped with `shlex.quote`.

None of this is blocking -- hardening/robustness refinements on an already-correct change.

libraries/libfc/include/fc/io/secure_file.hpp
```diff
@@ -0,0 +1,81 @@
...
+    * Writes bytes to the opened secret file.
+    *
+    * @param content content to write
+    * @throws std::ios_base::failure when the write fails
+    */
+   void write(std::string_view content);
+
+   /**
+    * Closes the opened secret file and reports close errors.
+    *
```
heifner, line 54: This comment implies destruction commits the file, but the destructor (`close_noexcept`) unlinks the temp with no `rename` -- it rolls back. Suggest stating plainly that destruction *without* a prior `close()` discards the pending file, and that `close()` is what commits. See the note on `close_noexcept` below.

libraries/libfc/src/io/secure_file.cpp
```diff
@@ -0,0 +1,290 @@
...
+#endif
+}
+
+void secure_output_file::close_noexcept() noexcept {
+#ifndef _WIN32
+   if (file_descriptor_ != invalid_file_descriptor) {
+      ::close(file_descriptor_);
+      file_descriptor_ = invalid_file_descriptor;
+   }
+   unlink_noexcept(temp_file_path_);
```
heifner, line 333: Destruction discards (unlinks the temp, no rename) rather than commits, so a caller who writes but forgets `close()` silently gets no file:

```cpp
{ fc::secure_output_file f(path); f.write(secret); }  // no close() -> no file, no error
```

This is a defensible RAII choice (don't commit half-written secrets; avoid a throwing `rename` in a noexcept dtor), but it contradicts the `close()` doc in the header. `write_secure_file` always calls `close()`, so the shipped call sites are safe. Worth a one-line test -- write, destroy, assert the file is absent -- to pin the contract.

libraries/libfc/src/io/secure_file.cpp
```diff
@@ -0,0 +1,290 @@
...
+   }
+
+   if (::close(fd) == -1) {
+      const int saved_errno = errno;
+      unlink_noexcept(temp_file_path_);
+      temp_file_path_.clear();
+      throw_file_failure(file_path_, "close temporary file", saved_errno);
+   }
+
+   if (::rename(temp_file_path_.c_str(), file_path_.c_str()) == -1) {
```
heifner, line 308: `close()` fsyncs the file data but not the parent directory after this `rename`. POSIX only guarantees the rename survives a crash once the containing directory is fsync'd, so on power loss right after `close()` returns success the secret file can be missing or reverted despite the reported success. For a helper framed as "committed with rename after a successful sync," fsyncing the directory fd here closes the durability gap. Low severity (crash window only), but it's the kind of completeness a hardening change is expected to cover.

libraries/libfc/src/io/secure_file.cpp
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * Returns the platform open flags needed for temporary secret output files.
+ */
+int secret_file_open_flags() {
+   int flags = O_WRONLY | O_CREAT | O_EXCL;
+#ifdef O_CLOEXEC
+   flags |= O_CLOEXEC;
+#endif
+#ifdef O_NOFOLLOW
+   flags |= O_NOFOLLOW;
```
heifner, line 43: `O_NOFOLLOW` (and the `lstat` check) only guard the final path component; a symlink in an intermediate directory is still followed. Full protection needs `openat` + `RESOLVE_NO_SYMLINKS`. Acceptable for this threat model (the user chooses where to write their own keys; umask and final-symlink are the stated targets) -- just worth a one-line note in the header so a reader doesn't assume total symlink safety.

libraries/libfc/include/fc/io/secure_file.hpp
```diff
@@ -0,0 +1,81 @@
...
+   void close_noexcept() noexcept;
+
+   std::filesystem::path file_path_;
+
+#ifndef _WIN32
+   static constexpr int invalid_file_descriptor = -1;
+   int                  file_descriptor_        = invalid_file_descriptor;
+   std::filesystem::path temp_file_path_;
+#else
+   std::ofstream file_;
```
heifner, line 78: On Windows the class degrades to a plain `ofstream` with `trunc`: no owner-only mode, no symlink/non-regular rejection, no temp+rename atomicity -- a `secure_output_file` that isn't secure on one platform. This stream is also `std::ios::binary` whereas the old clio/bls `ofstream` was text mode, so Windows key files change from CRLF to LF. POSIX is the target and CI is POSIX-only (the whole test suite is `#ifndef _WIN32`), so low impact -- but a header note setting expectations would help.

libraries/libfc/src/io/secure_file.cpp
```diff
@@ -0,0 +1,290 @@
...
+ */
+void close_descriptor_noexcept(int file_descriptor) noexcept {
+   if (file_descriptor >= 0)
+      ::close(file_descriptor);
+}
+
+/**
+ * Rejects existing final targets that should not be replaced by a secret file.
+ */
+void throw_if_target_is_not_replaceable(const std::filesystem::path& file_path);
```
heifner: This forward declaration looks unnecessary: `throw_if_target_is_not_replaceable` is defined at line 105 and first used at 143, and nothing between 91 and 105 references it.

programs/clio/main.cpp
```diff
@@ -3125,8 +3126,8 @@ int main( int argc, char** argv ) {
          std::cerr << localized("saving password to ${filename}", ("filename", password_file)) << std::endl;
          auto password_str = fc::json::to_pretty_string(v);
          boost::replace_all(password_str, "\"", "");
-         std::ofstream out( password_file.c_str() );
-         out << password_str;
+         password_out->write(password_str);
```
heifner: `password_out` is engaged only by the earlier `^` assert plus the `if (!password_file.empty())` emplace, so `operator->` here is UB if that invariant is ever broken by a future edit. `password_out.value()` would fail-fast instead. Minor.

heifner (approved):

PR #465, opened 2026-07-01
Opp: harden underwrite-commit dispatch against row bloat and name-abort

heifner:
## SEC-16 (WSA-030) — unmatched UIC commits bloat pending request rows

sysio.uwrit::rcrdcommit appended a commit_entry for the UIC's underwriter before checking whether the UIC's (chain, token, reserve) triple matched either leg of the pending request. A UIC matching neither leg left an inert
INTENT_SUBMITTED entry (and could re-arm a disqualified one) without arming a leg or reaching winner selection, so repeated unmatched commits with distinct account names grew commits_by unbounded — a depot RAM / scan-cost DoS.

Leg classification now runs against the pre-fetched request snapshot before any mutation. A UIC matching neither leg is dropped with no state change, so find_or_create_commit never runs and the row cannot grow. Per-underwriter dedup
already bounds legitimate racers.

## SEC-17 (WSA-031) — invalid UIC account names throw inside evalcons dispatch

sysio.msgch::dispatch_underwrite_commit constructed name{uic.uw_account.name} after only an empty-string check. The CDT name constructor aborts on an overlong / out-of-charset / bad-final-symbol string, so a decode-clean UIC with
a nonempty-but-invalid name aborted the whole evalcons/apply_consensus delivery and stalled consensus — the failure the never-throw dispatch convention forbids.

The relayed account string is now pre-validated through a shared, non-throwing sysio::opp::safe::parse_wire_account_name helper (new contracts/sysio.opp.common/name_ops.hpp) and dropped when malformed. The node-owner (parse_owner_name) and WIRE-recipient (parse_wire_name) paths are
refactored to reuse it so the account-name rules live in one place.

## Tests

sysio_dispatch_tests gains cases for unmatched-leg no-growth (single and distinct-underwriter), malformed-name drop-without-throw (uppercase, hyphen, over-long), and a matched-leg control. sysio.msgch and sysio.uwrit are recompiled; ABIs are unchanged. No reference-data regeneration is required - neither contract is embedded in the deep-mind / snapshot / consensus-blockchain reference flows.


contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -1466,18 +1463,28 @@ void uwrit::rcrdcommit(uint64_t uwreq_id,
...
+   // A UIC whose leg matches neither the source nor the destination is not bound to this request.
+   // Drop it with no mutation so a stream of unmatched commits cannot grow commits_by or re-arm a
+   // disqualified entry. Fail closed, never check().
+   if (!is_source && !is_dest) {
+      sysio::print("rcrdcommit: uwreq ", uwreq_id,
+                   " UIC leg matches neither source nor destination, skipping\n");
+      return;
+   }
+
    reqs.modify(same_payer, pk, [&](auto& r) {
```
huangminghuang, line 1496: This still leaves a row-growth path for UICs whose triple matches a real source or destination leg. `msgch` now only validates that `uw_account.name` is syntactically constructible before passing it here, and then this block calls `find_or_create_commit` before checking whether the account is a registered/ACTIVE underwriter or whether the UIC signature is valid. A consensus-delivered matched-leg UIC can therefore vary valid but nonexistent account names and append one `commit_entry` per name; on dual-outpost swaps a single-leg commit will not reach `try_select_winner`, and on single-leg swaps inactive/bad candidates are disqualified but the row remains. Can we drop non-underwriter/non-registered names before mutating, or otherwise bound `commits_by`, and add a regression with distinct valid account names on a matched leg?

huangminghuang (commented):
Review note: one remaining row-growth issue in the UIC commit path.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Good catch — the leg-binding guard only closed the unmatched-leg path. Fixed in 6c448e1b8d: `rcrdcommit` now drops a UIC before any mutation unless the underwriter is a registered ACTIVE underwriter (the same `is_active_underwriter` gate `try_select_winner` already applies), so `commits_by` can't grow for valid-but-unregistered names on a matched leg.

I gated on ACTIVE (not just registered) because the DISQUALIFIED re-arm already requires a fresh `rcrdcommit` to re-enter, so a placeholder for an inactive/wrong-role candidate only bloats the row — dropping it costs no liveness, and it removes the "disqualified but the row remains" case you flagged. `try_select_winner` keeps the check as defensive depth.

Added the matched-leg distinct-names regression (three valid non-underwriter names on a matched source leg → `commits_by` stays empty); the two `swap_winner_*_is_disqualified` cases become `swap_commit_*_is_dropped` since those candidates are now dropped at ingestion rather than recorded-then-disqualified.

huangminghuang (approved):
Static review of head 6c448e1b8d6b6a315db0bae00913588414a91f74 found no actionable issues.

PR #467, opened 2026-07-01
Gate block-producer scheduling on opreg collateral

heifner:
Block producers must now post a slashable collateral bond to be scheduled: each producer has to be an ACTIVE OPERATOR_TYPE_PRODUCER operator in sysio.opreg, the same collateral identity batch operators use. This mirrors the batch-operator delivery gate but applies at schedule-build time.

update_ranked_producers changes:
- skips any producer that is not opreg-ACTIVE (via a shared is_op_active predicate moved to opreg_status.hpp, now also used by reward distribution);
- backfills a vacated active slot from the next eligible standby (rank up to standby_end_rank), so a de-collateralized producer is replaced rather than freezing the schedule;
- replaces the "never publish a smaller schedule" guard with a BFT safety floor (min_schedule_size, default 4): the schedule may shrink below the nominal 21 but never below a value that would concentrate finality onto too few nodes; below the floor the last good schedule is retained.

Withdrawing collateral (status -> UNKNOWN), slashing, or termination drop the producer on the next onblock rebuild; re-depositing restores it. The ABI is unchanged, so no reference-data regeneration is required.

A new sysio_producer_eligibility_tests suite covers scheduling, removal, standby backfill, the floor, and the operator-type check. The finalizer-key tests now register their producers as opreg operators; shared BLS test vectors live in finalizer_test_keys.hpp. The TestHarness bootstrap deploys sysio.opreg and registers producers as operators so producer_rank_test.py exercises the gated path; other tests and the e2e bootstrap use setprodkeys plus bulk setfinalizer, where update_ranked_producers is inert and the gate has no effect until producer scheduling moves to the dynamic path.

min_schedule_size is the hard BFT-liveness floor of 4; raise it for a stronger anti-concentration posture.


huangminghuang (approved):
Approved after review at db2e82f47deab76ce461b7052af3655599401699. No findings from the opreg scheduling/reward gate, producer eligibility tests, or bootstrap harness changes.

PR #468, opened 2026-07-01
Test: cover invalid staker-name reject domain in dclaim onreward dispatch

heifner:
sysio.dclaim::onreward runs inside the OPP inbound dispatch chain (msgch::evalcons), where a throw rolls back the consensus-tipping delivery. It already validates the cross-chain-supplied staker_wire_account with is_valid_name_string before constructing name(), so an invalid spelling parks the credit by native address rather than aborting.

This adds the missing dispatch-level regression coverage for that guard: the invalid-name test now spans the full reject domain (overlong, uppercase, and hyphenated names) and a new positive case confirms a valid dotted+digit account name still credits pending_claims. Test-only; no contract, ABI, or reference-data change.

### DIFF
--- a/contracts/tests/sysio.dclaim_tests.cpp
+++ b/contracts/tests/sysio.dclaim_tests.cpp
@@ -245,14 +245,30 @@ BOOST_FIXTURE_TEST_CASE(flushexpired_reverts_expired_pending, sysio_dclaim_teste
 
 // onreward runs inside the OPP inbound dispatch chain (msgch::evalcons), where an abort rolls back
 // the consensus-tipping deliver and stalls epoch advancement. A cross-chain-supplied
-// staker_wire_account that name() would reject (>13 chars / bad alphabet) must NOT abort: it is
-// treated as unlinked and the credit is parked in unmapped_tokens by native address.
+// staker_wire_account that name() would reject (overlong at > 13 chars, or bearing a character
+// outside the ".12345a-z" name alphabet such as uppercase or '-') must NOT abort: onreward treats
+// it as unlinked and parks the credit in unmapped_tokens by native address. Each malformed spelling
+// is exercised on a distinct native address so it lands as its own row (unmapped + dedupe cursor
+// are keyed by native address), covering the full reject domain the dispatch name guard absorbs.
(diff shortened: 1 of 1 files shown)


contracts/tests/sysio.dclaim_tests.cpp
```diff
@@ -245,14 +245,30 @@ BOOST_FIXTURE_TEST_CASE(flushexpired_reverts_expired_pending, sysio_dclaim_teste
...
+// it as unlinked and parks the credit in unmapped_tokens by native address. Each malformed spelling
+// is exercised on a distinct native address so it lands as its own row (unmapped + dedupe cursor
+// are keyed by native address), covering the full reject domain the dispatch name guard absorbs.
 BOOST_FIXTURE_TEST_CASE(onreward_invalid_wire_account_parks_unmapped, sysio_dclaim_tester) { try {
-   BOOST_REQUIRE_EQUAL(success(),
-      onreward(MSGCH_ACCOUNT, 1, "thisnameistoolong", ChainKind::CHAIN_KIND_EVM, addr20,
-               1000, 7, 100));
-   // Soft-handled: credit parked as unmapped (unlinked), not aborted.
-   BOOST_REQUIRE(!unmapped_row(1).is_null());
+   const struct { const char* wire_account; char addr_byte; } cases[] = {
```
huangminghuang, line 256: This covers overlong strings and out-of-alphabet characters, but it still misses one CDT `name()` reject class: 13-character names whose final symbol is above `j` (for example, `aaaaaaaaaaaak`). Those are length-valid and alphabet-valid, but `name()` still aborts because the 13th symbol only has 4 bits. Since this test is meant to cover the full dispatch reject domain, please add one such case with a distinct native address.

heifner: Added in 742801b5 as a fourth case: `aaaaaaaaaaaak` (twelve `a`s + `k`, value 16) on a distinct native address. Confirmed it hits exactly this class in the guard -- `is_valid_name_string` rejects via `if (i == 12 && v > 15) return false;`, so `onreward` parks it in `unmapped_tokens` rather than aborting. The dispatch test now covers all three `name()` reject classes (overlong, out-of-alphabet, and 13th symbol above `j`). Good catch, thanks.

huangminghuang (approved):
Reviewed PR #468 at 742801b545. No actionable findings.

PR #469, opened 2026-07-02
Test: keep ship_restart_test clear of the state-history-stride boundary

heifner:
ship_restart_test failed on master CI ([run 28587809209](https://github.com/Wire-Network/wire-sysio/actions/runs/28587809209/job/84763751464)) with `OSError: [Errno 22] Invalid argument` raised by the test's own `f.truncate(newSize)` during the "Truncated index file test" phase.

The SHiP node runs with `--state-history-stride 200` and cluster bootstrap takes close to 200 half-second blocks, so the test's shutdown can land exactly on a stride boundary. When the SHiP node stops exactly at block 200, rotation has just moved every entry into `retained/` and the head `chain_state_history.log`/`.index` are 0 bytes: the CI artifact shows 0-byte head files, a fully populated `retained/chain_state_history-1-200.*`, and the SHiP node's last applied block exactly #200. The test then computes `newSize = getsize(index) - 8 = -8`, and `truncate(-8)` raises EINVAL. The preceding "index file removed" phase also passes vacuously when the files are empty. nodeop behaved correctly throughout; this is purely a test-timing bug, and upstream Spring's ship_restart_test.py has the same latent issue.

Before pausing the producer, wait until the head block is a few blocks past a stride boundary and not within pause latency of the next one, so the head log and index always contain entries when the SHiP node stops. The stride is now a named constant shared with the SHiP node arguments. When the guard engages, the rest of the test exercises the restart scenarios against a post-rotation head log with retained/ segments present, which is exactly the shape the CI failure exposed.

### DIFF
--- a/tests/ship_restart_test.py
+++ b/tests/ship_restart_test.py
@@ -34,6 +34,10 @@ testSuccessful=False
 prodNodeId = 0
 shipNodeId = 1
 
+# SHiP log rotation stride. The shutdown guard below keeps the stop point clear of stride
+# boundaries so the head log/index files the test operates on are never empty.
+shipStride = 200
+
 origStateHistoryLog   = ""
 stateHistoryLog       = ""
 origStateHistoryIndex = ""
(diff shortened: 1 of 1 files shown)


huangminghuang (approved):
Reviewed current head 08c99c350e448d163dacf1d730f505793cf2819e; no actionable findings.

PR #470, opened 2026-07-02
Test: cover malformed SwapRequest bytes through consensus dispatch

heifner:
sysio.msgch::dispatch_attestation forwards ATTESTATION_TYPE_SWAP_REQUEST bytes to sysio.uwrit::createuwreq undecoded, so createuwreq's own decode guard is all that stands between attacker-shaped attestation bytes and a throw that would abort the consensus-tipping evalcons delivery. That guard already fails soft on master (781aeceafe): a payload that fails to decode is logged and skipped -- no throw, no uwreq row -- and the duplicate-id guard runs before the decode.

This adds the missing dispatch-level regression coverage: one envelope drives empty bytes, a truncated protobuf tag, and undecodable junk bytes through the full deliver -> evalcons -> apply_consensus -> inline-createuwreq chain ahead of a valid swap (delivery succeeds, the malformed entries create no uwreq rows, the valid one behind them still creates exactly one), and a duplicate-id case proves re-delivery -- even with different or malformed bytes -- no-ops without touching the original row. Verified the envelope case fails against the pre-fix uwrit WASM with the exact audited abort ("failed to decode SwapRequest"). Test-only; no contract, ABI, or reference-data change.

### DIFF
--- a/contracts/tests/sysio.dispatch_tests.cpp
+++ b/contracts/tests/sysio.dispatch_tests.cpp
@@ -653,6 +653,18 @@ public:
          abi_serializer::create_yield_function(abi_serializer_max_time));
    }
 
+   /// Next attestation id `sysio.msgch::mint_att_id` will return: the `attseq` singleton's
+   /// `next`, or 1 before the row is materialised. Outbound queueouts share the same sequence
+   /// (e.g. reserve registration emits RESERVE_READY), so a test that needs the ids an inbound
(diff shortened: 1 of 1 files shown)


huangminghuang (approved):
Re-reviewed current head. No code findings from me; the observed GCC failure is in an unrelated long-running nodeop test, not this dispatch coverage.

PR #472, opened 2026-07-02
Test: Stabilize nodeop late block test reconnect timing

huangminghuang:
## Summary
- Stabilize `nodeop_late_block_test.py` by reconnecting the Node_02/Node_03 bridge as soon as Node_03 reaches `defproducerk`, before canonical-block polling can consume the producing-slot window.
- Keep the fast-path diagnostic assertion for `applying blocks while producing: head's branch is locked out`.
- Anchor fork-switch log detection to post-partition abandoned block numbers so pre-partition ramp-up micro-forks cannot poison the switch decision.
- Re-scan anchored switch lines at the final producer-window assertion so late `defproducerl` cascade switches still allow the expected `defproducera` outcome.
- Fix `nodeop_read_terminate_at_block_test.py` snapshot replay timing by waiting for each replay node to persist blocks relative to its own snapshot height instead of relying on a fixed block 250 cutoff.
- Verify the post-terminate relaunch from node state via `get_info`; after `--truncate-at-block`, the local fork database may correctly be pruned to the terminate block, while the later cluster restart still verifies sync advancement with peers.

## Diagnosis
- The scheduled macOS arm64 run failed in `nodeop_late_block_test.py` because Node_03 converged to the canonical branch, but the bridge reconnected after Node_03 had already left the relevant producing slot. That made the producer_plugin fast-path log timing-dependent.
- The Linux asserton shard later failed in `nodeop_read_terminate_at_block_lr_test` because replay terminate targets are snapshot-relative, but the test stopped production after a fixed block. A node could terminate at its expected block and then relaunch with a truncated fork database that had no block past the terminate height to log immediately.

Failed runs:
- https://github.com/Wire-Network/wire-sysio/actions/runs/28578252480/job/84731924627
- https://github.com/Wire-Network/wire-sysio/actions/runs/28604312519/job/84820500494?pr=472

## Review Follow-Up
- Addressed review feedback from @heifner by parsing `switching forks` log lines for the abandoned block number and producer, then filtering to switches after `firstIProdBlockNum`.
- This ignores benign pre-partition micro-forks while preserving the load-bearing rule that any later partition-fork switch from `defproducerl` means `defproducera` may win the final producer window.

## Validation
- `python3 -m py_compile tests/nodeop_late_block_test.py`
- `cmake -E copy tests/nodeop_late_block_test.py build/release-cdt-deb-opt/tests/nodeop_late_block_test.py`
- `cd build/release-cdt-deb-opt && python3 tests/nodeop_late_block_test.py -v --keep-logs`
- `python3 -m py_compile tests/nodeop_read_terminate_at_block_test.py`
- `git diff --check`
- `cmake -E copy tests/nodeop_read_terminate_at_block_test.py build/release-cdt-deb-opt/tests/nodeop_read_terminate_at_block_test.py`
- `cd build/release-cdt-deb-opt && python3 tests/nodeop_read_terminate_at_block_test.py -v`
- Review reported local UBSAN WSL runs that reproduced the final-window issue and validated the anchored-switch approach against the failed logs.


heifner (commented):
Ran this locally against an UBSAN build (WSL), which amplifies the timing races this test is sensitive to. The core fix works: reconnecting at defproducerk before the canonical-block polling put the reconnect inside node_03's producing window in all three runs, and the `applying blocks while producing` fast-path assertion fired every time; the convergence and canonical i-block checks passed as well. Moving the i-block capture after the relaunch is safe since those blocks are finalized by the 3/4-finalizer side and cannot reorg.

However, the final producer-window assertion failed 2 of 3 runs, via two different mechanisms in the block this PR rewrites:

1. **Poisoned switch-line anchor** (fails on this PR and on master alike). A benign pre-partition slot-boundary micro-fork during ramp-up -- `switching forks from ... (block number 176 defproducerl) ... to ... (block number 176 defproducera)`, a full minute before the partition -- matched the unanchored `findInLog("switching forks .* defproducer[jkl]")`. The defproducerl escape then read that same line and set `expectedProducers = {defproducera}`, while the window actually held exactly the post-switch k#285-292 + l#293-297 blocks this PR's {k, l} expectation anticipates, so it failed with `expected one of ['defproducera'] in blocks 285-297`. f94cf3d048 anchored only the second search; widening the first search from defproducerk to defproducer[jkl] in 983bb9a804 made it susceptible to ramp-up micro-forks.

2. **The any-later-l escape is load-bearing.** In another run, node_03 kept producing fork blocks for ~11s after the fast path fired (slow block application), cascaded through `switching forks from ... l#248 ... to ... i#228`, and its lone canonical l#229 then lost the slot-boundary race to a#229 -- zero k/l blocks made the canonical chain and defproducera won everything. The existing "any later switch naming defproducerl => expect a" logic catches exactly this cascade; an anchor keyed only off the first switch line's producer does not.

Suggested amendment -- anchor the switch-line search by the abandoned block's number (only partition-fork blocks are past firstIProdBlockNum) and keep the escape as "ANY anchored switch line naming defproducerl", re-evaluated at assert time since the cascade lines land well after the first switch:

```python
switchForkPattern = re.compile(r"switching forks from \w+ \(block number (\d+) (defproducer[jkl])\)")
def partitionForkSwitches():
    """Producers of the fork blocks node_03 abandoned, in log order (partition fork only)."""
    return [m.group(2) for line in node3.linesInLog("switching forks")
            if (m := switchForkPattern.search(line)) and int(m.group(1)) > firstIProdBlockNum]

switchedFrom = Utils.waitForObj(lambda: partitionForkSwitches() or None, timeout=forkSwitchLogTimeout)
assert switchedFrom, "Expected to find 'switching forks' from a node_03 producer in node_03 log"
...
expectedProducers = {"defproducerk", "defproducerl"}
if "defproducerl" in partitionForkSwitches():   # re-check: cascade switches land after the first wait
    expectedProducers = {"defproducera"}
```

(needs `import re`)

Validated: replaying this logic offline against both failed runs' node_03 logs produces the correct expectation for each ({k, l} for the first, {a} for the second), and an end-to-end run with the combined version passes.

heifner (commented):
Re-reviewed the follow-up commit. The anchored partitionForkSwitches implementation matches what the earlier review validated: the regex agrees with the controller.cpp fork-switch log format, linesInLog/waitForObj semantics check out, and the block-number anchor correctly excludes pre-partition micro-forks (their abandoned blocks are numbered below firstIProdBlockNum) while catching both the convergence switch and late cascade switches. One minor comment below.

tests/nodeop_late_block_test.py
```diff
@@ -166,17 +176,18 @@ def findApplyDuringProducing():
     # verify that defproducerk or defproducerl blocks made it into the canonical chain
     # It can take a while to resolve the fork, but should have at least one block from node_03's
     # producers unless defproducera wins the fork
-    expectedProd = "defproducerk"
-    if node3.findInLog("switching forks .* defproducerl", switchForkLineNum):
-        expectedProd = "defproducera"
-    iProdBlockNum += 12 # into the next set of blocks
+    expectedProducers = {"defproducerk", "defproducerl"}
+    if "defproducerl" in partitionForkSwitches():
```
heifner, line 180: Calling partitionForkSwitches() again here rather than reusing switchedFrom is load-bearing (cascade switches land after the first wait returns), but nothing in the code says so, and the natural cleanup of reusing switchedFrom would silently reintroduce the cascade failure mode. Worth carrying the comment over:

```suggestion
    if "defproducerl" in partitionForkSwitches():  # re-check: cascade switches land after the first wait
```

heifner (approved):

heifner (approved):

PR #473, opened 2026-07-02
Test: share the state-history-stride boundary guard across ship-log tests

heifner:
## Summary

`ship_log_util_test` failed on the PR #461 gcc job (run 28591056813): the producer settled on block 200, an exact `--state-history-stride 50` multiple. SHiP log rotation had just moved every entry into `retained/` (last bundle `chain_state_history-151-200`), leaving the head `chain_state_history.{log,index}` at 0 bytes, so the first `info` check found an empty log and aborted in `parseInfoBlocks`.

This is the same boundary hazard PR #469 fixed for `ship_restart_test`, which stops the producer and operates on the head SHiP files the same way.

## Fix

Add `Node.waitForBlockClearOfStride(stride)` and call it before stopping the producer in both `ship_log_util_test` and `ship_restart_test`, replacing #469's inline guard so the boundary calculation lives in one place. The helper advances the head to a safe offset past the boundary — with a margin on both sides to absorb the couple of blocks produced while the pause takes effect — so the head log/index always have entries when the node stops.

### DIFF
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -281,6 +281,22 @@ class Node(Transactions):
                 return False
         return True
 
+    def waitForBlockClearOfStride(self, stride, lowerMargin=5, upperMargin=10, timeout=None):
+        """Advance the head clear of a state-history-stride rotation boundary before stopping.
+
+        When the last stored block is exactly a stride multiple, SHiP log rotation has just moved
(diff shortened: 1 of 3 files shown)


huangminghuang (approved):
Reviewed PR #473 at ab9fc5cd05. No actionable findings.

PR #474, opened 2026-07-02
Opp: bind each batch-relayed outpost to its own remote identity

heifner:
Fixes SEC-37 / WSA-075 on the wire-sysio relay side.

The batch operator built every EVM outpost job from one global `eth_client_id` / `eth_opp_addr` / `eth_opp_inbound_addr` tuple, and every SVM job from one `sol_program_id`. `sysio.chains` allows multiple active non-WIRE rows of the same `ChainKind`, so two active EVM (or two active SVM) chains collapsed onto a single remote endpoint: one outpost's envelopes could be delivered under another's `chain_code`, counting as consensus evidence for the wrong chain, or stalling epoch advancement when the shared endpoint lacked a given chain's data. We need to support multiple EVM chains, so the registry-level "reject a second same-kind outpost" shortcut is out; the fix binds each chain to its own exact remote identity.

**Per-chain-code operator binding (`--batch-outpost`).** Each operator supplies the remote OPP contract identity per chain code with the repeatable `--batch-outpost` option, `CHAIN_CODE,opp_addr[,opp_inbound_addr]` — EVM rows require the OPP and OPPInbound contract addresses (0x-hex), SVM rows only the outpost program id (base58). Keying the binding by chain code rather than by RPC client keeps two WIRE rows on the same remote network distinct. Malformed specs and duplicate chain codes refuse startup. The binding intentionally stays operator-local config rather than on-chain state: WIRE cannot verify a remote address either way, envelope integrity already rests on majority epoch consensus across independently-configured operators, and the contract-side defense — each remote outpost verifying the WIRE chain code it is addressed with — is tracked as SEC-108.

**Per-chain relay (`batch_operator_plugin`).** `build_opp_jobs` resolves each active row's binding by chain code and auto-selects the Ethereum RPC client whose configured chain id matches the row's `external_chain_id` (the `--outpost-ethereum-client` spec already carries an optional chain id; `outpost_ethereum_client_plugin` gains a `get_client_by_chain_id` lookup that returns a unique match or nothing). Any active outpost with no binding, a partial binding, or no matching client is skipped with a warning — fail-closed — so it is never relayed through another chain's endpoint. For SVM the shared RPC client is retained (one Solana cluster serves every program) and only the program id becomes per-row.

**Breaking launch-config change.** The global `batch-eth-client-id`, `batch-eth-opp-addr`, `batch-eth-opp-inbound-addr`, and `batch-sol-program-id` options are removed (the collapse originated there). Deployments must instead register one `--outpost-ethereum-client` per EVM chain (with its chain id) and one `--batch-outpost` binding per active chain row. `batch-sol-client-id` is kept for the Solana RPC connection.

Tests: the spec parser lives in `outpost_binding.hpp` and is covered in `test_batch_operator_plugin` (EVM/SVM accept paths; arity, empty-field, and bad-chain-code rejects; option registration). `sysio.chains` and the contract test suites are unchanged from master.


huangminghuang (approved):
Reviewed the current diff and found no blocking issues. The per-chain outpost binding and fail-closed client selection look consistent with the intended SEC-37/WSA-075 fix.

PR #475, opened 2026-07-02
Chain: cap get_table_rows page size at max_return_items under a finite deadline

heifner:
## Summary

The get_kv_rows / get_table_rows unification (`9d19c546c8`) dropped the per-page row cap that `get_table_by_scope` still applies. Since then a single `get_table_rows` page scanned on the caller's raw `limit` with no `max_return_items` bound — staging full KV row buffers (up to the 1 MiB consensus row limit each) into `raw_row` vectors before the response byte budget is enforced. A small unauthenticated request could drive CPU/memory response materialization past the intended per-page bound.

This restores the clamp `get_table_by_scope` uses, applied to the resolved per-page limit when the scan is under a finite deadline:

- The in-process `all_rows` full-table walk stays exempt (`params_deadline` is `maximum()` by construction), so the batch-operator and underwriter cron plugins are unaffected — every in-process caller uses `limit = 1`, `find` (limit → 1), or `all_rows`.
- An operator-configured unlimited `http-max-response-time-ms` (`-1`) stays exempt, matching `get_table_by_scope`.
- Callers recover the remaining rows through the existing `more` / `next_key` pagination, so no data becomes unreachable.

A new `table_reader_tests` case pins both branches of the guard: a finite deadline caps a large `limit` at `max_return_items` with `more`/`next_key` set, while an unlimited deadline returns every row uncapped.

### DIFF
--- a/plugins/chain_plugin/src/chain_plugin.cpp
+++ b/plugins/chain_plugin/src/chain_plugin.cpp
@@ -2259,6 +2259,14 @@ read_only::get_table_rows( const read_only::get_table_rows_params& p, const fc::
    std::string_view ub_sv(ub_bytes.data(), ub_bytes.size());
 
    uint32_t limit = effective_limit;
+   // Bound the per-page row count exactly as get_table_by_scope does: when the scan is subject to a finite
+   // deadline, clamp the caller's limit to api_base::max_return_items so a single HTTP page cannot stage an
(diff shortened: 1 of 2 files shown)


huangminghuang (approved):
Reviewed the diff and found no blocking issues. git diff --check passed; local C++ test binaries were not run.

PR #476, opened 2026-07-02
Signature-provider: harden default key-file permissions and redact keys from logs

heifner:
## Summary

`signature_provider_manager_plugin` mishandled signing-key secrets in two ways:

- `save_default_signature_provider_specs()` wrote `default_signature_providers.json` — which holds auto-generated WIRE/BLS private keys in `KEY:<private-key>` form — via `fc::cfile::truncate_rw_mode` with no permission hardening, so the file was created with default umask permissions (typically `0644`, group/world readable). The node auto-creates this file without operator involvement, so the exposure is easy to miss.
- `plugin_initialize` debug-logged the full `--signature-provider` spec, which for `KEY:` providers contains the inline private key (any log aggregator then receives it).

This change:

- Restricts the persisted key file to owner-only (`0600`) via a `restrict_file_to_owner()` helper, applied **on the freshly-created empty file, before the secret is written**, so the key material is never momentarily written to a group/world-readable file.
- Redacts the inline `KEY:<private-key>` field from the spec debug-log through a new `redact_signature_provider_spec()` helper. Only the final comma-separated provider field is inspected and only the `KEY:` form is masked, so `KIOD:`/other providers (which reference external material), public keys, and a `KEY:`-prefixed name are left untouched.

Signing behavior is unchanged — keys still load and sign identically; they are simply no longer world-readable or logged. The plugin's other log sites already log only the key type and the name/public-key, so the redacted spec log was the only leak.

Covered by a redaction unit test (CSV/bare/KIOD/false-positive-name cases) and a permission integration test asserting the persisted key file is owner-only.

### DIFF
--- a/plugins/signature_provider_manager_plugin/src/signature_provider_manager_plugin.cpp
+++ b/plugins/signature_provider_manager_plugin/src/signature_provider_manager_plugin.cpp
@@ -10,7 +10,7 @@
 #include <fc/io/fstream.hpp>
 #include <fc/io/json.hpp>
 #include <fc-lite/algorithm.hpp>
-#include <gsl-lite/gsl-lite.hpp>
+#include <fc/io/secure_file.hpp>
 
 #include <filesystem>
 
@@ -30,9 +30,11 @@ std::filesystem::path default_signature_provider_spec_file() {
 /**
  * Restrict a file to owner read/write only (0600).
  *
- * Used for files that persist private signing keys so they are not left group- or world-readable under the
- * process umask. Best-effort: a failure to set permissions is logged, not fatal, so key persistence still
- * succeeds on platforms/filesystems that do not support POSIX permission bits.
+ * Used to bring pre-existing key files that predate the owner-only hardening (and were therefore created
+ * group/world readable under the process umask) down to owner-only when they are loaded. New key files are
+ * instead written through `fc::write_secure_file`, which creates them owner-only from the start. Best-effort:
+ * a failure to set permissions is logged, not fatal, so loading still succeeds on platforms/filesystems that
+ * do not support POSIX permission bits.
  *
  * @param file_path the file to restrict; it must already exist.
  */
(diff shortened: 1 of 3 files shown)


plugins/signature_provider_manager_plugin/src/signature_provider_manager_plugin.cpp
```diff
@@ -209,6 +230,9 @@ class signature_provider_manager_plugin_impl {
          fc::cfile file(def_sig_prov_file, fc::cfile::truncate_rw_mode);
          gsl_lite::final_action file_guard([&file]() { file.close(); });
 
+         // This file persists private signing keys. Restrict it to owner-only now, while it is still empty,
+         // so the key material is never momentarily written to a group/world-readable file under the umask.
+         restrict_file_to_owner(def_sig_prov_file);
```
huangminghuang: #464 is already in the base and provides fc::write_secure_file(), but this path still uses fc::cfile plus a local chmod helper for default_signature_providers.json. Please use the secure helper for new writes so this secret file gets the owner-only temp file, atomic publish, fsync, and symlink/non-regular-target protections from #464. This also needs to cover existing files: if an old 0644 default_signature_providers.json is loaded and no new default key is generated, changed stays false and the file is never rewritten or restricted. Please restrict or rewrite the loaded file so existing default key files are brought down to owner-only permissions too.

huangminghuang (commented):
Security review comments.

plugins/signature_provider_manager_plugin/src/signature_provider_manager_plugin.cpp
```diff
@@ -423,7 +447,7 @@ void signature_provider_manager_plugin::plugin_initialize(const variables_map& o
    if (options.contains(option_name_provider)) {
       auto specs = options.at(option_name_provider).as<std::vector<std::string>>();
       for (const auto& spec : specs) {
-         dlog("Registering signature provider from spec: {}", spec);
+         dlog("Registering signature provider from spec: {}", redact_signature_provider_spec(spec));
```
huangminghuang, line 458: This debug log now redacts KEY: providers, but create_provider() still formats the raw spec in the invalid comma-count assertion. A mistyped --signature-provider value that still contains KEY:<private-key> can therefore leak the private key through startup error logs. Please use redact_signature_provider_spec(spec) there too, or avoid echoing the provider field in that error.

plugins/signature_provider_manager_plugin/src/signature_provider_manager_plugin.cpp
heifner: Done in 158d3bc7e4: `save_default_signature_provider_specs` now writes through `fc::write_secure_file` (owner-only temp file, atomic rename into place), and `load_default_signature_provider_specs` restricts any pre-existing group/world-accessible file to owner-only before its specs are registered, so legacy 0644 files are brought down even when no new default key is generated and nothing rewrites the file. Added a test that seeds a 0644 key file and verifies it is owner-only after load with the content unchanged.

plugins/signature_provider_manager_plugin/src/signature_provider_manager_plugin.cpp
heifner: Done in 158d3bc7e4: the comma-count assertion now formats `redact_signature_provider_spec(spec)`, and a test asserts the error detail carries `KEY:<redacted>` and not the private key.

huangminghuang (approved):
Re-reviewed latest head. The previous findings are addressed; no remaining code findings from me.

PR #478, opened 2026-07-03
Test: Decide late-block producer window expectation from window content

heifner:
The `nodeop_late_block_test` assertion flipped its expected window producers to `defproducera` whenever any post-partition fork switch abandoned a `defproducerl` block. That condition also matches a routine slot-handoff micro-fork past the verification window - `defproducera` winning the boundary race against node_03's last `defproducerl` block when the block doesn't propagate through the 3-hop bridge path in time (observed in [this CI run](https://github.com/Wire-Network/wire-sysio/actions/runs/28664218404/job/85011947902): node_03 logged `switching forks from ... 299 defproducerl ... to ... 299 defproducera` while the window correctly held defproducerk/l blocks).

Decide from the window content instead: the check passes when the window holds any defproducerk/l block, and an all-defproducera window is accepted only when node_03's log shows a fork switch abandoning one of its own defproducerl blocks (the slow-reconnect case where node_03 burns its k and l slots on the abandoned fork). Verified locally under the UBSAN build across both shapes: normal recovery (window holds k/l) and late recovery (cascade `l#272 -> a#251`, window held defproducera).

### DIFF
--- a/tests/nodeop_late_block_test.py
+++ b/tests/nodeop_late_block_test.py
@@ -173,21 +173,27 @@ try:
         assert block["producer"] == "defproduceri", \
             f"expected defproduceri for block {canonicalBlock['block_num']}, instead: {block['producer']}"
 
-    # verify that defproducerk or defproducerl blocks made it into the canonical chain
-    # It can take a while to resolve the fork, but should have at least one block from node_03's
-    # producers unless defproducera wins the fork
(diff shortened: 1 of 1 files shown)


huangminghuang (approved):
Reviewed current head. No code findings from me; the producer-window logic keeps the intended coverage while avoiding the false expectation flip.

PR #479, opened 2026-07-03
Prometheus: bound p2p per-connection metric cardinality

heifner:
## Summary

Ultimately, `prometheus_plugin` may go away, but for now...

`prometheus_plugin` subscribes to net_plugin's P2P connection snapshots and, for each peer, adds ~18 `nodeop_p2p_*` gauge series keyed by a `connid` label derived from the peer's handshake node id (`node_id.short_id()`). It never removes those series when a peer disconnects, and the default P2P `allowed-connection = any` accepts peers without authentication. An unauthenticated peer that reconnects with fresh node ids therefore piles up ~18 stale series per generation, and every `/v1/prometheus/prometheus` scrape then serializes the whole retained registry — growing scrape CPU, memory, and response size over time.

This re-keys the per-connection series by the composite **`{remote_ip, connection_id}`** — the real socket peer IP (canonical; a completed TCP handshake proves the source, so minting N series needs N real IPs) and the node's own monotonic connection counter — neither of which the peer supplies in its handshake. It also prunes series for connections that are no longer in the snapshot each update, so the registry tracks live connections instead of accumulating. Removal is safe against concurrent scrapes because prometheus-cpp's `Family` guards its metric map with a mutex.

The redundant `nodeop_p2p_addr` (the IP is now a label on every series) and the unused `nodeop_p2p_peer_addr` families are removed; all other per-peer telemetry is preserved.

**Operator note — metric label schema change:** per-peer series that were keyed `connid="<node id>"` are now keyed `remote_ip="…"` / `connection_id="…"`, and `nodeop_p2p_addr` / `nodeop_p2p_peer_addr` no longer exist. Dashboards keyed on `connid` need updating. The old `connid` changed on every reconnect (so it was never stable for tracking a peer), whereas `remote_ip` is stable across a peer's reconnects.

A new `test_prometheus_plugin` target (auto-discovered by the `plugin_target` macro) covers it: one case asserts the series carry the local `{remote_ip, connection_id}` labels and none of the peer-supplied node id / advertised address; another asserts a connection dropping out of the snapshot has its series pruned.

### DIFF
--- a/plugins/prometheus_plugin/src/metrics.hpp
+++ b/plugins/prometheus_plugin/src/metrics.hpp
@@ -10,6 +10,13 @@
 #include <prometheus/registry.h>
 #include <prometheus/text_serializer.h>
 #include <fc/log/logger.hpp>
+
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
 namespace sysio::metrics {
 
 struct catalog_type {
@@ -44,7 +51,6 @@ struct catalog_type {
       Gauge& num_peers;
       Gauge& num_clients;
 
-      prometheus::Family<Gauge>& addr; // Empty gauge; ipv6 address can't be transmitted as a double
       prometheus::Family<Gauge>& port;
       prometheus::Family<Gauge>& connection_number;
       prometheus::Family<Gauge>& accepting_blocks;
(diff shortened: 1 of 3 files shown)


huangminghuang (approved):
Reviewed current head. No code findings from me; the P2P Prometheus series are now keyed by local connection identity and stale series are pruned.

PR #482, opened 2026-07-06
Underwriter: contain deferred-startup escapes and report the LIB gap on the sync gate

heifner:
Hardening follow-ups to the underwriter sync gate, from the PR #477 review (applies on top of the LIB-gate/preflight-retry follow-up).

**Contain deferred-startup escapes, with a boot/posted policy split.** The deferred startup body runs from tasks posted to the app executor — the sync-gate callback's post and the preflight retry timer — where an escaping exception unwinds into `application::exec()`, which `quit()`s and shuts down every plugin: a mid-operation teardown of the whole node, with `/v1/underwriter/*` still reporting `waiting_for_sync`. Escapes were possible from a throwing table decode in the preflight or registry read (both sat outside any handler), a non-fc exception from outpost client wiring, and cron `add_job` (outside the try entirely). Posted contexts now contain any escape as a terminal `startup_failed` gate state with a `detail` payload, using `FC_LOG_AND_DROP` — the tree's containment idiom, which deliberately rethrows `boost::interprocess::bad_alloc` so chainbase shared-memory exhaustion stays fatal. The synchronous `plugin_startup` path deliberately does not contain: a node that would boot without underwriting exits instead, fail-fast and supervisor-visible. Terminal states are structurally protected — `gate_state = active` is stored immediately after the scan cron is added (a later throw cannot leave the cron running under a terminal gate state), and the preflight retry task proceeds only while `preflight_retrying` is the live state (a stale retry can never overwrite a terminal state).

**Report the gate's actual criterion while waiting.** Since the gate arms on last-irreversible recency, the `waiting_for_sync` payload's head-based `head_behind_sec` could read ~0 on a finality-stalled node whose gate correctly stays closed. The payload now carries `head_behind_sec` and `lib_behind_sec`; durations travel through the plumbing as `fc::microseconds` (the absent-irreversible-root case a typed empty optional) and integers materialize only at the JSON boundary. `lib_behind_sec` is omitted while no fork-db root exists — an in-band -1 sentinel would collide with real negative clock-skew gaps after whole-second truncation — and negative gaps are reported as-is, since the skew is itself diagnostic signal. The chain reads happen only for the `waiting_for_sync` state; terminal responses do no chain access.

**Cleanups.** The `startup_armed` flag is removed: every armed attempt stores a non-waiting `gate_state` before returning, so `startup_attempted()` derives arming from the gate state, and an assert after the containment wrapper turns that invariant into a tripwire. The JSON field keys shared across the payload builder, the post-startup response builders, and the tests move behind `field::` constants in `sync_detail.hpp`, pinned once to their literal wire spellings via static_asserts so builder and tests cannot drift in tandem; a compile-time-pinned `active_status` constant replaces the per-request `active_status_name()`. `head_is_recent` is renamed `block_time_is_recent` — its only production caller feeds the last-irreversible block's time, and the head-named identifier sat exactly on the head-vs-LIB distinction the gate fix exists for. The `sync_detail.hpp` header docs drop the inaccurate "same recency notion producer_plugin uses" claim and describe the LIB criterion the gate actually measures.

Tests: `underwriter_sync_tests` covers the `startup_failed` payload, the omitted-key no-root case, negative skew gaps passing through, whole-second truncation, and both behind-gap fields; wire keys are pinned by static_asserts while the `status` values stay literal in the cases to pin the wire format.

### DIFF
--- a/plugins/underwriter_plugin/include/sysio/underwriter_plugin/sync_detail.hpp
+++ b/plugins/underwriter_plugin/include/sysio/underwriter_plugin/sync_detail.hpp
@@ -6,11 +6,12 @@
  *
  * The underwriter's startup preflight validates depot-side state (opreg
  * registration, chain registry, authex links) via LOCAL table reads. On a
- * cold-booting operator node those reads see mid-replay (possibly genesis)
+ * cold-booting operator node those reads see mid-sync (possibly genesis)
  * state and fail spuriously, so the plugin must not begin underwriting until
- * the chain plugin reports the node synced. "Synced" here is the same recency
- * notion producer_plugin uses to decide it may produce: the head block time is
- * within a small window of wall-clock now.
+ * the node is synced. "Synced" here means the LAST IRREVERSIBLE block's time
+ * is within a small window of wall-clock now — the irreversible state is what
+ * the plugin's table reads actually serve (operator daemons run read-mode =
+ * irreversible).
  */
 
 #include <fc/time.hpp>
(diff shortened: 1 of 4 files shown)


huangminghuang (approved):
Reviewed the underwriter sync-gate hardening and diagnostics changes. No actionable findings from my pass; required CI is green.

PR #483, opened 2026-07-06
Msgch: enforce the 64 KiB envelope cap on inbound deliver

heifner:
The outbound `buildenv` packer keeps every envelope within `MAX_ENVELOPE_BYTES` (64 KiB — the e2e-supported maximum, mirrored by the Ethereum and Solana outposts), but inbound `deliver` never applied the same bound: it authenticated the operator, decoded the envelope, checked the epoch and duplicates, then hashed and stored the raw bytes at any size the generic chain ceilings admit (`max_inline_action_size` ~512 KiB, `max_kv_value_size` 256 KiB). A scheduled ACTIVE batch operator could therefore persist inbound envelopes WIRE's own packer could never emit — asymmetric cross-chain size semantics plus avoidable envelope-table, consensus, and dispute-state pressure.

`deliver` now rejects `data.size() > MAX_ENVELOPE_BYTES` right after the empty-data check — before the chains lookup, protobuf decode, and sha256 — so oversized deliveries revert as cheaply as possible.

Tests (`sysio.dispatch_tests`):
- `deliver_oversized_envelope_reverts` — a decodable current-epoch envelope at cap+1 reverts and no `envelopes` row lands.
- `deliver_envelope_at_size_cap_succeeds` — an envelope exactly at the cap is accepted, stored, and dispatches.
- `encode_envelope_padded_to` builds decodable envelopes at an exact encoded size (probe-then-shrink padding; exact near the cap because every nested length varint sits in the same 3-byte band there).
- Corrects two stale comments describing same-operator duplicate delivery as silently dropped — it reverts as a duplicate.


huangminghuang (changes requested):
contracts need to be regenerated

huangminghuang (approved):

PR #484, opened 2026-07-06
Fix macOS auto BP gossip CPU budget flake

huangminghuang:
## Summary
- Keep the generated genesis CPU budgets unchanged for `auto_bp_gossip_peering_test` and all other `Cluster.launch` callers.
- Make `Transactions.pushMessage` accept a configurable `cpuRetryAttempts` value while preserving the existing default of 5 attempts.
- Have `Cluster.setFinalizers` use a larger retry window for transient `tx_cpu_usage_exceeded` failures during instant-finality bootstrap.

## CI failure
- Diagnosed from failed CI run: https://github.com/Wire-Network/wire-sysio/actions/runs/28735544717
- Failure was isolated to `auto_bp_gossip_peering_test` in the scheduled macOS arm64 sharded NP/LR step.
- The failing bootstrap action was `sysio::setfinalizer`, which intermittently exceeded subjective transaction CPU on the runner.

## Review notes
- Addresses review feedback to retry the transaction instead of increasing the max CPU budget.
- The current PR diff no longer changes `auto_bp_gossip_peering_test.py` or adds any genesis CPU-budget override API.

## Validation
- `python3 -m py_compile tests/TestHarness/Cluster.py tests/TestHarness/transactions.py tests/auto_bp_gossip_peering_test.py`
- `git diff --check origin/master...HEAD`

## Not run
- Full `auto_bp_gossip_peering_test`; this worktree does not have the macOS arm64 build/test artifacts needed to reproduce the scheduled CI shard locally.

### DIFF
--- a/tests/TestHarness/Cluster.py
+++ b/tests/TestHarness/Cluster.py
@@ -67,6 +67,9 @@ class Cluster(object):
     __finalizerCatchupMaxLagBlocks=2
     __finalizerCatchupStableRounds=3
     __finalizerCatchupReportInterval=5
+    # Default generated-genesis CPU limits used by Cluster.launch.
+    __defaultMaxBlockCpuUsage=400000
+    __defaultMaxTransactionCpuUsage=375000
 
     # pylint: disable=too-many-arguments
     def __init__(self, localCluster=True, host="localhost", port=None, walletHost="localhost", walletPort=None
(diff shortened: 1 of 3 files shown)


heifner (changes requested):
We should not increase the max cpu budget. Instead the transaction should be retried.

heifner (approved):

PR #486, opened 2026-07-06
Contracts: bound the per-epoch withdraw and swap-from-WIRE queue drains

heifner:
`sysio.epoch::advance` drains the opreg withdraw queue (`flushwtdw`) and the uwrit swap-from-WIRE queue (`drainfwq`) as inline actions, so both run inside advance's single transaction and share its CPU deadline (~150 ms — a hard, uncatchable wall). Neither drain was bounded: `flushwtdw` processed every matured row, and `drainfwq` copied the entire queue into a vector and processed all of it.

An operator splitting collateral into many one-unit withdraws (SEC-78), or a caller splitting escrowed WIRE into many from-WIRE swaps (SEC-77), could grow either queue until the drain exceeds the deadline. Because advance re-runs deterministically on the same queue content, the aborted transaction rolls back all progress and re-aborts on every retry — a permanent epoch stall, not just elevated resource use.

Merge with: Wire-Network/wire-libraries-ts#16 and Wire-Network/wire-tools-ts#24 — order in Cross-repo below.

## Fix

Both drains now process at most a fixed number of rows per advance (`MAX_WTDW_FLUSH_PER_EPOCH` / `MAX_FWQ_DRAIN_PER_EPOCH`, both 32) and leave the remainder queued for the next advance:

- Undrained withdraws keep their collateral in the operator's balance; undrained swaps keep their WIRE escrow in reserve custody — so carrying the remainder over loses nothing.
- `drainfwq` additionally drops its full-table pre-copy and walks the queue front-first (FIFO, erase-as-you-go), so its scan cost is bounded too.
- Both constants are conservatively sized to stay well under the transaction CPU budget shared with the rest of advance's fan-out; raise via contract upgrade if legitimate throughput ever approaches them.

Ingress is deliberately left to the existing economic throttles rather than a per-caller row cap: every `swapfromwire` is a `require_auth(user)` transaction that bills CPU/NET and escrows real WIRE, and withdraw rows can never exceed an operator's real deposited collateral. The bounded drain is the liveness guarantee; a count cap would mostly risk rejecting legitimate high-volume callers.

## Tests

- `flushwtdw_bounds_rows_per_epoch` (opreg): queues 40 withdraws, terminates so every matured row takes the erase branch, flushes → exactly 32 drain, 8 remain, second flush clears them.
- `drainfwq_bounds_rows_per_epoch` (dispatch): queues 40 from-WIRE swaps against a funded reserve, drains → exactly 32, 8 remain, second drain clears them. Also the first end-to-end coverage of draining a populated swap-from-WIRE queue.

ABI changes (from round 2 below): `sysio.uwrit` — `setconfig` and the `uwconfig` row gain `min_fromwire_amount` + `fromwire_revert_fee_bps`; `sysio.reserv` — `refundwire` gains `revert_fee_bps`. Regenerated `.abi`/`.wasm` are committed (`sysio.opreg` is a wasm-only rebuild, no schema change), and `SysioContractTypes` is regenerated in Wire-Network/wire-libraries-ts#16 (see Cross-repo). No snapshot/deep-mind reference data is affected.

Closes SEC-77 and SEC-78.

## Review round 2 — ingress economics (floor + revert fee)

Review flagged the residual left by the drain bound alone: dust rows escrow nothing meaningful, revert at drain (zero quote / missed variance), and refund in full — so spam could hold the 32-slot/epoch drain budget and grow system-paid queue RAM while recycling the same capital. Rather than the count caps considered and dropped above, this prices the queue slots:

- `swapfromwire` enforces a configurable minimum escrow (`uwconfig.min_fromwire_amount`, default 5 WIRE): sustaining a D-row backlog locks D x floor WIRE for the full queue latency the spammer creates, and floor-sized escrows stay priceable by the Bancor kernel (no zero-quote revert class).
- Drain-time reverts produced by the caller's own parameters (unpriceable target, variance exceeded) forfeit `uwconfig.fromwire_revert_fee_bps` (default 10 bps, placeholder pending fee calibration — retunable via `setconfig` without an upgrade), routed through the standard rewards/emissions split via `reserv::refundwire`'s new `revert_fee_bps` argument. System-caused reverts (chain deregistered, reserve deactivated or flipped private, internal encode failure) and uwreq-stage rejections refund in full.
- `setconfig` gains both knobs with guards: a positive floor, and a sub-100% revert fee so the post-fee refund transfer can never zero out inside the never-throw drain.

Tests cover the floor boundaries at default and lowered config, the fee'd variance revert (user net, rewards-bucket accrual, odd-split rounding), full refund on a system-caused revert, and the `setconfig` guard boundaries.

## Cross-repo

Merge order: this PR, then Wire-Network/wire-libraries-ts#16 (regenerated `SysioContractTypes` carrying the new `setconfig`/`refundwire` fields + a transitional `SystemContracts` alias), then Wire-Network/wire-tools-ts#24 (cluster bootstrap sends the new fields; dev floor pinned to the flows' 0.1 WIRE escrow; dev-node resource-monitor hardening).


contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -1624,16 +1624,27 @@ void uwrit::drainfwq() {
...
+   // walks the queue forward and the next advance resumes where this one
+   // stopped; undrained rows keep their WIRE escrow in reserve custody and
+   // drain a later epoch. Bounding this loop is what keeps `advance` inside its
+   // hard, uncatchable CPU deadline: an unbounded drain over an attacker-
+   // inflated queue would abort every advance and stall epoch progress chain-
+   // wide. `drained` is incremented before any of the per-row `continue`
+   // branches, so it counts every attempt (refund-and-drop included). Reading
+   // only the front row (not a full-table pre-copy) also bounds the scan cost.
+   uint32_t drained = 0;
+   while (drained < MAX_FWQ_DRAIN_PER_EPOCH) {
```
huangminghuang, line 1677: [P2] This cap keeps `advance` bounded, but the queue is still global FIFO with uncapped, system-paid rows. A caller can enqueue many tiny `swapfromwire` rows and monopolize all `MAX_FWQ_DRAIN_PER_EPOCH` slots for many epochs, delaying legitimate swaps while growing `sysio` RAM. Consider adding an outstanding-row cap, actor-paid RAM, per-user/fair scheduling, or a minimum amount calibrated to row cost and drain throughput.

huangminghuang (commented):
Posted inline review comments on the bounded drain loops.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -1074,13 +1074,21 @@ void opreg::flushwtdw(uint32_t current_epoch) {
...
+   // MAX_WTDW_FLUSH_PER_EPOCH rows per advance (SEC-78): the remaining matured
+   // rows flush on the next advance, which keeps this epoch-inline action inside
+   // the transaction CPU deadline it shares with the rest of advance's fan-out.
+   // `flushed` is incremented at the top of every iteration, before any of the
+   // per-row `continue` branches, so it counts every attempt.
+   uint32_t flushed = 0;
    auto it = idx.begin();
-   while (it != idx.end() && it->eligible_at_epoch <= current_epoch) {
+   while (it != idx.end() && it->eligible_at_epoch <= current_epoch &&
+          flushed < MAX_WTDW_FLUSH_PER_EPOCH) {
```
huangminghuang, line 1086: [P2] This fixes the hard epoch-stall path, but a single operator can still split collateral into many tiny withdraw rows and consume the global flush budget for many epochs. Because these rows are system-paid and `flushwtdw` is FIFO by eligibility, that can starve other operators' matured withdraws and grow `sysio` RAM. Consider a per-operator outstanding-row cap, actor-paid RAM, fair scheduling, or a minimum withdraw size tied to row/drain cost.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Addressed the economics in 4cf00823d9: swapfromwire now enforces a configurable minimum escrow (5 WIRE default), and drain-time reverts the caller's own parameters produce (zero quote, missed variance) forfeit a configurable fee routed like the settlement fee. That prices the queue slots directly — a sustained D-row backlog now locks D x 5 WIRE, and revert churn pays the system instead of recycling for free. System-caused reverts (reserve deactivated/private-flipped/deregistered after enqueue) still refund whole.

On the other suggested mechanisms: outstanding-row caps were in an earlier draft and dropped deliberately — per-user caps are sybil-bypassable and a global cap would let a spammer block enqueue for everyone, a worse DoS than delayed drains. Actor-paid RAM would depart from the platform RAM model (user-facing rows are sysio-paid platform-wide; ordinary accounts hold no billable RAM), and queue RAM is reclaimed at drain — growth reverses as the queue drains. Fair scheduling needs per-user state inside an epoch-inline never-throw action and is equally sybil-bypassable. Worth revisiting if the floor + fee prove insufficient in practice.

contracts/sysio.opreg/src/sysio.opreg.cpp
heifner: Same analysis as the fwqueue thread on caps / actor-paid RAM / scheduling. Two things lower this queue's residual further: the actor set is registered operators — role-minimum collateral, require_auth ingress, terminable — not the public; and a griefing operator is fully attributable, with termination cutting off further enqueue while their queued rows die without remit as they cycle through the flush. Dust-splitting within an operator's available balance remains possible, so if the swap-side floor pattern proves out we can apply the same shape here (minimum withdraw with a full-remaining-balance escape); left unchanged for now on the accountable-set rationale.

huangminghuang (approved):
Approved. With the clarified assumption that this is still pre-production/fresh-deployment work, I do not see any blocking code issues in the bounded drains, from-WIRE ingress floor, or revert-fee accounting. One small non-blocking cleanup: the PR body still has an older note saying ABIs are unchanged, but the later commit updates setconfig/refundwire/uw_config.

PR #487, opened 2026-07-07
Test: SHiP restart from snapshot with pre-existing state-history data

heifner:
Adds coverage to `ship_restart_test.py` for restarting a SHiP node from a snapshot while state-history data already exists, in both directions:

- **Snapshot older than the SHiP head** — the node starts, the overlapping blocks replayed from the block log are no-op rewrites of identical block ids, the pre-existing log and index survive byte-for-byte as a prefix of the grown files, and new blocks append contiguously.
- **Snapshot newer than the SHiP head** — the node refuses to start with the `skips over block` gap error instead of leaving a hole in the served history. This is the failure mode recently hit by a testnet SHiP node whose state-history logs ended well before the head of the snapshot it was restarted from.

The log-layer behavior for both cases is already covered by the `rewrite_same` family and the skip check in `tests/ship_log.cpp`; this adds the end-to-end path through a real `--snapshot` startup with the state_history_plugin enabled.

Also launches the cluster with `biosFinalizer=False`. The finalizer policy previously included the bios node with threshold 2, so shutting down the "unneeded" bios node permanently halted LIB advancement. The existing parts never noticed because they only assert on head movement, but the new part needs finality for snapshot completion (a pending snapshot is only written once its block becomes irreversible).

### DIFF
--- a/tests/ship_restart_test.py
+++ b/tests/ship_restart_test.py
@@ -6,13 +6,20 @@ import shutil
 import signal
 
 from TestHarness import Cluster, TestHelper, Utils, WalletMgr
+from TestHarness.Node import BlockType
 
 ###############################################################################
 # ship_restart_test
-# 
+#
 # This test verifies SHiP shuts down gracefully or recovers when restarting
 # with various scenarios of corrupted log and/or index files.
-# 
+#
+# It also verifies restarting from a snapshot with pre-existing SHiP data:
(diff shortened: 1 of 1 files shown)


huangminghuang (approved):
Reviewed the current diff and found no blocking issues. The added SHiP snapshot restart coverage looks consistent with the harness behavior and the intended gap/prefix scenarios.

PR #488, opened 2026-07-07
Feature: SSM signature provider

heifner:
**What the feature does**

Adds an `SSM:` signature-provider scheme alongside `KEY:`, `KIOD:`, and `KMS:`. The private key lives in AWS SSM Parameter Store as a KMS-encrypted SecureString; nodeop fetches it exactly once when the provider is created and signs locally thereafter. Semantically it is `KEY:` without the key material in config files, command lines, process listings, or shell history — access is IAM-gated and every decrypt is a CloudTrail event.

Because signing is local after the one startup fetch, `SSM:` works for every chain key type with a `KEY:` form (wire, wire_bls, ethereum, solana — the parameter value is exactly the string that would follow `KEY:`) and is fast enough for every signing path including producer block signing. That is why, unlike `KMS:`, nodeop registers it out of the box. Activation is solely the presence of an `SSM:` spec — registration alone is a registry entry with no AWS SDK init, threads, or network.

**Spec grammar**

`--signature-provider bp1,wire,wire,SYS…,SSM:us-east-1:/wire/prod/bp1` — shorthand `<region>:<parameter-name>` (everything after the first colon passes to GetParameter verbatim, so SSM's native `:version`/`:label` selectors work) or a full parameter ARN. Region is mandatory, matching `KMS:`.

**Behavior**

The fetch runs inside the spec handler during plugin initialization — a deliberate departure from the KMS provider's offline construction: the provider cannot exist without key material, and a boot-time failure with a precise error is the intended behavior for a misconfigured signer. Parameters not stored as SecureString are refused with the exact `put-parameter --type SecureString` remediation. The derived public key must match the spec's pinned key. Parse failures never echo the fetched value. Transient vs permanent AWS failures map to `signing_transient_exception` / `plugin_config_exception` via the SDK's own retry classification, permanent ones carrying an IAM hint (`ssm:GetParameter` + `kms:Decrypt`).

**Shared AWS glue**

The `kms/` sub-library's file-private SDK lifecycle, per-region client cache, and error classification are extracted into a `sigprov_aws_common` library both sub-libraries link — required for correctness, since two independent lifecycles in one binary would double-run `Aws::InitAPI`/`ShutdownAPI`. The KMS public API and message formats are unchanged. libfc gains runtime-dispatch native-string key parsers and the local-sign closure helper, shared by the plugin's `KEY:` path and the SSM handler.

**Scope notes**

nodeop links aws-sdk-cpp core+ssm on all platforms (~6 MB stripped; measured); sys-util and cranker-example do not register the scheme. `KMS:` remains unregistered in nodeop per the rationale in the original KMS work. vcpkg adds the `ssm` feature to the existing aws-sdk-cpp dependency; no registry or override changes. Tests: offline unit + plugin-routing suites via an injectable fetcher seam, an env-gated live round-trip (`SSM_LIVE_SPEC`/`SSM_LIVE_PUBKEY`), and an operator runbook under `ssm/test/README.md`.


huangminghuang (approved):
Approved. Non-blocking cleanup: the OVERVIEW wording says `SSM:` supports "all key types", but SUI is still explicitly unsupported. Consider phrasing that as "all KEY-supported key types" or listing `wire`, `wire_bls`, `ethereum`, and `solana` so the overview matches the implementation.

PR #490, opened 2026-07-07
Contracts: bound governance config inputs against overflow and safety-gate bypass

heifner:
System-contract config setters validated a lower bound but no upper bound. A privileged-but-mistaken governance action could accept a value that later overflowed a timestamp/denominator addition, overran a fixed-width index, or disabled a safety gate. This adds the missing upper bounds at each boundary and makes two emission-math sites overflow-safe directly, each with a named constant and reject/boundary tests.

Fixes (audit refs): WSA-38 opreg zero min_bond; WSA-39 uwrit lock duration; WSA-43 dclaim claim window; WSA-36 epoch group count + minimum_active multiply; WSA-45 snapshot min_providers; WSA-34 emissions round denominator; WSA-40 emissions pending accumulator.

The accumulator (WSA-40) saturates rather than throws because accrueepoch runs as an inline action on sysio.epoch::advance, where a throw would halt epoch advancement chain-wide — mirroring the existing sysio.dclaim reward-credit precedent. Recompiled the five affected contract wasms; all ABIs are unchanged.


contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -238,7 +238,13 @@ void epoch::setconfig(uint32_t epoch_duration_sec,
...
    check(operators_per_epoch > 0, "operators_per_epoch must be positive");
    check(batch_op_groups > 0, "batch_op_groups must be positive");
-   check(batch_operator_minimum_active == operators_per_epoch * batch_op_groups,
+   check(batch_op_groups <= MAX_BATCH_OP_GROUPS,
+         "batch_op_groups exceeds the uint8 group-index ceiling (255)");
+   // Widen the product to uint64 so a large operators_per_epoch cannot wrap the
+   // uint32 multiply into a small value that spuriously matches a small
+   // batch_operator_minimum_active; an honest product beyond uint32 simply fails
+   // the equality against the uint32 field, which is the intended rejection.
+   check(static_cast<uint64_t>(operators_per_epoch) * batch_op_groups == batch_operator_minimum_active,
```
huangminghuang, line 254: [P1] The widened multiply fixes the wrap check, but `operators_per_epoch` is still effectively unbounded. For example, `operators_per_epoch=UINT32_MAX`, `batch_op_groups=1`, `batch_operator_minimum_active=UINT32_MAX` passes this check; after an existing schedule is present, the next `advance` reserves vectors using `cfg.batch_op_groups * cfg.operators_per_epoch` and `cfg.operators_per_epoch` before any availability guard, which can abort epoch advancement chain-wide. Please add a practical upper bound on `operators_per_epoch` / `batch_operator_minimum_active`, not only on `batch_op_groups`.

contracts/sysio.system/src/emissions.cpp
```diff
@@ -490,7 +490,16 @@ void system_contract::accrueepoch(uint32_t epoch_index,
...
-   state.pending_emission_amount += per_epoch_emission;
+   // Saturating accumulate. accrueepoch runs as an inline action from
+   // sysio.epoch::advance, so this must never throw — a throw would abort epoch
+   // advancement chain-wide. Cap the running total at the asset maximum (2^62-1)
+   // instead: it is drained to zero every pay epoch and reaching the cap needs an
+   // absurdly misconfigured per-epoch emission, but capping keeps the sum from
+   // wrapping negative and corrupting the payout math either way.
+   {
+      const int64_t room = sysio::asset::max_amount - state.pending_emission_amount;
+      state.pending_emission_amount += (per_epoch_emission <= room ? per_epoch_emission : room);
```
huangminghuang: [P1] The accumulator now saturates here, but `sysio.epoch::check_emissions_ready` still computes the pay-epoch amount as `pending_emission_amount + emission_amount`. With `pay_cadence_epochs > 1`, a non-pay epoch can clamp pending to `asset::max_amount`; the following pay-epoch gate then requires `max + positive_emission`, which no WIRE balance can satisfy, so advancement stays blocked before this saturated path can drain. The gate needs to use the same saturated total that `accrueepoch`/`payepoch` will see, or config validation needs to prevent reaching this saturated state mid-period.

huangminghuang (commented):
Posted inline review comments on the remaining config-bound issues.

contracts/sysio.epoch/src/sysio.epoch.cpp
heifner: Added magnitude bounds alongside the equality: `operators_per_epoch` is now capped at 100 (`MAX_OPERATORS_PER_EPOCH`, bounding the per-epoch recorddel/termcheck fanout and the new-tail reserve in `advance`) and `batch_operator_minimum_active` at 1000 (`MAX_SCHEDULED_BATCH_OPERATORS`, bounding the resident window held in `epochstate` and the per-outpost BatchOperatorGroups attestation). The `UINT32_MAX / 1 / UINT32_MAX` triple from your example is covered by the new `setconfig_rejects_oversized_operators_per_epoch` test.

contracts/sysio.system/src/emissions.cpp
heifner: Took both halves. The gate now computes `period_emission` with the identical saturating accumulation `accrueepoch` applies -- extracted as `emissions::saturating_accrue` and shared by both sites, so the gate total, the stored pending amount, and `payepoch`'s equality check cannot diverge. To keep the saturated state from being reachable at all, both config boundaries now enforce `emissions::period_accrual_fits_asset_range` (pending + per-epoch ceiling x pay_cadence_epochs must fit in the asset range): `setemitcfg` rejects the annual-value/cadence side, and `sysio.epoch::setconfig` re-validates the stored emission config on duration changes since the per-epoch ceiling scales linearly with `epoch_duration_sec`. Covered by `setemitcfg_bounds_period_accrual_to_asset_range` and `epoch_setconfig_rejects_duration_breaking_accrual_bound`.

huangminghuang (approved):
Approved. I did not find any blocking issues in the config bounds or emissions accumulator changes. The guarded paths line up with the protected arithmetic/liveness concerns, and diff whitespace is clean. CI build/test jobs were still pending when reviewed.

PR #491, opened 2026-07-07
net_plugin: prune out-of-schedule bp peer connections in the connection monitor

heifner:
# net_plugin: reconcile bp gossip connections via the connection monitor instead of a schedule-edge disconnect sweep

## Problem

`auto_bp_gossip_peering_test` fails intermittently on its final check: after the producer
schedule shrinks, an unscheduled node keeps a gossip connection to a still-scheduled producer
for the entire 60s timeout (e.g. [run 28885478787](https://github.com/Wire-Network/wire-sysio/actions/runs/28885478787/job/85684524524)).

The nodeop logs from that run show a race between the two auto-bp-peering code paths
(node_07 = defproducerh, all within one second at 17:43:36):

| time | thread | event |
|---|---|---|
| .044 / .045 | net-0 / net-3 | incoming gossip messages trigger `connect_to_active_bp_peers()`, which snapshots the connect-address list from the **still-old schedule (all 21 producers)** |
| .053 | main | `on_active_schedule` fires (v1 -> v2, 4-producer schedule) |
| .127–.137 | main | the disconnect sweep closes all 17 dropped producers' addresses and **erases them from the connections index** (`disconnecting: 'localhost:27580' - 48`) |
| .1387–.1395 | net-3 | the stale connect loop, still iterating its 21-address snapshot, finds those addresses missing from the index and **re-creates connections 61–74 to every just-dropped producer** (`created connection - 74 to localhost:27580`) |

Once re-established, nothing ever disconnects the connection again: `on_active_schedule`
only acts on a schedule *version change* and the connection monitor only reaps closed
sockets. The prior flaky-test fixes (extending the retry window to 60s, 8472433b44 etc.)
could not help — the stale connection is not slow to prune, it is never pruned.

Two secondary defects with the same root:

- The disconnect sweep looks connections up by host in the (non-unique) `by_host` index, so
  it can act on a dead object while the live socket between the same pair survives. In the
  failing run, `disconnecting: 'localhost:27580' - 48` closed a connection object that had
  already lost the duplicate-handshake race at 17:42:53; the live socket was untouched.
- A closed outbound entry to a now-unscheduled producer left in the index is perpetually
  re-dialed by the connection monitor's reconnect logic once its duplicate disappears.

## Fix

Replace the one-shot disconnect sweep with periodic reconciliation in the connection
monitor:

- `bp_connection_manager::gossip_bp_addresses_to_drop()` (new): the gossip + configured bp
  peer addresses whose producers are neither in the active nor in the pending schedule — or
  every known bp peer address once this node has dropped out of the bp gossip group.
- `connections_manager::connection_monitor()` prunes outbound connections (open, or closed
  and awaiting reconnect) to those addresses each pass. Supplied peers
  (`p2p-peer-address` / `/v1/net/connect`) are never pruned, matching the old
  `disconnect_gossip_connection` exemption.
- `on_active_schedule()` keeps the group-membership decision (set
  `connect_bp_gossip_peers = false` when all of this node's bp accounts drop out of
  scheduling proximity) but no longer disconnects anything itself.
- `disconnect_gossip_connection()` is removed.

Why reconciliation is the right shape: a concurrent `connect_to_active_bp_peers()` working
from a pre-change address snapshot can now at worst resurrect a connection until the next
monitor pass (`connection-cleanup-period`, default 30s, 5s in the failing test), not
permanently. Pruning only *outbound* connections means each side closes the connections it
dialed, the TCP close heals the other end, and neither side ever closes an inbound
connection (which is what protects a manually requested connection on its accepting side).

## Overhead for nodes not using bp gossip

Most nodes run neither `--p2p-bp-gossip-endpoint` nor `--p2p-auto-bp-peer`. For them
`gossip_bp_addresses_to_drop()` returns immediately on the `auto_bp_peering_enabled()`
config check (no locks taken), and the per-connection prune check short-circuits on the
empty address set — one boolean test per monitor pass, zero per-connection work. Nodes with
only `--p2p-auto-bp-peer` build a set bounded by their configured peer count once per pass.

## Testing

- `test_net_plugin`: reworked `test_on_active_schedule1/2` to assert the prune set, added
  `test_drop_out_of_gossip_group` (drop-out sets `connect_bp_gossip_peers=false` and prunes
  every known address; a new pending schedule re-enters the group) and
  `test_no_drop_out_while_pending` (a bp account still in the pending schedule keeps the
  node in the group and its address unpruned). All 78 cases pass.
- `tests/auto_bp_gossip_peering_test.py` passes locally, 2/2 runs (the failing scenario:
  schedule shrink 21 -> 4, node_19 keeps the manual connection and drops the stale gossip
  peer). The test log shows the reconciliation converging: node_19's stale connections
  prune down each cleanup cycle (20 -> 18 -> 15 -> 12 ...), defproducerh gone within ~3s.
- `tests/auto_bp_peering_test.py` passes locally in both variants (`-v` as CI runs it, and
  `--activate-if`) — the auto-bp-peer connect path.

### DIFF
--- a/plugins/net_plugin/include/sysio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/sysio/net_plugin/auto_bp_peering.hpp
@@ -597,13 +597,19 @@ public:
                // do not hold mutexes when calling resolve_and_connect which acquires connections mutex since other threads
                // can be holding connections mutex when trying to acquire these mutexes
                address_set_t addresses = find_gossip_bp_addresses(pending_connections, "connect");
-               for (const auto& add : addresses) {
-                  self()->connections.resolve_and_connect(add, self()->get_first_p2p_address());
-               }
 
+               // publish the new pending set BEFORE dialing: a concurrent connection_monitor pass
+               // must see these producers in its keep set, otherwise it can prune the
+               // just-established pending-only connections as out-of-schedule — and with
+               // pending_schedule_version advanced below they would not be redialed until the
+               // schedule activates
                {
                   fc::lock_guard g(mtx);
                   pending_bps = std::move(pending_connections);
+               }
+
+               for (const auto& add : addresses) {
+                  self()->connections.resolve_and_connect(add, self()->get_first_p2p_address());
                }
 
                pending_schedule_version = schedule.version;
--- a/plugins/net_plugin/include/sysio/net_plugin/auto_bp_peering.hpp
+++ b/plugins/net_plugin/include/sysio/net_plugin/auto_bp_peering.hpp
@@ -51,7 +51,6 @@ class bp_connection_manager {
    } config; // thread safe only because modified at plugin startup currently
 
    // the following members are only accessed from main thread
-   name_set_t             pending_bps;
    uint32_t               pending_schedule_version = 0;
    uint32_t               active_schedule_version  = 0;
 
@@ -60,6 +59,7 @@ class bp_connection_manager {
 
    fc::mutex                     mtx;
    gossip_buffer_initial_factory initial_gossip_msg_factory GUARDED_BY(mtx);
+   name_set_t                    pending_bps GUARDED_BY(mtx); // written on main thread, read by gossip_bp_addresses_to_drop() on net threads
    name_set_t                    active_bps GUARDED_BY(mtx);
    name_set_t                    active_schedule GUARDED_BY(mtx);
 
@@ -116,6 +116,16 @@ class bp_connection_manager {
    void set_active_bps(name_set_t bps) {
       fc::lock_guard g(mtx);
       active_bps = std::move(bps);
+   }
+   // for testing
+   name_set_t get_pending_bps() {
+      fc::lock_guard g(mtx);
+      return pending_bps;
+   }
+   // for testing
+   void set_pending_bps(name_set_t bps) {
+      fc::lock_guard g(mtx);
+      pending_bps = std::move(bps);
    }
 
 public:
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/include/sysio/net_plugin/auto_bp_peering.hpp
```diff
@@ -557,12 +601,16 @@ class bp_connection_manager {
                   self()->connections.resolve_and_connect(add, self()->get_first_p2p_address());
                }
 
-               pending_bps = std::move(pending_connections);
+               {
+                  fc::lock_guard g(mtx);
+                  pending_bps = std::move(pending_connections);
```
huangminghuang, line 608: `connection_monitor()` now uses `pending_bps` to decide which BP peer addresses are safe to keep, but this writes `pending_bps` only after the new pending peers have already been added via `resolve_and_connect()`. If the monitor runs in that window, it can prune those pending-only outbound connections as out-of-schedule; then `pending_schedule_version` advances and the same pending schedule will not redial them. Consider storing `pending_connections` into `pending_bps` under the mutex before the connect loop, so the monitor sees the keep set before the connections become visible.

heifner: Good catch — fixed in ca3e4e1cdf exactly as suggested: the pending set is now stored under the mutex after `find_gossip_bp_addresses` computes the dial list but before the `resolve_and_connect` loop, so a concurrent monitor pass sees the new producers in its keep set before any pending-only connection becomes visible. The unit test's `resolve_and_connect` mock now asserts `pending_bps` is already published when each dial goes out (that assertion fails against the previous ordering). `test_net_plugin` 78/78 and `auto_bp_gossip_peering_test.py` re-verified locally.

huangminghuang (approved):
Approved. The pending keep-set race from the earlier review is addressed by publishing pending_bps before dialing, with a unit assertion covering the ordering. I saw no remaining code-review findings; just let the remaining CI jobs finish before merge.

PR #494, opened 2026-07-08
Test: Decide late-block lockout fast-path expectation from producer state

heifner:
`nodeop_late_block_test` failed on master ([run 28960279897](https://github.com/Wire-Network/wire-sysio/actions/runs/28960279897/job/85929850599)) asserting the lockout fast-path diagnostic ("applying blocks while producing: head's branch is locked out") in node_03's log, while every behavioral assertion passed: node_03 switched forks, converged to the canonical branch at the fork height, and the canonical defproduceri blocks were present.

The fast path only runs while node_03 is actively building a block, and every production round ends with a ~450ms gap where it is not: the produce-block-offset ramp completes the round's last block early, and the next round's first block may not start before its production window opens ("Not starting block until"). In the failing run the bridge relaunch took ~4.3s and delivered the canonical branch inside the gap right after node_03 produced its defproducerk round's last block, so the blocks were applied through the regular non-producing path and the diagnostic was legitimately absent. The reconnect timing cannot pin which case occurs, so the unconditional assertion is a latent race.

Decide the expectation from the producer state at the first post-partition fork switch instead: the fast-path line is required exactly when node_03's log shows it had started a block of its own that was still unfinished at switch time (the fast path itself aborts that block); otherwise its absence is the expected outcome. A regression that defers applying a lockout branch while a block is under construction still fails.

Validated by replaying the decision against the failing run's node_03 log (waived) and the six saved node_03 logs from the PR #478 validation runs (strict branch, diagnostic present), and with local runs of the updated test, which took the strict branch and passed.

### DIFF
--- a/tests/nodeop_late_block_test.py
+++ b/tests/nodeop_late_block_test.py
@@ -4,6 +4,7 @@ import re
 import shutil
 import signal
 import time
+from datetime import datetime
 from TestHarness import Cluster, TestHelper, Utils, WalletMgr
 from TestHarness.Node import BlockType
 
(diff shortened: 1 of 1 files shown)


huangminghuang (approved):
Approved. The updated expectation keeps the strict lockout fast-path assertion when node_03 was actively building an unfinished block, while allowing the legitimate between-block path. I found no code-review issues; please let the pending CI jobs finish before merging.

PR #495, opened 2026-07-08
Msgch: chain outbound OPP envelopes and verify inbound previous_envelope_hash

heifner:
Completes the depot side of the envelope-chaining work started in Wire-Network/wire-ethereum#128 (SEC-36); tracked as SEC-107.

The epoch digest is keccak256 over the canonical field-complete envelope encoding with the envelope_hash field blanked, exactly what the generated Solidity codec computes. buildenv now emits that canonical encoding directly (the stored blob is the digest preimage), chains each emit from its predecessor's digest, and apply_consensus verifies inbound envelopes against a per-outpost chain tip recorded on outpcons. A non-empty prev-hash that breaks the chain drops the envelope without aborting the delivering transaction; an empty prev-hash is still accepted and logged (staged, mirroring the original Ethereum posture).

Two interim accommodations, both confirmed against a live flow-swap-from-wire run: the Solana outpost chains CROSS-STREAM (its outbound envelopes link to the digest of the last depot envelope it accepted, because its single previous_epoch_hash slot is only advanced by epoch_in), so apply_consensus also accepts a prev that exactly equals the depot's own last outbound emit digest for the outpost; SEC-114 tracks aligning the outpost to per-stream chaining, after which that branch and the staged empty-prev acceptance are removed. In the e2e run all four legs verified: WIRE to ETH under the hard epochIn flip, ETH to WIRE, WIRE to SOL (the Solana epoch_in already assumes canonical preimage bytes), and SOL to WIRE via the cross-stream link.

Golden vectors generated from the deployed Solidity OPPCommon.epochHash are pinned in both this repo's tests and wire-ethereum's, so any canonical-encoding drift fails both sides. The Ethereum-side enforcement flip (reject empty prev-hash for non-genesis) is Wire-Network/wire-ethereum#132, stacked on Wire-Network/wire-ethereum#128 and gated on this PR landing and deploying.


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -887,6 +903,60 @@ void apply_consensus(name self, uint64_t chain_code, uint32_t epoch_index,
...
+         opcons.contains(opc_pk) ? opcons.get(opc_pk).envelope_digest : checksum256{};
+      if (chain_tip != checksum256{}) {
+         if (envelope.previous_envelope_hash.empty()) {
+            sysio::print_f("apply_consensus: chain_code=%llu epoch=%u accepted with empty "
+                           "previous_envelope_hash (staged chain enforcement)\n",
+                           static_cast<unsigned long long>(chain_code), epoch_index);
+         } else {
+            const auto prev = to_checksum256_exact(envelope.previous_envelope_hash);
+            bool continues_chain = prev.has_value() && *prev == chain_tip;
+            if (!continues_chain && prev.has_value()) {
```
huangminghuang: This fallback should probably be scoped to SVM outposts. The PR describes it as an interim Solana-only accommodation, but as written any outpost can bypass the normal inbound-chain check whenever its non-empty previous_envelope_hash matches the depot's latest outbound envelope digest for that chain_code. That weakens the new verification for EVM outposts too, and the test currently exercises ETH, so the broader behavior is locked in. Can we look up the chain row before this check, require CHAIN_KIND_SVM for the cross-stream alternate, and add a negative EVM case?

libraries/opp/proto/sysio/opp/opp.proto
```diff
@@ -63,6 +63,15 @@ message Message {
...
 // Envelope that spans a range of messages within an epoch.
+//
+// Envelopes chain per (depot, outpost) stream: `previous_envelope_hash` carries the epoch digest
+// of the previously emitted envelope, defined as keccak256 over the FIELD-COMPLETE canonical
+// encoding (every singular field written, including proto3 defaults) with `envelope_hash` blanked.
+// That canonical encoding is produced by the generated Solidity/Solana codecs and, on the WIRE
+// depot, by `contracts/sysio.opp.common/include/sysio.opp.common/opp_canonical_codec.hpp`.
+// Any change to this message or the types it embeds (Endpoints, ChainId, Message, MessageHeader,
+// MessagePayload, AttestationEntry, EncodingFlags) MUST be mirrored there, and the cross-language
+// golden vectors in `contracts/tests/sysio.msgch_tests.cpp` regenerated.
```
huangminghuang: Tiny cleanup: the golden vectors moved into contracts/tests/sysio.msgch_chain_tests.cpp, but this maintenance note still points to contracts/tests/sysio.msgch_tests.cpp. The same stale path appears in opp_canonical_codec.hpp, so updating both would make the next proto/codec refresh easier to follow.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: Scoped in 3386ec6f91: the source chain row is resolved once before the verification block (it already fed the audit-log endpoints projection below, so this is a hoist rather than a new lookup) and the cross-stream alternate now requires CHAIN_KIND_SVM. EVM outposts get no fallback. The positive test moved to the SOL outpost and a negative EVM case pins the drop.

libraries/opp/proto/sysio/opp/opp.proto
heifner: Fixed in 765e99b696: both notes now point at contracts/tests/sysio.msgch_chain_tests.cpp.

huangminghuang (approved):
Re-reviewed the latest head. The SVM-scoped fallback and golden-vector path cleanup are addressed, and I don't have any remaining code findings. Apple Silicon and OPP bundle checks are green; Linux matrix was still running at review time.

huangminghuang (approved):

PR #496, opened 2026-07-09
SEC-61: Validate generated Solidity enum decodes

huangminghuang:
## Summary
- Adds checked `fromRaw(uint64)` enum conversion to generated Solidity enum libraries and routes scalar, repeated, and map enum decode paths through it.
- Tightens generated `isValid` to exact declared enum values, including sparse enums, so unknown protobuf enum discriminants fail closed instead of narrowing into valid UDVT values.
- Deduplicates aliased enum numbers for generated validity/conversion branches while retaining alias constants, and reports invalid raw values with `InvalidEnumValue(uint64 raw)`.

## Validation
- Focused `protoc-gen-solidity` Jest coverage for enum generation, alias deduplication, and enum field decode paths.
- Full OPP tools workspace tests passed locally.
- OPP bundle generation passed; generated audit targets now use `AttestationTypeLib.fromRaw(_v)` and `ActionTypeLib.fromRaw(_v)`.
- `solc` compile passed for generated `Types.sol`, `Opp.sol`, `Attestations.sol`, and `Bootstrap.sol`.
- GitHub CI is green, including Linux build/test/package matrix, Generate OPP Bundles, Root Node tooling audit, and Apple Silicon.

## Notes
- No `contracts/sysio.*` changes; system-contract e2e gate is not applicable.
- The fail-closed behavior means enum value additions require a coordinated schema/code redeploy for Solidity consumers.

### DIFF
--- a/libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
+++ b/libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
@@ -68,23 +68,35 @@ export function genEnumDefinition(desc: EnumDescriptor): string {
   lines.push(`using {${lib}.isValid} for ${name} global;`)
   lines.push(``)
 
-  // Library with constants and isValid
+  // Library with constants and checked conversions.
   lines.push(`library ${lib} {`)
 
   for (const val of desc.values) {
-    lines.push(`    ${name} constant ${val.name} = ${name}.wrap(${val.number});`)
+    lines.push(
+      `    ${name} constant ${val.name} = ${name}.wrap(${val.number});`
+    )
   }
 
-  if (desc.values.length > 0) {
-    const maxVal = desc.values.reduce(
-      (max, v) => (v.number > max.number ? v : max),
-      desc.values[0]
-    )
-    lines.push(``)
-    lines.push(`    function isValid(${name} _v) internal pure returns (bool) {`)
-    lines.push(`        return ${name}.unwrap(_v) <= ${name}.unwrap(${maxVal.name});`)
-    lines.push(`    }`)
+  const uniqueValues = [
+    ...new Map(desc.values.map(val => [val.number, val])).values()
+  ]
+  const checks = uniqueValues.map(val => `_raw == ${val.number}`).join(" || ")
+
+  lines.push(``)
+  lines.push(`    function isValid(${name} _v) internal pure returns (bool) {`)
+  lines.push(`        uint64 _raw = uint64(${name}.unwrap(_v));`)
+  lines.push(`        return ${checks || "false"};`)
+  lines.push(`    }`)
+
+  lines.push(``)
+  lines.push(
+    `    function fromRaw(uint64 _raw) internal pure returns (${name}) {`
+  )
+  for (const val of uniqueValues) {
+    lines.push(`        if (_raw == ${val.number}) return ${val.name};`)
   }
+  lines.push(`        revert("Invalid enum value");`)
+  lines.push(`    }`)
 
   lines.push(`}`)
 
--- a/libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
+++ b/libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
@@ -70,6 +70,8 @@ export function genEnumDefinition(desc: EnumDescriptor): string {
 
   // Library with constants and checked conversions.
   lines.push(`library ${lib} {`)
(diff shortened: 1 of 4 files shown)


libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
```diff
@@ -68,23 +68,35 @@ export function genEnumDefinition(desc: EnumDescriptor): string {
...
-  if (desc.values.length > 0) {
-    const maxVal = desc.values.reduce(
-      (max, v) => (v.number > max.number ? v : max),
-      desc.values[0]
-    )
-    lines.push(``)
-    lines.push(`    function isValid(${name} _v) internal pure returns (bool) {`)
-    lines.push(`        return ${name}.unwrap(_v) <= ${name}.unwrap(${maxVal.name});`)
-    lines.push(`    }`)
+  const uniqueValues = [
```
heifner: This dedup is the subtlest new logic in the PR and nothing tests it. I ran the generator on an `allow_alias` enum (`UNKNOWN=0, RUNNING=1, STARTED=1`) to confirm it's correct — `isValid` emits `_raw == 0 || _raw == 1` (no duplicate) and `fromRaw` a single `if (_raw == 1) return STARTED;`. Worth a focused test on an aliased descriptor to lock that in; a future refactor could otherwise reintroduce a duplicate `if (_raw == N)` and only CI's solc step would catch it. Minor behavioral note: `new Map` keeps the last entry per number, so `fromRaw` returns the last-declared alias (`STARTED`), not the first (`RUNNING`) — semantically identical since both `wrap(1)`, just flagging where that's decided.

libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
```diff
@@ -68,23 +68,35 @@ export function genEnumDefinition(desc: EnumDescriptor): string {
...
+  lines.push(`    }`)
+
+  lines.push(``)
+  lines.push(
+    `    function fromRaw(uint64 _raw) internal pure returns (${name}) {`
+  )
+  for (const val of uniqueValues) {
+    lines.push(`        if (_raw == ${val.number}) return ${val.name};`)
   }
+  lines.push(`        revert("Invalid enum value");`)
```
heifner: Two small notes on the revert:

1. Failing closed here is stricter than protobuf's default (which preserves unknown enum values for forward compatibility) — correct for SEC-61, and it aligns with the host C++ rejecting undeclared values, so no complaint. Worth making explicit that this bakes in "enum evolution needs a coordinated redeploy": once deployed, this Solidity reverts on any enum value added in a later schema revision.
2. The reason string is identical for every enum and drops the offending value, so an on-chain failure can't say which enum rejected what. A per-lib custom error (`error InvalidEnumValue(uint64 raw); ... revert InvalidEnumValue(_raw);`) is more debuggable and cheaper gas on 0.8.4+. Enhancement, not a blocker.

heifner (commented):
SEC-61 review — no blockers, and I verified the fix mechanically: after the change `git grep '.wrap('` across the generator returns only the compile-time constant, so every runtime decode narrowing site (scalar / repeated / map-value) now routes through `fromRaw`, and oneof enum members flow through the same singular path. Ran the jest suite (19/19) and exercised the generator on an `allow_alias` enum to confirm the dedup emits compilable Solidity. Nice side benefit beyond fail-closed: exact-match `isValid` makes the Solidity validity set identical to the host's declared-values-only check, closing the sparse-gap divergence. A couple of non-blocking notes inline.

libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
huangminghuang: Addressed in 62a21e6c0e: added focused alias coverage, made dedup explicit, and preserved the first declared alias as the canonical `fromRaw` return while still emitting all alias constants.

libraries/opp/tools/protoc-gen-solidity/src/generator/enum.ts
huangminghuang: Addressed in 62a21e6c0e: generated enum libs now use `error InvalidEnumValue(uint64 raw);` and `revert InvalidEnumValue(_raw);`. I also captured the coordinated enum-evolution implication in the PR notes.

heifner (approved):

heifner (approved):

PR #498, opened 2026-07-09
Chore: prune stale ccache entries

huangminghuang:
## Summary
- Prevent Linux and macOS PR jobs from saving new ccache blobs while still allowing them to restore existing caches.
- Add a scheduled/manual Actions cache cleanup workflow that deletes PR-scoped ccache entries and prunes older non-PR ccache entries beyond a rolling keep count.
- Harden the cleanup workflow with a concurrency guard, `keep_per_prefix >= 1` validation, generic ccache prefix grouping for future platforms, and best-effort deletion so racy missing cache ids do not abort the whole cleanup.
- Leave vcpkg binary caches and CTest cost data untouched.

## Why
The macOS ccache miss in run 29034160705 was caused by repo-wide Actions cache pressure and eviction. The repo had exceeded the default cache storage size, and the macOS ccache family had been evicted while larger Linux ccache entries remained. PR-scoped ccache saves add pressure without providing broadly reusable caches, so PR jobs should be restore-only.

## Validation
- Parsed all `.github/workflows/*.yaml` files with PyYAML.
- Ran `git diff --check`.
- Exercised the cleanup jq against the live cache list in dry-run form; it selected PR-scoped ccache entries and preserved current master caches.
- Exercised the cleanup jq against synthetic unknown-platform `ccache-*` keys to verify future ccache prefixes are still pruned.

### DIFF
--- a/.github/workflows/actions_cache_cleanup.yaml
+++ b/.github/workflows/actions_cache_cleanup.yaml
@@ -19,6 +19,10 @@ permissions:
   actions: write
   contents: read
 
+concurrency:
+  group: actions-cache-cleanup
+  cancel-in-progress: false
+
 defaults:
   run:
     shell: bash
@@ -43,6 +47,10 @@ jobs:
               exit 1
               ;;
           esac
+          if (( KEEP_PER_PREFIX < 1 )); then
+            echo "::error::keep_per_prefix must be >= 1 to preserve the restore baseline"
+            exit 1
+          fi
 
           caches_json="$(mktemp)"
           delete_list="$(mktemp)"
@@ -54,14 +62,7 @@ jobs:
 
           jq -s -r --argjson keep "$KEEP_PER_PREFIX" '
             def prefix:
-              if .key | startswith("ccache-macos-arm64-") then "ccache-macos-arm64"
-              elif .key | startswith("ccache-ubuntu24-") then "ccache-ubuntu24"
-              elif .key | startswith("ccache-asserton-") then "ccache-asserton"
-              elif .key | startswith("ccache-ubsan-") then "ccache-ubsan"
-              elif .key | startswith("ccache-asan-") then "ccache-asan"
-              elif .key | startswith("ccache-gcc-") then "ccache-gcc"
-              else empty
-              end;
+              .key | sub("-[0-9a-f]{7,40}$"; "");
 
             def candidate:
               . + {prefix: prefix};
@@ -104,5 +105,7 @@ jobs:
             fi
 
             echo "Deleting ${cache_key} (${cache_ref}, ${reason})"
-            gh cache delete "$cache_id" --repo "$GH_REPO"
+            if ! gh cache delete "$cache_id" --repo "$GH_REPO"; then
+              echo "::warning::failed to delete ${cache_key} (${cache_ref}); continuing"
+            fi
           done < "$delete_list"
--- a/.github/workflows/actions_cache_cleanup.yaml
+++ b/.github/workflows/actions_cache_cleanup.yaml
@@ -1,0 +1,108 @@
+name: "Actions cache cleanup"
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "List eligible ccache entries without deleting them"
+        type: boolean
+        default: false
+      keep_per_prefix:
+        description: "Non-PR ccache entries to keep per platform prefix"
+        type: number
+        default: 2
+  schedule:
+    # Keep Actions cache usage below the repository cap so ccache families do not thrash each other.
+    - cron: "35 3 * * *"
+
+permissions:
+  actions: write
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  prune-ccache:
+    name: Prune ccache entries
+    runs-on: ubuntu-latest
+    env:
+      CACHE_CLEANUP_DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}
+      GH_REPO: ${{ github.repository }}
+      GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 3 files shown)


heifner (commented):
CI-infra review — no blockers. Injection-safety is solid: both dispatch inputs go through `env:` and are referenced as quoted `$VAR`, with an allow-list check on `keep_per_prefix`, and no `${{ }}` is interpolated into the script body. Triggers are `workflow_dispatch` + `schedule` only, so there's no fork/PR RCE surface. I ran the jq against synthetic cache data — PR-scope deletion and keep-per-prefix pruning both behave correctly at keep=2/1/0, and all six current platform prefixes match the build matrix. A few robustness items inline, all worth-discussing, none blocking.

.github/workflows/actions_cache_cleanup.yaml
```diff
@@ -0,0 +1,108 @@
...
+          echo "Total eligible size: ${total_mib} MiB"
+
+          while IFS=$'\t' read -r cache_id cache_key cache_ref reason size_bytes; do
+            if [[ "$CACHE_CLEANUP_DRY_RUN" == "true" ]]; then
+              echo "Dry run: would delete ${cache_key} (${cache_ref}, ${reason})"
+              continue
+            fi
+
+            echo "Deleting ${cache_key} (${cache_ref}, ${reason})"
+            gh cache delete "$cache_id" --repo "$GH_REPO"
```
heifner: Under `set -euo pipefail`, one failed delete aborts the rest of the loop and fails the job. There's an unavoidable TOCTOU window between the `gh api` listing and this delete — GitHub's LRU can evict an entry, or a concurrent run (nightly `schedule` overlapping a manual dispatch) can delete it first; `gh cache delete` exits non-zero on a missing id, so a single racy miss skips every remaining eligible entry. Two small hardenings:

```bash
if ! gh cache delete "$cache_id" --repo "$GH_REPO"; then
  echo "::warning::failed to delete ${cache_key} (${cache_ref}); continuing"
fi
```

and a top-level concurrency guard so scheduled and manual runs can't race:

```yaml
concurrency:
  group: actions-cache-cleanup
  cancel-in-progress: false
```

.github/workflows/actions_cache_cleanup.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+          jq -s -r --argjson keep "$KEEP_PER_PREFIX" '
+            def prefix:
+              if .key | startswith("ccache-macos-arm64-") then "ccache-macos-arm64"
+              elif .key | startswith("ccache-ubuntu24-") then "ccache-ubuntu24"
+              elif .key | startswith("ccache-asserton-") then "ccache-asserton"
+              elif .key | startswith("ccache-ubsan-") then "ccache-ubsan"
+              elif .key | startswith("ccache-asan-") then "ccache-asan"
+              elif .key | startswith("ccache-gcc-") then "ccache-gcc"
+              else empty
```
heifner: `. + {prefix: prefix}` where `prefix` hits this `else empty` yields `empty`, which drops the element from `$all` entirely — so an unrecognized `ccache-*` key is skipped by *both* branches, including PR-scoped ones. I confirmed with a synthetic `ccache-<unknown>-…` entry on `refs/pull/…`: it's never selected at keep=2/1/0. Harmless today (all six prefixes match the current matrix), but it silently couples this script to `linux_amd64_build.yaml`'s `matrix.platform` + the macOS key — add or rename a platform and its caches leak, and any lingering PR-scoped entry for it never gets cleaned. The PR-scope branch in particular doesn't need the whitelist: either run it over the raw list, or derive the prefix generically (e.g. group by `.key | sub("-[0-9a-f]{7,40}$"; "")`) so new platforms are auto-handled. At minimum, a comment that this list must track the build matrix.

.github/workflows/actions_cache_cleanup.yaml
```diff
@@ -0,0 +1,108 @@
...
+      CACHE_CLEANUP_DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}
+      GH_REPO: ${{ github.repository }}
+      GH_TOKEN: ${{ github.token }}
+      KEEP_PER_PREFIX: ${{ github.event.inputs.keep_per_prefix || '2' }}
+    steps:
+      - name: Delete stale ccache entries
+        run: |
+          set -euo pipefail
+
+          case "$KEEP_PER_PREFIX" in
```
heifner, line 44: This accepts `0`, and keep=0 deletes *all* non-PR entries — including the `refs/heads/master` caches that, after this PR, are the only thing PR builds can restore from. The scheduled path is safe (falls back to the default `2`), so this only bites a manual dispatch that types 0, but since 0 defeats the restore-only design it's worth rejecting:

```bash
if (( KEEP_PER_PREFIX < 1 )); then
  echo "::error::keep_per_prefix must be >= 1 to preserve the restore baseline"; exit 1
fi
```

.github/workflows/actions_cache_cleanup.yaml
huangminghuang: Addressed in `86b669a16e`: added workflow-level `concurrency` and made `gh cache delete` best-effort so racy missing cache ids warn and continue.

.github/workflows/actions_cache_cleanup.yaml
huangminghuang: Addressed in `86b669a16e`: ccache prefixes are now derived generically from the trailing SHA, so unknown/future `ccache-*` platform keys are still considered for PR-scope deletion and keep-per-prefix pruning.

.github/workflows/actions_cache_cleanup.yaml
huangminghuang: Addressed in `86b669a16e`: `keep_per_prefix` now rejects values below 1 so manual cleanup cannot remove every non-PR restore baseline.

heifner (approved):

PR #500, opened 2026-07-09
Fix SEC-73 Solana vector decode bounds

huangminghuang:
## Summary
- Bound templated Borsh `read_vec<T>()` lengths against remaining bytes before `reserve()`.
- Bound Anchor IDL `Vec<T>` account decoding using conservative minimum encoded sizes before materializing `fc::variants`.
- Preserve valid zero-sized element vectors with an explicit materialization cap, and reject huge zero-sized lengths before allocation.
- Add regression tests for huge short-payload vector lengths, valid dynamic `Vec<string>` decoding, and zero-sized `Vec<EmptyStruct>` behavior.

## Validation
- `git diff --check`
- `cmake --build build/sec-73-release --target test_fc -- -j"$(nproc)"`
- `cmake --build build/sec-73-release --target test_outpost_solana_client_plugin -- -j"$(nproc)"`
- `./build/sec-73-release/libraries/libfc/test/test_fc --run_test=solana_client_tests`
- `ctest --test-dir build/sec-73-release -j"$(nproc)" -R '^test_fc$' --output-on-failure`
- `ctest --test-dir build/sec-73-release -j"$(nproc)" -R '^test_outpost_solana_client_plugin$' --output-on-failure`

## Review Notes
- Security review subagent found the initial zero-sized element edge; this PR includes the follow-up fix and the second review pass reported no findings.
- The zero-sized vector cap intentionally rejects vectors above 65,536 materialized elements because their encoded bytes cannot bound allocation work.
- No system-contract artifacts or generated OPP bindings are changed.

### DIFF
--- a/libraries/libfc/include/fc/network/solana/solana_borsh.hpp
+++ b/libraries/libfc/include/fc/network/solana/solana_borsh.hpp
@@ -197,12 +197,18 @@ private:
    size_t _pos;
 
    void ensure_remaining(size_t n) const {
-      FC_ASSERT(_pos + n <= _size, "Borsh decoder: not enough data, need {} bytes, have {}", n, remaining());
+      FC_ASSERT(n <= remaining(), "Borsh decoder: not enough data, need {} bytes, have {}", n, remaining());
    }
 
    // Helper to read primitive type
    template <typename T>
    T read_primitive();
+
+   /**
+    * @brief Smallest Borsh byte count for a supported Vec<T> element.
+    */
+   template <typename T>
+   static constexpr size_t min_vec_element_size();
 };
 
 //=============================================================================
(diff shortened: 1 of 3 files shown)


heifner (approved):
Verified the bounds are conservative lower bounds against `decode_type`'s actual per-type consumption, so no valid data gets rejected, and that the zero-sized-element cap preserves valid `Vec<EmptyStruct>` decoding (the edge a naive min>=1 fix would have broken). `min_vec_element_size` covers exactly `read_vec`'s supported type set, and the `ensure_remaining` rewrite to `n <= remaining()` also quietly fixes a size_t-wrap false-pass for huge `n`. Nice fix.

Two non-blocking suggestions:

- `min_borsh_encoded_size` / `min_borsh_fields_encoded_size` recurse through defined types with no depth limit, so a cyclic IDL (`struct A { b: B }`, `struct B { a: A }`) would overflow the stack at bound-computation time. `decode_type` / `decode_fields` already have the identical pre-existing recursion and IDLs come from operator-provided files, so this is not a regression - but a shared depth cap would harden both cheaply. Fine as a follow-up.

- The new tests exercise the extremes (`u32::max`) but not the boundaries the `<=` asserts define: `len == remaining / min_element_size` vs one past it, and the zero-sized cap at 65536 vs 65537. Off-by-ones in three separate asserts are exactly what boundary cases catch.

One optional nicety: `min_borsh_encoded_size` returns 1 for any enum. A comment noting the looseness is deliberate (a lower bound only needs to be <= the actual minimum) would keep a future reader from "tightening" it.

PR #502, opened 2026-07-10
outpost_solana: read inbound envelopes through the IDL decoder

valthon:
## Problem

nodeop read the on-chain `LatestOutboundEnvelope` account with **hardcoded Borsh offsets** that assumed the standalone `opp_outpost` field order (`{epoch_index, checksum, data, bump}` — `epoch_index` at byte 8, past the 8-byte Anchor discriminator). The folded `liqsol_core` program declares `{bump, epoch_index, checksum, data}`, shifting `epoch_index` to byte 9, so the hardcoded reader decoded the leading `bump` (`0xFF`) as the low byte of the epoch (`0xFF | 1<<8 = 511`): `read_inbound_envelope` never matched the requested epoch, every SOL→depot envelope relay silently stalled, and no reward or deposit reached the depot.

## Change

Read the account through libfc's IDL-driven `decode_account_data` — the same path this class already uses for `Reserve` / `OutpostConfig`. It verifies the 8-byte Anchor discriminator and follows the loaded IDL's declared field order, so **one nodeop decodes both program layouts value-exactly** and the hand-derived offset arithmetic (and its whole bug class) is gone.

Hardening around the read:
- **IDL ↔ program-id match** — the loaded IDL is selected by its declared `address` matching the deployed program id, so `--solana-idl-file` order can never decide which same-named IDL version's field order drives decoding; multiple candidates with none matching refuse to boot rather than silently pick one.
- **Checksum validation** — the stored `keccak256(encoded_envelope)` is verified, so a decode that read the wrong bytes for `data` (field-order drift) is caught and rejected instead of stalling.
- **Owner check** — accounts whose `owner != program_id` are rejected before decoding.
- **Role-gated boot check** — only roles that read inbound envelopes (`batch_operator`) require a decodable `LatestOutboundEnvelope`; the underwriter (which never reads inbound) is not bricked at boot by an instructions-only IDL.
- Drift signals (undecodable account, stored epoch ahead of request, checksum mismatch) log at **warning** level so a stalled relay is diagnosable; normal emit-cadence lag stays at debug.

Also **retires the now-unused `add_attestation`** binding + stub instruction: the folded `liqsol_core` IDL dropped it (attestations flow through `epoch_in` / `emit_outbound_envelope`), it had zero call sites, and binding it via `get_idl("add_attestation")` threw at construction against the folded IDL — nodeop now consumes the IDL verbatim.

**Backwards-compatible:** the standalone `opp_outpost` layout still decodes correctly (its field order drives the decode just as the integrated one does), and dropping the never-called `add_attestation` binding is inert for every program variant.

## Verification

- Plugin unit tests green — `test_outpost_solana_client_plugin` (37 cases): the complete post-fetch read path driven end-to-end against synthesized accounts for **both** program layouts, the epoch=511 reproduction (integrated bytes read through a standalone-order IDL → rejected), every drift/reject path (bad discriminator, checksum mismatch, epoch-ahead, truncation, oversize, bad protobuf, owner mismatch), the IDL-selection rules, and a pin on libfc's `bytes`→base64 rendering the reader depends on.
- `nodeop` links (the `solana_outpost_role` threading through `batch_operator` / `underwriter`).

## Companion PRs

- **wire-solana** (`next-integ`, Wire-Network/wire-solana#385) — the folded `liqsol_core` outpost + `--features development` E2E build.
- **wire-tools-ts** (Wire-Network/wire-tools-ts#25) — harness support for the integrated outpost + real staking-yield flow.

https://claude.ai/code/session_015CbgvM8EhrPL1VjX8RhNze


heifner (dismissed):
Verified the offset derivation against both known layouts - the standalone order resolves epoch/vec_len/data to 8/44/48 and the integrated `liqsol_core` order to 9/45/49 - and that every subsequent buffer read stays bounds-safe (the `data_off` early-return plus the epoch-precedes-data assert cover the epoch and length reads). Also confirmed caller behavior: `outpost_opp_job::run_inbound` catches `fc::exception`, logs, and retries next cycle, so the new asserts fail loudly without taking down nodeop. Deriving from the IDL rather than adding a second hardcoded layout is the right fix.

Comments, roughly in priority order:

1. Test coverage: `resolve_latest_layout` and `borsh_fixed_size` are the heart of the fix but sit in the anonymous namespace with no tests. The plugin already unit-tests `outpost_solana_client_detail` functions (`extract_inbound_recipient_pubkeys` et al.), so moving these there enables table-driven cases: both field orders pinned to their exact offsets, a variable-length field before `data` throws, missing `epoch_index`/`data` throws. Given the RCA was wrong offsets silently stalling the relay, a regression test pinning the offsets for both IDLs is exactly the guard this change exists to provide.

2. Fields are located by name only. If a future layout declares `epoch_index` as u16/u64, `read_u32_le` misreads it; if `data` were a fixed `[u8; N]` there is no Borsh length prefix and `vec_len_off` would read payload bytes as a length. Two cheap asserts (`epoch_index` is u32, `data` is bytes/`Vec<u8>`) would complete the fail-loudly guarantee this PR is built around.

3. The RCA comment has the direction inverted: it says "the integrated offsets decoding the standalone account produced the epoch=511 RCA", but the removed constants hardcoded the standalone offsets (epoch at byte 8), and 511 = `0xFF | 1<<8` arises when those offsets read the integrated account (bump 0xFF at byte 8, epoch 1 at byte 9). The PR body states it correctly. Worth fixing since this comment is the postmortem record future readers will trust.

4. Design question, your call: the comment justifying per-read resolution notes the IDL is immutable after construction. Resolving once in the constructor would fail at boot on an undecodable IDL instead of at the first inbound poll (where the job loop wlogs and retries forever), surfacing operator misconfiguration much earlier. Per-read is defensible given the cost is negligible next to the RPC.

5. Minor: `(*type.array_len) * (*element)` in `borsh_fixed_size` is an unchecked multiply. Operator-file trust plus the downstream `buf.size()` checks make it low risk, but PR #500 in flight adds `checked_mul_size` in libfc for the same shape of computation. Related: `borsh_fixed_size` and #500's `min_borsh_encoded_size` look near-identical but are semantically different (exact fixed size vs conservative lower bound - `string` is nullopt in one, 4 in the other); a cross-referencing comment would keep someone from deduplicating them into a bug later.

heifner (approved):

PR #504, opened 2026-07-10
Move underwriter diagnostics to a dedicated http api category

heifner:
Fixes SEC-65 / WSA-146.

`/v1/underwriter/stats` and `/v1/underwriter/commits` were registered under `api_category::node` (`UINT32_MAX`). The HTTP dispatch predicate is a bit-overlap test, so node-category handlers match every listener category set, and the underwriter's operator metadata (account identity, client ids, outpost contract addresses, counters, outstanding commit ledger) was reachable on category-isolated public listeners — e.g. a public `snapshot_ro` bind configured per the snapshot README's isolation guidance.

- New `api_category::underwriter` (`1 << 12`); both underwriter endpoints register under it. They remain reachable on the default all-category listeners (`http-server-address` / `unix-socket-path`); category-isolated listeners no longer serve them unless the operator opts in with `--http-category-address=underwriter,<addr>` (gated on `--plugin=sysio::underwriter_plugin` like every other category).
- Startup warning when the underwriter category is exposed on a non-loopback address, mirroring the snapshot_ro exposure notice.
- `valid_category_addresses` / `invalid_category_addresses` http_plugin unit tests extended: the underwriter listener serves its own category plus node handlers; other isolated listeners return 404 for the underwriter route and do not enumerate it in `get_supported_apis`; the missing-plugin config error is covered.
- snapshot_api_plugin README: the isolation examples now state that node-global endpoints (`/v1/chain/get_info`, `/v1/node/get_supported_apis`) remain reachable on every listener by design (upstream behavior, unchanged here).
- underwriter_plugin README: new HTTP diagnostics section documenting both endpoints, the startup-gate `status` payloads, and listener exposure — default all-category listeners unchanged, category-isolated deployments opt in with a private `--http-category-address=underwriter,127.0.0.1:<port>` listener. The example sets `--underwriter-enabled true` (endpoint registration is skipped while the underwriter is disabled), and `/v1/chain/get_info` is qualified as present only when `sysio::chain_api_plugin` is loaded.
- The diagnostic-endpoint item is dropped from the README deferred list now that the endpoints exist; no `clio` wrapper is planned, and the stale source comments claiming one (plus the "unconditional registration" wording) are removed.

### DIFF
--- a/plugins/underwriter_plugin/README.md
+++ b/plugins/underwriter_plugin/README.md
@@ -126,6 +126,11 @@ completes they report the startup-gate state instead of the payload
 `wiring_failed` / `startup_failed` with `detail`); once the gate opens
 they serve the payloads above with `status: "active"`.
 
+The endpoints are registered only when the underwriter is enabled:
+with the plugin loaded but `--underwriter-enabled false` (the
+default), `plugin_startup` skips endpoint registration and every
+listener returns 404 for these routes.
+
 ### Listener exposure
 
 The endpoints live in the dedicated `underwriter` HTTP API category —
@@ -148,12 +153,16 @@ nodeop \
   --http-server-address http-category-address \
   --http-category-address underwriter,127.0.0.1:8890 \
   --plugin sysio::underwriter_plugin \
+  --underwriter-enabled true \
   ...
 ```
 
-The listener also serves the node-global endpoints
-(`/v1/chain/get_info`, `/v1/node/get_supported_apis`), which are
-reachable on every listener by design. Like every category,
+The listener also serves the node-global endpoints, which are
+reachable on every listener by design: `/v1/node/get_supported_apis`
+(always registered) and `/v1/chain/get_info` when
+`sysio::chain_api_plugin` is loaded — the underwriter depends only on
+`chain_plugin`, so the example above does not load it. Like every
+category,
 `underwriter` is validated against its owning plugin: naming it in
 `--http-category-address` without `--plugin sysio::underwriter_plugin`
 is a startup configuration error. Binding it to a non-loopback address
--- a/plugins/underwriter_plugin/README.md
+++ b/plugins/underwriter_plugin/README.md
@@ -106,6 +106,63 @@ verifies the signature against every permission on `uw_account` via the
 > the underwriter's per-chain contract / program address now lives in that
 > entry rather than in a per-family scalar option.
 
+## HTTP diagnostics
+
+Read-only diagnostic endpoints, served by `http_plugin` on the
+read-only exec queue:
+
+- `/v1/underwriter/stats` — session counters + config snapshot:
+  underwriter account, enabled/active flags, scan + timeout intervals,
+  per-chain outpost wiring (`chain_code`, `kind`, `client_id`,
+  `commit_addr`, `source_deposit_addr`), uwreq/commit/failure/mismatch
+  counters, outstanding-commit count, SOL source-deposit cursor health.
+- `/v1/underwriter/commits` — outstanding confirmed commits, one entry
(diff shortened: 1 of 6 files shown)


huangminghuang (commented):
[P2] Please document the new underwriter HTTP category in `plugins/underwriter_plugin/README.md`.

The README still describes the diagnostic HTTP endpoint as deferred, although `/v1/underwriter/stats` and `/v1/underwriter/commits` already exist. With this PR, category-isolated deployments will return 404 for those endpoints unless `underwriter` is explicitly bound. Please add an HTTP diagnostics section showing a private `--http-category-address=underwriter,127.0.0.1:<port>` listener, explain that default all-category listeners remain unchanged, and revise the deferred item so only the planned `clio` wrapper remains deferred.

plugins/underwriter_plugin/README.md
```diff
@@ -106,6 +106,63 @@ verifies the signature against every permission on `uw_account` via the
...
+(`--http-server-address http-category-address`) must opt in
+explicitly: listeners without the `underwriter` category return 404
+for these routes and omit them from `/v1/node/get_supported_apis`.
+Bind the category to loopback or a private management network:
+
+```
+nodeop \
+  --http-server-address http-category-address \
+  --http-category-address underwriter,127.0.0.1:8890 \
+  --plugin sysio::underwriter_plugin \
```
huangminghuang, line 155: [P2] This example does not actually expose the documented endpoints. `--underwriter-enabled` defaults to `false`, and `underwriter_plugin::plugin_startup()` returns before `register_http_endpoints()` when disabled, so this configuration produces 404s. Please add `--underwriter-enabled true`, explicitly qualify the example as requiring an already-enabled underwriter configuration, or move endpoint registration before the disabled guard if disabled-state diagnostics are intended.

plugins/underwriter_plugin/README.md
```diff
@@ -106,6 +106,63 @@ verifies the signature against every permission on `uw_account` via the
...
+  ...
+```
+
+The listener also serves the node-global endpoints
+(`/v1/chain/get_info`, `/v1/node/get_supported_apis`), which are
```
huangminghuang: [P3] `/v1/chain/get_info` is not unconditional here: its node-category handler is registered only when `sysio::chain_api_plugin` is loaded, and `underwriter_plugin` depends on `chain_plugin`, not `chain_api_plugin`. The example above therefore guarantees only `/v1/node/get_supported_apis`. Please qualify `get_info` with “when `sysio::chain_api_plugin` is loaded,” matching the minimal snapshot example, or load that plugin explicitly.

plugins/underwriter_plugin/README.md
heifner: Added `--underwriter-enabled true` to the example and stated the precondition explicitly in 48476dac9a. Kept registration behind the disabled guard: the gate payloads have no disabled state, so registering while disabled would report a misleading `waiting_for_sync`. If disabled-state diagnostics are wanted, I would add a dedicated status value as a follow-up.

plugins/underwriter_plugin/README.md
heifner: Qualified in 48476dac9a — the section now guarantees only `/v1/node/get_supported_apis` and notes `/v1/chain/get_info` requires `sysio::chain_api_plugin`, matching the snapshot README minimal-example phrasing.

huangminghuang (approved):
Approved — the implementation is sound and the documentation follow-ups address the review findings.

Before merge, please refresh the PR description as required by the repository instructions so it captures the complete current PR: the underwriter HTTP diagnostics README section, the `--underwriter-enabled true` prerequisite for endpoint registration, the conditional `/v1/chain/get_info` availability, and the final validation/CI results.

PR #505, opened 2026-07-10
SEC-123: Cap Solana IDL decode recursion depth

huangminghuang:
## Summary

- Cap nested Solana IDL type descent at 32 across conservative Borsh size estimation and recursive decoding.
- Preserve existing decoder entry points while returning an explicit assertion for cyclic or excessively deep IDLs.
- Add regression coverage for direct and vector-bound cycles, accepted/rejected depth boundaries, remaining-byte vector bounds, and zero-sized vector materialization bounds.

## Validation

- Focused Release validation: `solana_client_tests`, `test_fc`, and `test_outpost_solana_client_plugin`.
- Broad parallel-safe Release validation: 332/332 tests passed.
- GitHub PR CI passed across Apple Silicon, Ubuntu 24, assertions-on, GCC, ASan, and UBSan: [workflow run](https://github.com/Wire-Network/wire-sysio/actions/runs/29104585382).

## Risk

IDL values requiring more than 32 nested type descents are now rejected deliberately. No system-contract or generated OPP artifacts are changed, so the platform system-contract e2e gate does not apply.

### DIFF
--- a/libraries/libfc/include/fc/network/solana/solana_client.hpp
+++ b/libraries/libfc/include/fc/network/solana/solana_client.hpp
@@ -51,6 +51,16 @@ struct solana_confirm_options {
 
 inline constexpr solana_confirm_options solana_confirm_option_defaults{};
 
+/**
+ * @brief Maximum nested IDL type descents during Borsh size estimation and decoding.
+ */
+inline constexpr size_t max_idl_type_nesting_depth = 32;
+
+/**
+ * @brief Maximum materialized elements for an IDL vector whose element encoding can consume zero bytes.
+ */
+inline constexpr uint32_t max_zero_sized_idl_vector_elements = 64u * 1024u;
+
 class solana_client;
 using solana_client_ptr = std::shared_ptr<solana_client>;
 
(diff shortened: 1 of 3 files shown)


heifner (approved):
Read both recursion trees (Borsh size estimation and decode) end-to-end. The cap is threaded correctly through every nested descent and asserts before recursing, so actual recursion stays ~32 frames and fails with a clean `fc::assert_exception` instead of a stack overflow. Siblings correctly share depth; only genuine nesting increments. Nice detail: depth-exceed *throws* rather than returning `nullopt`, so a too-deep vec element can't be smuggled through with `len == 0` — the cyclic-vector-bound test locks that path.

Beyond DoS, this improves determinism: it replaces a platform-dependent stack overflow (crash some nodes, not others) with a deterministic reject at the same depth everywhere. Tests are boundary-precise (accept at max, reject at max+1 with the specific error, cycles rejected in both the decode and size-estimation trees) and key off the constant rather than a hardcoded 32. CI green across ASan/UBSan/GCC/assertions. LGTM.

PR #506, opened 2026-07-10
Opp: define message header semantics and derive them in buildenv

heifner:
Part 1 of SEC-102 (WSA-005 item 3): standardize the OPP envelope semantic header so every emitter populates it identically and the depot can eventually validate it.

- `opp.proto`: normative `MessageHeader` semantics — keccak256 over the field-complete canonical encoding for `payload_checksum`/`header_checksum`, `message_id` = header checksum with the first 8 bytes replaced by the big-endian sequence number (matching the Solidity `OPPCommon` derivation), millisecond timestamps (the documented intent; only the emit sites had drifted to seconds), `endpoints` reserved-default, `data_size == len(data)`.
- `encoding_flags` (and its orphaned enum types) removed with `reserved 4;`: never read by any implementation, cannot self-describe parse conventions, and a self-declared hash algorithm is a downgrade knob — per-chain hash selection belongs on the `sysio.chains` registry row if ever needed.
- `sysio.msgch::buildenv` derives the full header on every (re)build inside the trim loop and chains message ids per outpost via a new `last_message_id` column on the one-deep `outenvelopes` row.
- `opp_canonical_codec.hpp` gains shared derivation helpers (`encode(MessagePayload/MessageHeader)`, `header_digest`, `derive_message_id`, `message_sequence`) reused by the emitter, the test oracle, and the follow-up inbound validator.
- Golden vectors regenerated with spec-derived headers; the identical pins go into the wire-solana and wire-ethereum vector suites next, which re-confirm them against the Rust and Solidity codecs.

SEC-102 PR set (review in this order): this PR (spec + depot emit) → Wire-Network/wire-solana#370 (Solana emit population) → Wire-Network/wire-ethereum#137 (Ethereum ms units + vector re-confirmation) → #508 (depot enforcement, stacked on this branch).


contracts/sysio.opp.common/include/sysio.opp.common/opp_canonical_codec.hpp
```diff
@@ -364,6 +341,63 @@ inline sysio::checksum256 epoch_digest(const Envelope& env) {
...
+   for (size_t i = 0; i < 8; ++i) {
+      bytes[i] = static_cast<uint8_t>(sequence >> (8 * (7 - i)));
+   }
+   return sysio::checksum256{bytes};
+}
+
+/// The big-endian sequence number carried in the first 8 bytes of a wire `message_id`. Empty
+/// (stream genesis) — and any out-of-spec id shorter than 8 bytes — reads as 0, so the successor
+/// message's sequence number is 1.
+inline uint64_t message_sequence(const std::vector<char>& message_id) {
```
huangminghuang: Could we make this reject non-canonical lengths rather than mapping them? The proto permits `previous_message_id` to be empty only at stream genesis and otherwise requires 32 bytes. As written, every 1–7-byte value aliases genesis and every length of at least 8 bytes is accepted as a sequence source. PR #508 calls this helper during semantic-header validation without a separate exact-length check, so a malformed header can still recompute as valid. Returning an optional/error for lengths other than 0 or 32 (or enforcing that before this helper) would keep the verifier strict.

heifner: Done — `message_sequence` now returns `std::optional<uint64_t>`, nullopt for every length other than 0 or 32 (898e5aa7fa here), and #508's validator drops on nullopt before deriving the expected id (c79a6194cf), with a new forgery case where the checksum and id are honestly re-derived over a 5-byte `previous_message_id` so the drop is attributable to the length rule alone. `buildenv` hard-checks its own stored tip — corrupted own-state fails loudly, unlike the inbound never-throw path. The outpost emit-side mirrors only ever read their own well-formed stored tips, so they keep the plain form.

huangminghuang (approved):

PR #508, opened 2026-07-10
Opp: enforce semantic message header validation in apply_consensus

heifner:
Final wire-sysio half of SEC-102 (WSA-005 item 3), stacked on #506, completing the emitter PRs (#506, Wire-Network/wire-solana#370, Wire-Network/wire-ethereum#137). Direct enforcement — no staged mode — per the pre-launch dispensation.

- `apply_consensus` validates every message's semantic header against the spec derivation before dispatch: `data_size == len(data)` per attestation, `payload_size`/`payload_checksum` over the canonical payload bytes, `header_checksum` over the blanked field-complete header, and `message_id` = header checksum + big-endian sequence splice at previous + 1. Mismatches drop the envelope (never-throw); delivery rows stay intact for audit. Checks are self-contained per message — stream ordering is already covered by the envelope-level chain — and zero-message envelopes (EVM epoch acks) remain valid.
- The canonical-encoding oracle moves to a shared `contracts/tests/opp_envelope_oracle.hpp`; the msgch chain, dispatch, and dispute fixtures all derive spec headers through it.
- New forgery coverage: five independent field forgeries dropped on correctly-chained envelopes, plus a resume control.

Deployment ordering: wire-solana#370 must be deployed to a shared network before this rolls out there, or its previously-empty-header envelopes drop. Ethereum already emits populated headers.

Targets master but depends on #506 (spec + depot emit) — merge #506 first; until then this PR's diff also shows #506's changes.


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -328,6 +328,84 @@ name resolve_account_from_op_address(const opp::types::ChainAddress& op_address)
...
+/// Validate every message's semantic header (opp.proto MessageHeader) against the spec
+/// derivation shared by all three emitters: each attestation's `data_size` must equal its `data`
+/// length, `payload_size` / `payload_checksum` must recompute over the payload's canonical
+/// bytes, `header_checksum` must recompute over the field-complete header with `message_id` and
+/// `header_checksum` blanked, and `message_id` must be that checksum with its first 8 bytes
+/// replaced by `previous_message_id`'s sequence number + 1.
+///
+/// Every check is SELF-CONTAINED (no cross-envelope state): stream ordering is already enforced
+/// by the envelope-level chain in `apply_consensus`, so a per-outpost inbound message tip is
+/// deliberately not tracked here. A zero-message envelope is trivially valid -- the EVM outpost
```
huangminghuang: [P1] Enforce the message-stream link to prevent replay

This validates each message only against its self-declared `previous_message_id`; it never requires that ID to equal the last accepted message for this outpost or the preceding message in the envelope. A correctly envelope-chained successor can therefore replay an earlier valid `Message` unchanged and it is accepted and dispatched. I confirmed this with a contract test: the replay created another attestation row and advanced `outpcons`. For `OPERATOR_ACTION(DEPOSIT_REQUEST)`, `depositinle` has no message-ID deduplication and credits the balance again, so one escrow deposit can be credited multiple times. Envelope chaining orders envelopes but does not enforce the separate message chain. Persist the inbound message tip per outpost, require the first message to continue it, chain subsequent messages to one another, and update the tip only after acceptance.

heifner: Fixed in 8347d1be4d. The depot now persists a per-outpost inbound message tip (`message_tip` on the `outpcons` row) and `semantic_headers_ok` requires every accepted message to continue it: the first message's `previous_message_id` must equal the stored tip, subsequent messages in an envelope chain to one another, and the tip advances only on acceptance (unmoved by zero-message acks). As with the envelope chain, the first message ever accepted for an outpost bootstraps regardless. New test `inbound_message_replay_dropped` reproduces your exact scenario — an old value-bearing Message replayed verbatim inside a correctly envelope-chained successor is now dropped before dispatch, with no new attestation row and the tip unmoved.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -944,6 +1049,20 @@ void apply_consensus(name self, uint64_t chain_code, uint32_t epoch_index,
...
+   // message must continue the per-outpost inbound message chain from `inbound_message_tip` --
+   // this is what prevents replay of an earlier valid message inside a correctly envelope-chained
+   // successor. A violation is a malformed, forged, or replayed envelope -- dropped before any
+   // dispatch, consensus record, or cleanup (delivery rows stay intact for audit/dispute), per
+   // the never-throw handler convention. `new_message_tip` becomes the outpost's message tip on
+   // acceptance (unchanged for a zero-message ack).
+   checksum256 new_message_tip{};
+   if (!semantic_headers_ok(envelope, inbound_message_tip, new_message_tip, chain_code,
+                            epoch_index)) {
+      return;
```
huangminghuang: [P1] Do not resolve a dispute when its winner is rejected

This soft return is correct on the normal delivery path, but `resolvedisp` also calls this void `apply_consensus`. `chalg::chkdispute` marks the dispute RESOLVED before invoking `resolvedisp`, then unpauses the epoch afterward. If the voted winner fails this new semantic validation, this return therefore reports success without writing `outpcons`. The epoch is then permanently stranded: `chkcons` waits for the current-epoch consensus row, while `evalcons` returns whenever a dispute row exists, including a RESOLVED one. A contract-level reproduction ends with the dispute RESOLVED and the epoch unpaused but no `outpcons`; another `evalcons` is a no-op. Please make `resolvedisp` assert that the selected winner was recorded (or otherwise propagate the validation failure), so the enclosing `chkdispute` transaction rolls back and leaves the dispute OPEN/paused.

heifner: Fixed in a260f48ceb. `apply_consensus` now returns `bool` (nodiscard) — true on acceptance or an idempotent re-fire, false on an envelope-chain / semantic-header drop. `evalcons` `(void)`-discards it (soft drop preserved — never throw on the deliver path), while `resolvedisp` `check()`s it, so a rejected voted winner asserts and rolls the whole `chkdispute` crank back: the dispute stays OPEN/paused and recoverable rather than resolve-and-stranded, the same safe-stall contract as the existing `check(found, ...)` above it. New test `chkdispute_rejects_invalid_winner_stays_open` reproduces your scenario (voted winner with a forged payload_checksum → chkdispute reverts, dispute stays OPEN, epoch paused, no outpcons row); `chkdispute_fast_path_resolves` remains the valid-winner control.

huangminghuang (approved):

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1296,9 +1428,12 @@ void msgch::evalcons(uint64_t chain_code, uint32_t epoch_index) {
...
    // Consensus reached: store + dispatch the winning envelope and record the per-outpost winner
    // (so sysio.epoch::advance can classify each delivery). advance itself is triggered by chkcons
-   // once next_epoch_start has passed.
-   apply_consensus(get_self(), chain_code, epoch_index,
-                   checksum_data[consensus_group], seen_checksums[consensus_group]);
+   // once next_epoch_start has passed. On the evalcons path a rejected winner is a SOFT drop: the
+   // result is intentionally discarded and the epoch simply stays un-consensused (no outpcons row)
+   // until a valid delivery arrives or the split opens a dispute -- never throw here, a throw
+   // reverts the delivering operator's tx and stalls consensus chain-wide.
+   (void)apply_consensus(get_self(), chain_code, epoch_index,
```
huangminghuang, line 1486: [P1] Make a rejected consensus winner recoverable

Discarding `false` here does not actually leave a path for a later valid delivery. If all operators (or a post-boundary majority) relay the same envelope with an invalid semantic header, that checksum keeps satisfying the consensus test, but `apply_consensus` never writes the current `outpcons` row. Those operators cannot submit a correction because `deliver` rejects a second row from the same operator, and any remaining minority cannot displace the existing majority; `maybe_open_dispute` is also never reached because `consensus_reached` remains true. `chkcons` then waits forever. The forgery test masks this by directly authorizing `epoch::advance` rather than exercising the production `chkcons` gate. Please quarantine/dispute/clear the rejected winner or otherwise allow replacement so this soft-drop path can recover without privileged manual state changes.

huangminghuang (commented):
One blocking availability issue remains in the rejected-winner path. The persisted-row compatibility concern was retracted because deployments are still development-stage and do not require backward-compatible state.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: Fixed in f141560a28 by validating the semantic header + inbound chain at **ingress** rather than only at consensus. `deliver` now runs the same `inbound_envelope_valid` check (envelope-chain + `semantic_headers_ok`) and reverts on failure — factored into a helper shared with `apply_consensus`. So a forged/malformed/stale envelope is never recorded, can't reach the consensus tally, and can't leave a phantom `consensus_reached` with no `outpcons` row; a reverted tx records nothing, so the operator can immediately re-deliver a corrected envelope. The check runs *after* the duplicate check, so a same-operator re-delivery still reports "already delivered". `apply_consensus` keeps the check as defense in depth (soft-drop on the evalcons path, `check()` on `resolvedisp`).

New `forged_delivery_does_not_strand_chkcons` exercises the production `chkcons` gate (not a direct `epoch::advance`): a forged delivery reverts, a valid delivery reaches consensus, and `chkcons` proceeds to attempt advance — proving no stall. The forgery/replay/cross-stream tests now assert the ingress revert, and `chkdispute_rejects_invalid_winner_stays_open` is superseded by `forged_dispute_winner_rejected_at_delivery` (a forged winner can no longer be delivered, so it can never reach `resolvedisp` — that guard remains as defense in depth).

PR #510, opened 2026-07-10
SEC-67: Bound external debugging backpressure

huangminghuang:
## Summary

- bound the external debugging sink to a configurable number of pending OPP envelopes
- apply a configurable deadline to debugging-server validation and envelope delivery requests
- make queue admission, drop accounting, and shutdown safe under concurrent signal delivery
- preserve hostname configurations while preventing runtime requests from re-entering potentially unbounded DNS resolution

Jira: [SEC-67](https://wire-network.atlassian.net/browse/SEC-67)

## Root cause

The batch-operator debugging signal fed full OPP envelopes into an unbounded worker queue. If the optional debugging server accepted a connection but stopped responding, the single delivery worker remained occupied while subsequent envelopes accumulated without a memory bound.

## Changes

- add `--ext-debugging-max-pending-envelopes` with a default of 16 waiting envelopes
- add `--ext-debugging-request-timeout-ms` with a default of 5000 milliseconds
- add optional bounded admission and `try_push` support to `worker_task_queue` without changing existing unbounded callers
- introduce a dedicated debugging event sink that logs/counts rejected and shutdown-discarded envelopes
- disconnect, stop, join, and discard pending work safely during plugin shutdown
- add a JSON-RPC endpoint refresh policy; the debugging client resolves its configured hostname at startup and reuses that endpoint set for bounded runtime behavior
- add queue, transport policy, configuration, backpressure, drop-accounting, and shutdown-lifetime tests

## Operational impact

- the default memory bound is 16 waiting envelopes plus at most one active delivery
- a stalled debugging-server request releases the worker after five seconds by default
- existing hostname URLs such as `localhost` remain valid
- DNS changes to the debugging hostname require restarting `nodeop`
- both new limits are configurable and reject zero values

## Validation

- Release build targets: `test_fc`, `test_external_debugging_plugin`, and `nodeop`
- `test_fc`, `test_batch_operator_plugin`, and `test_external_debugging_plugin`: 3/3 passed
- focused `worker_task_queue_tests`: 12/12 passed
- `git diff --check`
- final review-specific agent pass: no remaining actionable findings

[SEC-67]: https://wire-network.atlassian.net/browse/SEC-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/libraries/libfc/include/fc/network/json_rpc/json_rpc_client.hpp
+++ b/libraries/libfc/include/fc/network/json_rpc/json_rpc_client.hpp
@@ -23,6 +23,9 @@ namespace asio = boost::asio;
 // -----------------------------------------------------------------------
 enum class http_verb { GET, PUT, POST, DELETE_ };
 
+/// Control whether connection failures invalidate the client's startup DNS result.
+enum class endpoint_refresh_policy { on_connection_failure, never };
+
 // JSON-RPC error type
 struct json_rpc_error : fc::exception {
    int      code;
@@ -35,8 +38,8 @@ struct json_rpc_error : fc::exception {
 // raw HTTP verb support for REST-style endpoints.
 class json_rpc_client {
 public:
-   explicit json_rpc_client(fc::url                           url,
-                            const std::optional<std::string>& user_agent = std::nullopt);
+   explicit json_rpc_client(fc::url url, const std::optional<std::string>& user_agent = std::nullopt,
+                            endpoint_refresh_policy refresh_policy = endpoint_refresh_policy::on_connection_failure);
 
    // -----------------------------------------------------------------------
    //  JSON-RPC 2.0 methods (unchanged, backwards compatible)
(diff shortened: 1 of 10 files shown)


heifner (approved):
Approving. Clean, well-layered fix for the unbounded-queue DoS — a bounded pending queue (excess dropped and counted) plus a per-request deadline that releases a stalled worker. The shutdown/emission lifetime is the subtle part and it's handled correctly: the signal slot owns a `shared_ptr` to the sink so an emission racing with shutdown stays safe on a path that never touches the worker's raw `impl`, the worker is always joined before `_impl` teardown, and `std::call_once` makes `stop()` idempotent — with a dedicated test pinning it. As a bonus, shutdown is now bounded where it could previously hang forever.

A few non-blocking notes for follow-up consideration:

1. Startup DNS resolution in the `json_rpc_client` constructor is still unbounded (resolves with no deadline). Runtime re-resolution and per-request transport are now bounded, but a hung DNS lookup at startup would still block `plugin_startup`. Might be worth having the startup resolve honor `request_timeout_ms` as well.

2. `validate_server()` is now deadline-bounded at startup — a net improvement (fail-fast over hang), just a behavior change worth confirming 5s is enough headroom for a slow-but-reachable server.

3. The `never` refresh policy means an IP change for the debug host requires a nodeop restart; drops are counted and logged so it's observable, and this matches the documented intent.

Minor: `discard_pending()` has a redundant local, and `_pending_capacity` duplicates the queue's `max_pending_items` for the drop log — neither blocking.

PR #512, opened 2026-07-13
SEC-125: Bound startup DNS resolution

huangminghuang:
## Summary

- bound deadline-aware hostname resolution on a single process-lifetime Boost.Asio resolver worker, without adding a DNS dependency
- limit platform resolution to one in-flight lookup and keep late completion state independent of the caller/client lifetime
- apply the configured external-debugging startup deadline across both client construction and server validation
- cover stalled and late completion, expired and consumed startup budgets, successful completion, and a real deadline-bound localhost lookup

## Validation

- `json_rpc_client_tests`
- affected `test_fc`, `test_batch_operator_plugin`, and `test_external_debugging_plugin` targets
- broad Release CTest sweep: 333/333 passed
- GitHub CI: all required checks passed
- system-contract e2e gate: not applicable

## Risk

A timed-out platform `getaddrinfo` may finish in the background because it cannot be portably interrupted. A single in-flight permit prevents queued lookup growth, and the resolver runtime intentionally survives process teardown so shutdown cannot rejoin a stalled lookup.

### DIFF
--- a/libraries/libfc/CMakeLists.txt
+++ b/libraries/libfc/CMakeLists.txt
@@ -74,7 +74,6 @@ set(fc_sources
 )
 
 find_package(fmt CONFIG REQUIRED)
-find_package(c-ares CONFIG REQUIRED)
 
 file(GLOB_RECURSE fc_headers src/*.hpp src/*.h include/*.hpp include/*.h)
 add_library(fc ${fc_sources} ${fc_headers})
@@ -125,8 +124,6 @@ target_link_libraries(
         fmt::fmt
         ${PLATFORM_SPECIFIC_LIBS}
         ${CMAKE_DL_LIBS}
-        PRIVATE
-        c-ares::cares
 )
 
 target_compile_definitions(fc PUBLIC -DFC_AVAILABLE=1)
--- a/libraries/libfc/CMakeLists.txt
+++ b/libraries/libfc/CMakeLists.txt
@@ -74,6 +74,7 @@ set(fc_sources
 )
 
 find_package(fmt CONFIG REQUIRED)
+find_package(c-ares CONFIG REQUIRED)
 
 file(GLOB_RECURSE fc_headers src/*.hpp src/*.h include/*.hpp include/*.h)
 add_library(fc ${fc_sources} ${fc_headers})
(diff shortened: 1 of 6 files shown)


heifner (approved):
Solid, well-designed fix for the unbounded startup DNS hang, with genuinely thorough tests and green CI (asan/ubsan/gcc/asserton).

I traced the one thing that looked concurrency-sensitive — `resolver_cancel_guard` calling `resolver->cancel()` from the caller thread while the resolve runs on asio's resolver thread pool — and it's **race-free by construction**. `basic_resolver::cancel()` is just `impl.reset(nullptr, noop_deleter())` on a `shared_cancel_token_type` (`shared_ptr<void>`), and the background `resolve_query_op` holds its own `weak_cancel_token_type` copy passed to `background_getaddrinfo`. The only cross-thread shared state is the shared_ptr's atomic control block, so the caller-side `cancel()` and the resolver-thread token check never touch the same mutable object — that's exactly the pattern asio's cancellation token exists for. (For a genuinely stuck `getaddrinfo` it's a no-op anyway, as the Risk section notes.)

Lifetime / exception-safety holds up too: `deadline_resolution_state` and `platform_resolver_permit` are `shared_ptr`-captured by the async completion, so they correctly outlive a timed-out caller (nicely covered by `initial_endpoint_resolution_cancels_stalled_query` invoking `late_completion` after the client is gone); the `cancel_guard` arms before the wait, so a starter that consumes the deadline still cancels (`..._cancels_after_starter_consumes_deadline`); and the intentional `resolver_runtime()` leak is the right call to avoid a teardown join on a stuck lookup.

Non-blocking:

- `(void)remaining_until(deadline_abs, OP_RESOLVE);` used purely for its throw-on-expiry side effect appears ~4×, and the first one in `start_platform_resolution` is redundant with the `remaining_until` in the permit-ctor argument (which also throws). A named `throw_if_deadline_expired(...)` would read clearer and let you drop the redundant call.
- Bounding is opt-in (only when a `deadline_scope` is active) and only `external_debugging_plugin` is wrapped. Worth a one-line note on whether other startup `json_rpc_client` paths (e.g. `batch_operator_plugin`) are intentionally out of scope for SEC-125.
- Worth stating explicitly that a permanently-stuck `getaddrinfo` holds the single permit for the process lifetime, so subsequent resolutions fail-fast rather than hang — the intended degradation.

PR #514, opened 2026-07-14
Cmake: re-glob test sources with CONFIGURE_DEPENDS

heifner:
Adding a new test file today requires remembering to rerun cmake by hand: the test-suite source lists come from configure-time `file(GLOB)`, so an incremental ninja build silently skips the new file — local build green, test never ran, CI fails later.

This adds `CONFIGURE_DEPENDS` to every in-tree test/benchmark source glob (unittests, plugin tests, contracts tests, wasm-spec generated tests, chaindb/custom_appbase/libfc-lite/opp tests, benchmark). Ninja now re-checks the globbed directories on each build and regenerates automatically when files appear or disappear. The per-build cost is a stat of a dozen directories. `libraries/appbase` is excluded (submodule); vendored wasm-jit is untouched.

Adds ~7ms to a build.


plugins/http_plugin/test/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-file(GLOB UNIT_TESTS "*.cpp")
+file(GLOB UNIT_TESTS CONFIGURE_DEPENDS "*.cpp")
```
huangminghuang, line 1: The shared plugin test path is still configure-time-only. `plugin_target()` uses `file(GLOB_RECURSE TEST_FILES ${CMAKE_CURRENT_SOURCE_DIR}/test/*.cpp)` in `cmake/plugin-tools.cmake` (and another glob for `ARG_TEST_SOURCE_GLOBS`) without `CONFIGURE_DEPENDS`. As a result, newly added tests remain silently excluded for the batch-operator, cron, outpost-client, Ethereum/Solana outpost-client, and underwriter plugin suites. Please add `CONFIGURE_DEPENDS` to both shared test globs as part of this coverage fix.

heifner: Good catch — the sweep only covered CMakeLists.txt files and missed the .cmake module. Both plugin_target test globs fixed in 1b70bbf2f3. Note the same helper's src/*.cpp and SOURCE_GLOBS globs (lines 38/45) share the gap for plugin sources; left out here to keep this PR scoped to test sources.

cmake/plugin-tools.cmake
```diff
@@ -87,14 +87,14 @@ macro(plugin_target TARGET_NAME)
 
             set(TEST_FILES "")
             if (IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/test)
-                file(GLOB_RECURSE TEST_FILES ${CMAKE_CURRENT_SOURCE_DIR}/test/*.cpp)
+                file(GLOB_RECURSE TEST_FILES CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/test/*.cpp)
```
huangminghuang, line 90: One active plugin test glob is still configure-time-only: `plugins/signature_provider_manager_plugin/ssm/test/CMakeLists.txt` starts with `file(GLOB_RECURSE SRC_FILES *.cpp)`. The SSM test directory is added whenever `ENABLE_TESTS` is on, so adding a new SSM test `.cpp` would still be silently skipped by incremental builds. Please add `CONFIGURE_DEPENDS` there as well.

heifner: Fixed in 02820923d2. Also re-ran the full sweep on this branch (CMakeLists.txt + .cmake modules): the only remaining non-CONFIGURE_DEPENDS globs are a single-fixed-filename glob (chaindb benchmark's bench.cpp), library-source globs (fc-test, sysio_testing — a missing file there breaks the link rather than silently skipping), and vendored wasm-jit .wast data files. Nothing else in the silent-test-skip class.

huangminghuang, line 90: The repository-wide sweep still finds one source of the same test-skipping behavior: `tools/plop/templates/create-cxx-library/CMakeLists.txt.hbs` emits `file(GLOB_RECURSE TEST_FILES test/*.cpp test/*.hpp)` without `CONFIGURE_DEPENDS`. Every library created with the active `create-cxx-library` generator would therefore retain this bug. Please update the template glob too. I checked all other test/benchmark glob sites; after the plugin-target and SSM follow-ups, this is the only remaining PR-scope omission.

heifner: Fixed in 42ac7ff1f9 — the generator template now emits CONFIGURE_DEPENDS so scaffolded libraries get the fix for free. Re-swept extension-agnostically this time (all files, not just CMakeLists.txt/.cmake): the only other file(GLOB) hits near test keywords are SysioTester.cmake.in's installed-artifact globs (package-config context, not test sources), so I agree this closes out the PR scope.

huangminghuang (approved):

PR #515, opened 2026-07-14
Opreg: prune delivery-log rows that age out of the termination window

heifner:
`recorddel` adds one `dellog` row per group member per epoch (ram payer `sysio`) and nothing ever erases them — `prune()` sweeps terminated operators only. Rows older than `terminate_window_ms` are invisible to `termcheck`'s window scan, so they accumulate as permanent dead weight in the RAM pool. Surfaced as a follow-up during the PR #493 review.

This implements the retention the table's doc comment already promised ("rows older than the window are discarded by `prune` / on-write"): each `recorddel` sweeps up to `MAX_DELLOG_PRUNE_PER_WRITE` (4) expired rows before inserting, and `prune()` sweeps up to `MAX_DELLOG_PRUNE_PER_CRANK` (64) per crank. `log_id` allocation order matches `ts_ms` order, so the primary-index walk is oldest-first and stops at the first still-in-window row — cost is O(erased), never O(table). The window-open computation is a shared helper now also used by `termcheck`.

Also decouples `MAX_TERMINATE_MAX_CONSECUTIVE_MISSES` from its `DEFAULT_` alias and adds `static_assert`s pinning both termination defaults inside their accepted bounds, so bumping a production default can never silently widen the setconfig ceiling.

No ABI change — `sysio.opreg.abi` is byte-identical; the rebuilt `sysio.opreg.wasm` is included per the checked-in-wasm convention.


contracts/tests/sysio.opreg_tests.cpp
```diff
@@ -861,6 +873,115 @@ BOOST_FIXTURE_TEST_CASE(termcheck_keeps_bootstrapped_operator_exempt, sysio_opre
...
+   // Second write clears the remaining two and stops at the in-window row.
+   BOOST_REQUIRE_EQUAL(success(), recorddel("batchop.a"_n, 8, /*delivered=*/true));
+   produce_blocks();
+   BOOST_REQUIRE(get_dellog_entry(5).is_null());
+   BOOST_REQUIRE(get_dellog_entry(6).is_null());
+   BOOST_REQUIRE(!get_dellog_entry(7).is_null());
+   BOOST_REQUIRE(!get_dellog_entry(8).is_null());
+} FC_LOG_AND_RETHROW() }
+
+BOOST_FIXTURE_TEST_CASE(prune_sweeps_expired_dellog_rows, sysio_opreg_tester) { try {
```
huangminghuang, line 933: P3: Pin the crank’s 64-row bound. This test proves that `prune()` removes five expired rows, but it never exercises `MAX_DELLOG_PRUNE_PER_CRANK = 64`. A regression that removes the bound or introduces an off-by-one would still pass. Please seed at least 65 expired rows, assert one crank removes exactly 64 and leaves the remainder, then verify a second crank clears it.

heifner: Done in 98f6b3d730 — the case now seeds kDellogPrunePerCrank + 1 (65) expired rows, asserts the first crank erases exactly 64 oldest-first with row 65 surviving, and a second crank clears it. Full contracts_unit_test green.

huangminghuang (approved):

PR #516, opened 2026-07-14
Add bounded council election contract

huangminghuang:
## Summary

Introduce the `sysio.councl` system contract, which fills 21 council seats through a bounded tier-1 → tier-2 → tier-3 nomination and strict-priority voting process.

The contract:

- maintains a self-paid, capped council-candidate registry;
- freezes an explicitly ordered 21-owner tier-1 roster plus complete tier-2/tier-3 escalation snapshots from `sysio.roa`;
- lets each seat owner nominate first, then escalates failed or timed-out attempts through selected tier-2 and non-repeating tier-3 proposers;
- elects from three-candidate slates using strict candidate priority and greater-than-two-thirds thresholds;
- uses a deterministic entropy accumulator and a lazy Fisher–Yates remap to bound tier-3 selection storage and prevent repeats within a seat;
- provides governance recovery through `forceback`, `forceassign`, and active-election reset; and
- retains completed council generations while purging bounded ephemeral state between elections.

## Safety, lifecycle, and review hardening

- Make generation-scoped `sysio.roa::nodeowners` the authority for tier membership and counts. ROA now enforces the shared tier caps directly, including before emissions configuration, while `sysio.system::nodecount` remains an optional emissions-distribution mirror.
- Add canonical ROA accessors for the active generation and tier counts. Council completeness uses the authoritative generation-scoped count; `sysio.chalg` uses the shared generation accessor when freezing its newer per-dispute Tier-1 electorate and quorum from `master`.
- Make `loadtier(tier, max_rows)` bound source reads as well as writes with persistent primary-key cursors. Identity deduplication and explicit subsequent passes absorb earlier-sorting ROA additions without quadratic wrap scans.
- Require both tier scans to complete and verify snapshot sizes against the captured ROA generation before opening the election.
- Split cleanup into initialization-abort, active-abort, and completed-generation modes. Initialization aborts preserve candidate-paid rows in the same generation; active aborts remove partial council results and advance the generation; completed cleanup retains council history.
- Permit governance to reset any active READY election, clear the live `state` singleton when cleanup completes, keep final `active_seat` within its documented 0–20 range, and diagnose unknown cleanup modes/stages explicitly.
- Add optional `expected_round` binding to `repcandidate` and `vote`. Bound actions fail and roll back if lazy settlement crosses the round boundary; omitted values retain the existing settlement-only behavior. Explicit inclusive deadline checks protect both payload paths.
- Add compile-time guards that preserve equivalent T1 and proposer-auto-yes thresholds if seat constants change, and avoid iterator recreation during batched erasure.
- Consolidate ABI loading, action pushing, and the 19-field emissions fixture in shared system-contract test support.
- Document the sharper entropy limitation: an actor controlling several eligible identities can simulate the deterministic post-action selection offline and submit only a favorable identity in the same transition-triggering action.

## Base integration and generated artifacts

- Merge current `master` (`79a2e8e335`) and reconcile the newer frozen dispute-electorate implementation with this PR's canonical ROA accessors.
- Rebuild and commit `sysio.councl.abi` plus the `sysio.councl`, `sysio.roa`, and merged `sysio.chalg` WASM artifacts.
- Regenerate the merged block-log and snapshot reference files and update the expected snapshot head ID.
- Keep the three changed contract artifacts byte-for-byte identical to their clean `contracts_project` outputs.

## Tests and validation

- Clean `contracts_project` build — all 48 CDT contract artifacts passed.
- Merged focused contract suite — 74 council-math, council-integration, and dispute cases passed.
- Council coverage includes bounded source scanning, ROA churn/rescan, no-emissions initialization, authoritative tier caps, fail-loud round binding, active abort, loading-abort candidate preservation, final-state cleanup, and complete 21-seat runs.
- Dispute coverage includes the `master` electorate snapshot/quorum behavior, fixed in-flight eligibility, concurrent-dispute pause handling, and semantic-header validation.
- Reference validation — `deep_mind_tests`, all 8 snapshot part-2 cases, Savanna consensus snapshot compatibility, and `sysio_util_snapshot_info_test` passed.
- Generated-artifact hash comparison and `git diff --check origin/master` passed.

## Reviewer notes

- Entropy is deterministic and auditable, but not an ungrindable beacon. Caller identity and timing remain influenceable; removing that class of manipulation requires a future-block VRF or equivalent source. The same-call multi-identity simulation path is explicitly accepted and documented.
- A one-member tier reaches threshold through the proposer auto-yes. Because node owners may also be candidates, that proposer can nominate and immediately seat themselves; the contract intentionally imposes neither an electorate-size floor nor proposer/candidate exclusion.
- Automatic escalation is bounded, but final completion can require governance `forceassign`; the contract does not promise unconditional wall-clock completion when governance is unavailable.
- `expected_round` is opt-in for compatibility. Unbound late nominations and ballots continue to act as successful settlement-only cranks rather than applying stale payloads.


brianjohnson5972 (dismissed):
Looks good

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+   config_state cfg = cg.get("contract not initialized");
+   check(cfg.init_phase == councl::init_phase::LOADING, "not in the loading phase");
+
+   // The snapshot scope and tier values together guarantee identity as well as count completeness.
+   check(cfg.network_gen == roa_network_gen(), "roa network generation changed during initialization");
+   const uint32_t c2 = tier_count(councl::election_tier::T2);
+   const uint32_t c3 = tier_count(councl::election_tier::T3);
+   check(c2 <= sysiosystem::emissions::T2_MAX_NODE_OWNERS, "tier-2 count exceeds the system owner cap");
+   check(c3 <= sysiosystem::emissions::T3_MAX_NODE_OWNERS, "tier-3 count exceeds the system owner cap");
+   check(cfg.t2_loaded == c2, "tier-2 snapshot incomplete");
```
heifner, line 479: **[P1]** `finalizeinit` gates snapshot completeness on `sysio.system::nodecount`, but that counter is cumulative/unscoped and only written when emitcfg exists, while the snapshot counts per-network_gen roa rows — the two diverge and permanently brick init.

A tier-2 owner forcereg'd during bootstrap before `setemitcfg` lands in `roa::nodeowners`, but `nodecount.t2_count` is never incremented (`addnodeowner` in sysio.roa gates the counter update on `emitcfg.exists()`). `loadtier` then snapshots N rows so `cfg.t2_loaded == N`, `tier_count(T2)` returns N-1, and this check aborts with "tier-2 snapshot incomplete". `loadtier` only appends; `reset()` + `purge()` bumps the generation and re-snapshots the identical N-vs-(N-1) state; and sysio.system exposes no action to repair nodecount — so no council election can ever finalize. The same equality is also unsatisfiable for every generation after any future roa network_gen bump (per-generation snapshot vs all-time counter).

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+   check(max_rows > 0, "max_rows must be positive");
+
+   config_t cg(get_self());
+   config_state cfg = cg.get("contract not initialized");
+   check(cfg.init_phase == councl::init_phase::CLEANING, "generation cleanup is not active");
+
+   uint32_t remaining = max_rows;
+   while (remaining > 0 && cfg.cleanup_stage != councl::cleanup_stage::COMPLETE) {
+      uint32_t erased = 0;
+      switch (cfg.cleanup_stage) {
```
heifner, line 537: **[P1]** This switch has no `default` arm, so any out-of-range `cleanup_stage` byte leaves `erased == 0` and the surrounding `while (remaining > 0 && stage != COMPLETE)` loop never decrements `remaining` nor reaches COMPLETE — an unbreakable infinite loop with no exit from CLEANING.

An in-place upgrade that renumbers or inserts a cleanup_stage enumerator (or any storage corruption) while a generation is mid-CLEANING makes every `purge` call burn the full CPU deadline and revert; `reset()` requires init_phase LOADING or READY, so CLEANING has no other exit and the contract is permanently stuck with no governance recovery. A `default: check(false, "unknown cleanup stage")` arm turns this into a diagnosable assert.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+   // Enumerate roa's tier-1 owners (the authoritative set the ordering must permute).
+   roa::nodeowners_t no(councl::ROA_ACCOUNT, ng);
+   auto t1_idx = no.get_index<"bytier"_n>();
+   std::vector<name> t1;
+   for (auto it = t1_idx.lower_bound(t1_raw); it != t1_idx.end(); ++it) {
+      if (it->tier != t1_raw)
+         break;
+      t1.push_back(it->owner);
+   }
+   check(t1.size() == councl::SEATS, "roa tier-1 owner count does not match the council seat count");
```
heifner, line 389: **[P1]** This requires roa to hold exactly 21 tier-1 owners, but the only thing capping the tier-1 population is `addnodeowner`'s check against the same lagging `nodecount` counter — so roa can come to hold 22+ tier-1 owners and `startinit` becomes permanently unsatisfiable.

On a chain where one tier-1 owner was registered pre-emitcfg (nodecount.t1_count == 20 while roa holds 21 tier-1 rows — exactly the state this PR's test fixture is in), `addnodeowner`'s `t1_count < T1_MAX_NODE_OWNERS` still has headroom, so a 22nd `forcereg(_, 1)` succeeds. The bytier scan then yields t1.size() == 22, this check throws for every possible `ordered_owners`, and roa has no nodeowners removal action — the election can never be started.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+void council::reset() {
+   require_auth(get_self());
+   config_t cg(get_self());
+   config_state cfg = cg.get("contract not initialized");
+   const bool aborting_initialization = cfg.init_phase == councl::init_phase::LOADING;
+   if (!aborting_initialization) {
+      check(cfg.init_phase == councl::init_phase::READY,
+            "reset requires a loading or completed election generation");
+      state_t sg(get_self());
+      check(!sg.exists() || sg.get().phase == councl::election_phase::DONE, "current election is not complete");
```
heifner: **[P1]** A running (READY, non-DONE) election cannot be aborted — this guard rejects any phase other than DONE — and `time_slot_sec` is written only by `startinit` (REG) and purge-completion (CLEANING), so a mis-set slot locks the chain into an unabortable forced march of up to 21 x 30 days.

If governance calls `startinit(2592000, ...)` intending 259200, every window is now 30 days; `forceback` requires the attempt to have elapsed and each `forceassign` resets round_open_ts, so reaching DONE takes >= 630 wall-clock days of governance actions, each writing a permanent council_row — no reset, no abort, no way to shorten the slot or discard the bad generation without polluting the retained history. An authorized abort action (or making time_slot_sec adjustable mid-generation) closes this.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+// ===========================================================================
+uint8_t council::roa_network_gen() const {
+   roa::roastate_t rs(councl::ROA_ACCOUNT);
+   check(rs.exists(), "roa state not initialized");
+   return rs.get().network_gen;
+}
+
+uint32_t council::tier_count(councl::election_tier tier) const {
+   sysiosystem::emissions::nodecountstate_t nc(councl::SYSTEM_ACCOUNT);
+   check(nc.exists(), "sysio.system nodecount not initialized");
```
heifner: **[P1]** `tier_count()` hard-requires the `nodecount` singleton to exist, with no fallback to the roa rows just snapshotted, so a valid election with zero tier-2/tier-3 owners (or a chain where emissions was never configured) cannot finalize.

A network with a complete 21-owner tier-1 roster and no tier-2/tier-3 owners never triggers `addnodeowner`, so the lazily-created singleton does not exist; `finalizeinit` then aborts here with "sysio.system nodecount not initialized" even though `cfg.t2_loaded == cfg.t3_loaded == 0` is the correct, complete answer — blocking an otherwise fully-specified election.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+         }
+         break;
+      }
+      case councl::cleanup_stage::COMPLETE:
+         break;
+      }
+      remaining -= erased;
+   }
+
+   if (cfg.cleanup_stage == councl::cleanup_stage::COMPLETE) {
```
heifner, line 603: **[P2]** `purge()` never clears the unscoped `state` singleton (it is only ever `.set()` in `finalizeinit`, never `.remove()`), so after a generation is purged the contract keeps advertising the previous election's terminal state through the next generation's REG and LOADING.

After gen 0 completes -> reset -> purge -> gen 1 REG, the state row still reads phase=DONE, active_seat=21, seats_filled=21, with the gen-0 slate/tallies/acc and a stale voted_bitmap; election_state carries no generation stamp, so an explorer/wallet/monitor cannot tell it is stale and reports gen 1 as a finished election until finalizeinit overwrites it. reset()'s phase==DONE guard is likewise evaluated against the prior generation's row, making it load-bearing only in appearance — and active_seat=21 is already outside its own documented 0..20 range. Clearing the singleton in the COMPLETE arm (or stamping election_state with election_gen) fixes both.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+         return;
+      const uint64_t snapshot_index = static_cast<uint64_t>(already) + written;
+      table.emplace(RAM_PAYER, council::index_key{snapshot_index}, Row{snapshot_index, owner.owner});
+      ++written;
+   };
+
+   // The primary nodeowners key is owner identity. Resume immediately after the last appended
+   // owner, then wrap once so owners inserted earlier during a prior batch are still absorbed.
+   auto it = already == 0 ? owners.begin() : owners.upper_bound(roa::nodeowner_key{resume_after.value});
+   for (; it != owners.end() && written < max_rows; ++it)
```
heifner: **[P2]** `max_rows` bounds writes but not reads: this forward pass scans the entire roa nodeowners scope (all three tiers), and the wrap pass rescans it a second time, so `loadtier`'s per-transaction cost is independent of its argument — the comment about "avoiding a full tier-prefix rescan" is backwards.

At the system caps (21 T1 + 84 T2 + 1000 T3 = 1105 rows in one network_gen scope), a `loadtier(3, 1)` call — including the idempotent "is it done yet" final call, where written == 0 < max_rows so the wrap fires — walks ~1105 primary rows forward plus ~1105 on the wrap, each with a full 4-asset row deserialize plus a byowner secondary find: ~6300 KV intrinsics to append zero rows. Shrinking max_rows (the only knob) does not reduce the cost and makes the wrap more likely. The resume cursor is also the last-appended owner, so once a wrap absorbs a low-sorting owner the cursor regresses and every subsequent batch re-scans nearly the whole table, turning a churned batched load quadratic. Bounded at today's caps, but a hard brick if T3_MAX_NODE_OWNERS is ever raised.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+void council::settle(name caller) {
+   require_auth(caller);
+   config_t cg(get_self());
+   config_state cfg = cg.get("contract not initialized");
+   check(cfg.init_phase == councl::init_phase::READY, "election is not running");
+
+   state_t sg(get_self());
+   election_state st = sg.get("election state missing");
+   check(st.phase != councl::election_phase::DONE, "election is complete");
+   do_stir(st, ACTION_SETTLE, caller);
```
heifner, line 743: **[P2]** Proposer selection is grindable within a single transaction by a single account: settle/stir/vote/repcandidate mix a caller-chosen name into the entropy accumulator immediately before the same call's selection draw, with no commit-reveal, future-block dependence, or rate limit.

settle does do_stir(st, ACTION_SETTLE, caller) then resolve_or_settle -> fail_attempt -> select_tier2_proposer, where selection_index hashes (acc, active_seat, round_id) and the new acc = sha256(pack(acc_prev, "settle", caller, stir_count + 1)) — every input is public state plus one attacker-chosen name. An attacker with K identities precomputes offline which caller value selects a colluding tier-2/tier-3 owner (n2 <= 84, n3 <= 1000 draws), then submits only that settle in the first block past the deadline, steering which account nominates the 3-candidate slate for every escalated seat (tier-3 grants an extra draw per retry). The PR discloses grinding as a sign-off item, but this single-account/single-transaction offline-simulation path is sharper than "choosing when to crank" — worth stating explicitly in the disclosure, and a cheap partial mitigation exists (e.g. only electorate members' calls feed the accumulator).

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+   config_state cfg = cg.get("contract not initialized");
+   check(cfg.init_phase == councl::init_phase::READY, "election is not running");
+
+   state_t sg(get_self());
+   election_state st = sg.get("election state missing");
+   do_stir(st, ACTION_VOTE, voter);
+   const uint64_t prior_round = st.round_id;
+   const auto prior_phase = st.phase;
+   resolve_or_settle(st, cfg);
+   if (st.round_id != prior_round || st.phase != prior_phase) {
```
heifner, line 699: **[P2]** A ballot that lands one block past the deadline returns transaction SUCCESS while silently discarding the caller's input, and the action exposes no round/nonce to bind the vote to the round it was cast for.

A voter signing at vote_deadline - 1s whose transaction lands at vote_deadline + 1s triggers resolve_or_settle -> try_resolve -> FAIL -> fail_attempt (round_id increments), this early return fires, the tallies and bitmap bit are never applied, and the voter gets a success receipt indistinguishable from a counted vote. With no expected-round parameter, an automated voting client cannot make the ballot conditional — silently disenfranchising voters at every deadline boundary. The settle-on-late-arrival design itself is tested as intended; the gap is the success-receipt ambiguity. An optional expected_round parameter (checked after settlement) would let clients opt into fail-loud. Same applies to repcandidate.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+   bool drained;
+};
+
+/// Erase at most `max_rows` rows and report both the number removed and whether the table drained.
+template <typename Table>
+erase_result erase_rows(Table& table, uint32_t max_rows) {
+   uint32_t erased = 0;
+   auto it = table.begin();
+   while (it != table.end() && erased < max_rows) {
+      it = table.erase(it);
```
heifner: **[P2]** `erase_rows` erases through an lvalue iterator; `kv::table_impl::erase` takes its const_iterator by value, and the iterator copy constructor performs a fresh kv_it_create plus a full lower_bound re-seek — roughly doubling the host-call cost of every purge pass.

A `purge(2000)` draining a full generation (1000 candidates + 21 roster + 84 tier2 + 1000 tier3 + remap) pays create + lower_bound + next + key + value + erase (+ kv_idx_remove) per row instead of next + erase on one long-lived iterator, so the action does ~2x the intrinsics it was sized for and can exceed max_transaction_cpu_usage, reverting with an opaque deadline error rather than a contract assert. `auto k = it.key(); ++it; table.erase(k);` (or `table.erase(std::move(it))`) avoids the copy.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+      auto cr = cands.get(cand_key{c.value}, "candidate not registered");
+      check(!cr.elected, "candidate already elected to a seat");
+   }
+
+   // Open the voting round; tier-2/3 seed the proposer's auto-yes on all three candidates.
+   st.c1 = c1;
+   st.c2 = c2;
+   st.c3 = c3;
+   uint32_t bitmap_members = 0;
+   if (st.tier == councl::election_tier::T1) {
```
heifner, line 663: **[P3]** The T1 electorate rule (elect_N = T1_VOTERS = 20, no auto-yes) and the T2/T3 rule (elect_N = N including the proposer, auto-yes seeded) are hand-coded separately and are numerically equivalent only at the current SEATS = 21; the static_assert(T1_VOTERS == SEATS - 1) does not catch the threshold divergence.

floor(2n/3) steps at 2 of every 3 n, so at SEATS = 22 the two rules diverge: T1 would require 15 real yes votes (8 NOs to veto) while T2/T3 require 14 (7) — silently making tier 1 a stricter supermajority than tiers 2/3, with no compile error and no test coverage if emissions::T1_MAX_NODE_OWNERS is ever changed. Present-day wrinkle: elect_N means "voter count" for T1 but "voter count + 1" for T2/T3, so any turnout computed from elect_N is off by one for T2/T3.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+   config_state cfg = cg.get("contract not initialized");
+   check(cfg.init_phase == councl::init_phase::READY, "election is not running");
+
+   state_t sg(get_self());
+   election_state st = sg.get("election state missing");
+   do_stir(st, ACTION_REPCANDIDATE, proposer);
+   const uint64_t prior_round = st.round_id;
+   const auto prior_phase = st.phase;
+   resolve_or_settle(st, cfg);
+   if (st.round_id != prior_round || st.phase != prior_phase) {
```
heifner, line 639: **[P2]** The explicit propose/vote deadline check was removed in favor of this implicit (round_id, phase)-changed early return, so deadline enforcement now depends entirely on the unasserted invariant that every settlement transition mutates round_id or phase.

Today every fail_attempt/advance_seat path bumps round_id (via open_tier_attempt) or sets BACKSTOP/DONE, so a late repcandidate/vote is caught — but nothing asserts this, and DESIGN.md still specifies an explicit `check(now <= round_open_ts + time_slot_sec)`. Any future same-round retry, or a BACKSTOP-to-BACKSTOP transition that leaves both unchanged, would silently admit a stale nomination/ballot as on-time — and this guard is the sole deadline enforcement in both repcandidate and vote. Worth restoring an explicit deadline check, or at minimum asserting the transition invariant where settlements are applied.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+// ===========================================================================
+void council::do_stir(election_state& st, name action_tag, name actor) {
+   ++st.stir_count;
+   st.acc = hash_args(st.acc, action_tag, actor, st.stir_count);
+}
+
+// ===========================================================================
+//  Cross-contract reads (sysio.roa / sysio.system)
+// ===========================================================================
+uint8_t council::roa_network_gen() const {
```
heifner, line 159: **[P3]** `roa_network_gen()` and `tier_count()` re-implement the roastate network_gen and nodecount reads already hand-rolled in sysio.chalg — which, notably, treats a missing nodecount as 0 where this contract check()-aborts.

If roa gains a second live generation or node_count_state changes shape, both hand-rolled readers must be edited in lock-step; a partial fix leaves chalg validating a disputant against generation N while councl froze its electorate from N-1. A shared inline accessor in sysio.roa.hpp / emissions.hpp (the documented cross-contract read points) would give one canonical answer and remove the divergence surface behind the nodecount findings on startinit/finalizeinit.

contracts/sysio.councl/src/sysio.councl.cpp
```diff
@@ -0,0 +1,744 @@
...
+   const bool aborting_initialization = cfg.init_phase == councl::init_phase::LOADING;
+   if (!aborting_initialization) {
+      check(cfg.init_phase == councl::init_phase::READY,
+            "reset requires a loading or completed election generation");
+      state_t sg(get_self());
+      check(!sg.exists() || sg.get().phase == councl::election_phase::DONE, "current election is not complete");
+   }
+
+   cfg.init_phase = councl::init_phase::CLEANING;
+   cfg.cleanup_stage = councl::cleanup_stage::CANDIDATES;
```
heifner: **[P2]** Aborting a failed snapshot load routes through the same purge pipeline starting at cleanup_stage::CANDIDATES, so reset() from LOADING destroys every candidate's self-paid registration row even though the candidate registry is unrelated to the tier-load failure being aborted.

If `finalizeinit` fails for any reason (roa churn, or the roa/nodecount divergence flagged there), governance's only recovery is reset() + purge(), which drains up to MAX_CANDIDATES = 1000 candidate-billed rows, zeroes cand_count, and bumps election_gen — every candidate must re-discover the abort and re-register in the new generation (the loading_generation_can_be_aborted test shows exactly this re-registration). The abort path could instead reset only the tier-load state (t2_loaded/t3_loaded/tier2/tier3/roster) and preserve the REG-phase registry.

contracts/tests/sysio.councl_tests.cpp
```diff
@@ -0,0 +1,1230 @@
...
+
+   void load_abi(name account, abi_serializer& out_ser) {
+      const auto* accnt = control->find_account_metadata(account);
+      BOOST_REQUIRE(accnt != nullptr);
+      abi_def parsed;
+      BOOST_REQUIRE_EQUAL(abi_serializer::to_abi(accnt->abi, parsed), true);
+      out_ser.set_abi(std::move(parsed), abi_serializer::create_yield_function(abi_serializer_max_time));
+   }
+
+   void setup_emission_config() {
```
heifner, line 143: **[P3]** setup_emission_config() re-types the full 19-field setemitcfg blob verbatim from sysio.dispute_tests.cpp, and the push()/load_abi() fixture pair is a character-for-character copy of the same helpers there — none live in the shared sysio.system_tester.hpp.

Adding or renaming one field in sysio.system's emission_config now silently breaks three separate fixtures (dispute, epoch_flushwtdw, councl) that each hard-code the whole struct; whoever adds the field fixes two and the third aborts at construction with an ABI-pack error naming no field, taking down all ~36 council cases. Likewise a change to the tester's transaction-header/signing API must be patched in each private push() copy. Extracting these into the shared tester removes the multi-copy maintenance trap.

heifner (commented):
Review of the new sysio.councl contract at head 6c2ecaa. 15 inline comments, tagged **[P1]**/**[P2]**/**[P3]**.

The P1s share one root cause: the contract treats `sysio.system::nodecount` as the authority on `roa::nodeowners` membership, but the two diverge — `nodecount` is a cumulative, unscoped counter that only increments once emitcfg exists, while the election snapshots per-network_gen roa rows that are populated unconditionally. Under bootstrap orderings where any owner registers before setemitcfg (the state this PR's own fixture is in), the mismatch permanently bricks `finalizeinit`, and one extra tier-1 registration makes `startinit` unsatisfiable — with no repair action in either contract. A shared canonical accessor for the roa/nodecount reads (currently also hand-rolled in sysio.chalg) would collapse most of that divergence surface.

The second cluster is recovery-path robustness: `purge()`'s stage switch has no default arm (an out-of-range stage byte is an infinite loop with no exit from CLEANING), a running election cannot be aborted while `time_slot_sec` is immutable after `startinit`, and the `state` singleton survives generation rollover advertising the previous election as current.

On the disclosed proposer-selection grinding: it is sharper than the sign-off note suggests — a single account can grind the draw offline within a single transaction, because the caller's name feeds the entropy accumulator immediately before the same call's selection draw.

Verified correct while reviewing: the core election math, Fisher-Yates tier-3 selection, gr_scope packing, serialization determinism, deadline arithmetic, and RAM billing.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229` (retained through `62bea66c16`). `finalizeinit` now requires completed bounded source scans and compares the snapshots with generation-scoped `sysio.roa::nodeowners` counts through the shared ROA accessor; it no longer reads the lagging cumulative `sysio.system::nodecount`. Added a regression covering nonzero tier-2/tier-3 owners registered without emissions configuration.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. Cleanup mode is validated before purge processing, and the cleanup-stage switch now has `default: check(false, "unknown cleanup stage")`, so an invalid stored stage fails diagnostically instead of looping without progress.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. `regnodeowner` now enforces each tier cap directly against the authoritative, generation-scoped ROA rows before allocation/insertion, independent of emissions configuration and the lagging nodecount singleton. The regression confirms a 22nd tier-1 owner is rejected even when nodecount has not caught up.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. `reset()` now permits an authorized abort from any READY election; the cleanup path removes snapshots, remaps, and partial council rows before advancing the generation. A regression starts a partially completed election, aborts it, and verifies the partial result and global state are removed.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. Council finalization now obtains generation-scoped counts from `sysio.roa::nodeowners` via the shared accessor and no longer requires the optional nodecount singleton. Added a regression that finalizes a 21-owner tier-1 election with no emissions configuration and zero tier-2/tier-3 owners.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. Purge completion now removes the unscoped `state` singleton before returning to REG, and completion keeps `active_seat` at the valid final index (20) instead of advancing it to 21. The lifecycle tests assert both behaviors.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. `loadtier` now stores a primary-source cursor and scan-complete flag per tier, and each call scans at most `max_rows` ROA rows rather than bounding only inserts. Deduplication remains by identity, and a later pass can absorb rows inserted earlier in sort order; tests cover bounded cursor progress and churn rescan.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed as an explicit risk disclosure in `58ffcde229`, not as a cryptographic mitigation. DESIGN/README now state that an actor controlling multiple eligible identities can simulate deterministic post-action selection offline and submit only the favorable identity in the same transition-triggering action; eliminating that capability requires an external unpredictability source such as VRF.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. `repcandidate` and `vote` now accept an optional trailing `expected_round`; when supplied it is checked after settlement, so a deadline-crossing transition makes the transaction fail and roll back instead of returning success with discarded input. Omitting it preserves the existing settlement-only compatibility path, and tests cover the fail-loud rollback behavior.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. The purge helper now calls `table.erase(std::move(it))`, avoiding the iterator copy/reseek on every erased row.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. Added compile-time checks that the tier-1 and tier-2/tier-3 rules remain equivalent for both success and elimination thresholds: `win_threshold(T1_VOTERS) + 1 == win_threshold(SEATS)` and `elim_threshold(T1_VOTERS) == elim_threshold(SEATS)`.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. The nomination and vote paths again enforce the explicit inclusive deadline after settlement, and the optional `expected_round` path additionally binds fail-loud callers to the intended round. Tests exercise the deadline boundary and rollback semantics.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229` and integrated with the current base in `62bea66c16`. `sysio.roa.hpp` now provides the canonical cross-contract `current_network_gen` and generation-scoped `nodeowner_count` accessors; council uses both. The merged dispute implementation also uses the shared network-generation accessor when freezing its per-dispute electorate snapshot, removing the hand-rolled generation read.

contracts/sysio.councl/src/sysio.councl.cpp
huangminghuang: Addressed in `58ffcde229`. Cleanup now distinguishes a LOADING abort from completed/READY cleanup: the LOADING path begins at roster/snapshot data, preserves candidate registrations, and keeps the same election generation. A regression verifies candidates remain registered after aborting a failed load.

contracts/tests/sysio.councl_tests.cpp
huangminghuang: Addressed in `58ffcde229`. The shared contract test support now owns `load_account_abi`, `push_contract_action`, and `default_emission_config`, and the council and dispute fixtures were migrated to those helpers. This removes their duplicated ABI-loading, transaction-push, and 19-field emission-config setup code.

heifner (approved):
Approving. Re-reviewed at 62bea66c.

The nodecount-vs-roa divergence behind most of the earlier findings is fixed at the root rather than per-symptom: `roa::current_network_gen()` / `roa::nodeowner_count()` are now the single authoritative source, councl and chalg both consume them, and roa enforces its own tier caps against its own rows. That one change closes the finalizeinit block, the startinit over-population case, the missing-nodecount abort, and the duplicated readers together. The persistent loadtier cursor, the two abort modes, the `expected_round` opt-in, the restored deadline checks, and the threshold static_asserts all hold up against the resulting code and the new tests. The master merge leaves the councl and roa sources untouched, and the committed WASM/ABI were genuinely rebuilt from the fixed source.

Two non-blocking notes, neither of which applies to the launch configuration:

Proposer grinding is unreachable with an all-tier-1 population. `fail_attempt` only draws when `cfg.n2 > 0` or tier-3 slots remain, so at `n2 == n3 == 0` every failed attempt goes straight to BACKSTOP and `selection_index` is dead code. It reactivates on the first generation run after tier-2/tier-3 onboarding begins. One detail worth having on record for that point: `stir(caller)` is unrestricted and repeatable, and stirring inside a live window is a state-machine no-op, so the search does not actually require the several eligible identities DESIGN.md section 5 assumes — a single account can batch stirs before the deadline and settle after, with the whole search done offline. Restricting `stir` to election participants would at least make the documented cost model accurate once the lower tiers go live.

The tier-cap check that moved into `roa::regnodeowner` is an O(tier-population) scan per registration where it used to be an O(1) counter read. At 21 owners that is nothing; if tier-3 ever fills toward its 1000 cap, a per-(generation, tier) counter maintained in roa would restore O(1) without giving up the authoritative-count correctness the fix introduced.

Worth being aware of rather than acting on: with the lower tiers empty, every seat whose slate stalls routes to governance `forceassign`, and with 20 tier-1 voters any 7 of them can eliminate a candidate — so a blocking minority puts a seat in governance's hands rather than escalating through the tiers.

PR #519, opened 2026-07-15
Chain: Remove base_per_transaction_net_usage consensus parameter

heifner:
## Summary
Removes the `base_per_transaction_net_usage` chain config parameter. Since #111 replaced the per-transaction base NET charge with the computed per-action overhead (`billable_net_per_action_overhead`), the parameter has been dead weight: only read by `chain_config::validate()` and echoed through SHiP and contract ABIs. A silently-unused consensus parameter invites confusion -- governance could vote the value and nothing would happen.

## Consensus impact
- The field packs into `genesis_state`, so the chain id of every genesis changes: this is a regenesis-level change and a running network cannot adopt it in place (the reference `snap_v1` chain id is now `087244f65e31c0106a58554b8f855e30ae657efb98c6c40348bb14db8bdb3f8e`).
- The id-tagged `set_parameters_packed`/`get_parameters_packed` protocol renumbers: ids above 3 shift down by one, `PARAMS_COUNT` 20 -> 19. `sysio.system::setparams` and the `params_test`/`action_results` test contracts are renumbered to match.
- Packed `chain_config_v0` shrinks 80 -> 76 bytes (`system_host_tests` stability check updated).
- SHiP `chain_config_v0` loses the field in both the ABI and the `global_property` serializer; old SHiP streams and old snapshots are incompatible, which regenesis covers.

## Lockstep
- Requires Wire-Network/wire-cdt#101: the host unpacks the `set_blockchain_parameters_packed` blob directly into `chain_config_v0`, so CDT's `blockchain_parameters` must match field-for-field. `sysio.bios`, `sysio.system`, `params_test`, and `action_results` are rebuilt against that change (`get/set_blockchain_parameters` are compiled into CDT's `libsysio`, so the rebuild required a fully rebuilt CDT, not just the patched header).
- Companions: Wire-Network/wire-libraries-ts#45 (regenerated `SysioContractTypes.ts`) and Wire-Network/wire-tools-ts#28 (generated genesis + bootstrap doc).
- Hyperion needs no code change: it decodes SHiP deltas with the ABI streamed at the websocket handshake. Deployment genesis files (hyperion-docker-new, docker-wire-api-node, cloud-infra) are an ops item at regenesis.

## Regenerated reference data
Deep-mind log, snapshot compatibility files (`blocks.*`, `snap_v1.*`), `consensus_blockchain`, and the `sys-util snapshot info` expectations.


libraries/chain/chain_config.cpp
```diff
@@ -19,12 +19,6 @@ namespace sysio { namespace chain {
       SYS_ASSERT( max_transaction_cpu_usage < max_block_cpu_usage, action_validate_exception,
                   "max transaction cpu usage must be less than max block cpu usage" );
 
-      SYS_ASSERT( base_per_transaction_net_usage < max_transaction_net_usage, action_validate_exception,
-                  "base net usage per transaction must be less than the max transaction net usage" );
-      SYS_ASSERT( (max_transaction_net_usage - base_per_transaction_net_usage) >= config::min_net_usage_delta_between_base_and_max_for_trx,
-                  action_validate_exception,
-                  "max transaction net usage must be at least {} bytes larger than base net usage per transaction",
-                  config::min_net_usage_delta_between_base_and_max_for_trx );
       SYS_ASSERT( context_free_discount_net_usage_den > 0, action_validate_exception,
```
huangminghuang, line 26: [P1] Preserve a minimum transaction NET limit

Removing the base field also removes the only lower-bound validation for `max_transaction_net_usage`. A value of zero now passes validation whenever `max_block_net_usage` is positive. `transaction_context` then caps every non-read-only transaction at this value and rejects every billable action—including a corrective `setparams` transaction—leaving the chain recoverable only by hard fork. Please preserve a base-independent minimum transaction NET limit and add rejection tests for zero and undersized values.

heifner: Fixed in 26039824b0: validate() now enforces a base-independent floor, config::min_max_transaction_net_usage = 10 KiB (matching the old effective bound from the delta check), with the recovery rationale documented on the constant. system_host_tests covers zero and floor-1 (both rejected) and the floor value (accepted).

huangminghuang (approved):

PR #522, opened 2026-07-16
Chain: gate operator-daemon startup on controller::is_synced()

heifner:
Replaces #520 — same goal, reduced mechanism.

## Problem

Every test cluster's boot window logged error-class lines from batch_operator (Account/Contract Query Exception 3060002/3060003): plugin_startup eagerly read `sysio.chains` before a cold-booting operator node's local replay reached the deploy blocks. The underwriter already solved this privately with an `accepted_block`-signal gate plus an inner executor re-post.

## Change

- **`controller::is_synced(now, recency_window)`** plus a wall-clock `is_synced()`: the LIB-recency predicate (fork_db root block time within `controller::default_sync_recency_ms` of now), computed on demand, never cached. Lives beside the fork_db accessors it reads and is thread-safe like them. LIB rather than head deliberately: operator daemons run read-mode = irreversible, so the irreversible state is what their table reads serve.
- **No new signal.** The predicate is a level, and the wake-up a gated consumer needs already exists as the `channels::irreversible_block` appbase channel — a LIB advance is the only event that can turn the predicate true, and channel deliveries are posted to the application executor, so callbacks run on the main thread after the triggering block fully commits (the post-commit contract the underwriter previously re-implemented with its inner post). Steady-state cost is zero: the daemons are subscribed only while gated, and `channel::publish` skips posting with no subscribers.
- **underwriter_plugin**: the gate subscription moves from the raw controller `accepted_block` signal to the channel; the private predicate, `sync_recency_ms`, and the inner re-post are deleted. Single gated path — there is deliberately no already-synced fast path: operator daemons boot with genesis-stale LIB in every topology, and a node that somehow is synced at startup is released by the next LIB advance.
- **batch_operator_plugin**: the startup body (outpost discovery, cron_service sizing/creation, epoch_tick, relay jobs) moves verbatim into `impl::run_deferred_startup` behind the same gate — the boot-window read errors disappear, and cron sizing sees the real outpost count instead of always hitting the `MIN_CRON_THREADS` floor.
- **Uniform fail-fast**: a terminal deferred-startup failure logs and quits the node in both daemons — an operator daemon that is not performing its duties is supervisor-visible instead of hidden behind a running process. The underwriter's bounded preflight retry grace still absorbs transients; the escape_policy split (abort vs contain, previously selected implicitly by restart timing) is deleted, with `underwriter_detail::is_terminal_failure` classifying the terminal states in one place.
- **read-mode = irreversible is required** by both daemons (FC_ASSERT at plugin_initialize when the plugin is enabled). Together with producer_plugin's inverse assert (no producer-name under irreversible read-mode), producer/daemon co-hosting is now impossible by configuration.

## Behavior deltas vs #520

- No already-synced fast path; release comes at the next LIB advance (in practice the daemons are never synced at plugin_startup).
- Terminal deferred-startup failure quits the node. Previously a terminally failed underwriter stayed up serving `preflight_failed`/`wiring_failed` gate payloads over `/v1/underwriter/*` indefinitely; that surface now exists during the retry grace and the shutdown window only. Monitoring that polls a failed-but-alive underwriter will see process exit instead.
- No reference-data changes ride along (this branch is based past #519, so the regenerated-data conflicts on #520 do not carry over).

## Considered and deliberately not replaced

Other in-sync checks in the tree measure different things and stay as they are: producer_plugin's max-irreversible-block-age production gate (needs the raw age for its log/API surface and deliberately treats a snapshot-started producer as synced so it can revive a chain), `block_state::is_recent`'s 30s head-recency vote gate (gating votes on LIB recency would deadlock finality recovery — votes are what create finality), producer_plugin's stale-production wake (head-based so producers can revive a stale chain), net_plugin's peer-relative sync_manager, and the vote-recency production pause.

## Tests

`unittests/controller_sync_tests.cpp`: the boundary cases drive `is_synced(now, window)` with instants offset from the tester chain's actual last-irreversible block time (window edge, past-window, zero window, clock-skew-ahead, replay gap), so every comparison exercises the real controller path deterministically; wall-clock cases cover the convenience overload. The underwriter's migrated predicate cases were removed with its private predicate; its gate-payload cases remain, plus a pinning test for `is_terminal_failure`.


plugins/batch_operator_plugin/src/batch_operator_plugin.cpp
```diff
@@ -882,50 +988,45 @@ void batch_operator_plugin::plugin_startup() {
...
+   // released by the next LIB advance.
+   auto& chain = _impl->chain_plug->chain();
+   ilog("batch_operator_plugin: waiting for chain sync before outpost discovery "
+        "(head {} is {}s behind now; irreversible state is {}s behind)",
+        chain.head().block_num(),
+        (fc::time_point::now() - chain.head().block_time()).to_seconds(),
+        chain.fork_db_has_root()
+           ? std::to_string((fc::time_point::now() - chain.fork_db_root().block_time()).to_seconds())
+           : "n/a");
+   _impl->sync_gate_subscription =
```
huangminghuang, line 1019: [P2] Re-check the sync level after subscribing

chain_plugin::plugin_startup() loads and replays persisted chain state before this dependent plugin reaches plugin_startup(), so a normal restart can arrive here with a recent LIB after all startup replay events have already been published. Because is_synced() is evaluated only inside a future channel callback, the relay never starts if no later LIB advances (for example, a restart during a network stall), even though the level is already true. Please subscribe first and then immediately invoke the same guarded maybe_start check, and mirror it in underwriter_plugin, so both the already-synced and lost-wakeup cases are covered. A restart test with no subsequent LIB would pin this behavior.

heifner: The mechanism is real (the condition is a level, the gate observes edges), but here the wake-up and the work supply are the same event: everything this plugin does is driven by newly-finalized WIRE state, so when no LIB advance follows startup there is also nothing to relay — the first finalized block re-arms the gate and creates the first actionable state at the same instant. The scenario also self-extinguishes: it needs the restart to land inside the recency window (about 3s effective) at the onset of a finality stall, and within 5s the stalled LIB ages out of the window, so is_synced() is false for the remainder of the stall under either shape — an inline check only wins that fleeting race, and the daemon it starts has nothing to act on and cannot finalize its own pushes. The one asymmetry is pre-stall finalized outbound backlog owed to the external chains, which a running daemon could push during the stall; that is bounded by the stall duration behind a ~3-second restart race, and we prefer recovery-time delivery there over carrying an inline start path that cannot execute in any deployable topology (read-mode = irreversible is required and producer co-hosting is rejected) plus a restart test that must pause finality to pin it. Codified the wake-up == work-supply rationale in a comment at the subscribe sites (f17feb8272).

huangminghuang (approved):
Re-reviewed at f17feb8272. The sync-gate restart race is a narrow, self-recovering operational edge case; the documented tradeoff is acceptable and non-blocking. No blocking findings.

PR #524, opened 2026-07-16
Opp: stamp outbound envelope destination endpoints (SEC-108)

heifner:
Outbound OPP envelopes carried no destination identity: `Envelope.endpoints` was left at proto3 default by every emitter, so the only thing tying an envelope to an outpost was each batch operator's local `--batch-outpost` routing config. A fleet misconfigured the same way on every operator could deliver chain A's envelope to chain B's outpost, and since envelopes to two outposts with identical stream history are byte-identical, the receiving contract's chain checks cannot reject such a misdelivery — it would be accepted, advancing B's inbound tip to A's digest and wedging every subsequent legitimate delivery.

`buildenv` now stamps the envelope's route endpoints from the destination's `sysio.chains` row (`start = WIRE`, `end = {kind, external_chain_id}`) — the same value it already wrote to the `envlog` audit table. Endpoints are part of the canonical field-complete encoding, so the destination binding rides inside the epoch-digest preimage and is covered by delivery consensus. Receiving outposts verify it against their own chain identity and reject mismatches: wire-ethereum compares `block.chainid` (config-free), wire-solana checks the SVM kind — companion PRs on those repos.

`regchain` now rejects a second EVM row with an already-registered `external_chain_id`, keeping the destination binding injective across EVM rows. SVM rows are exempt — Solana clusters have no numeric chain id, so the Solana check is kind-only; distinguishing two SVM outposts would need the registry slug carried in the envelope, a documented follow-up if a second SVM chain onboards.

Depot-shape golden vectors A and B are re-pinned for the populated endpoints (the semantic-header derivation is untouched, so all header pins are unchanged; vector C already covered populated-endpoints encoding cross-language). `msgch.wasm` and `chains.wasm` recompiled; both ABIs unchanged.

SEC-108 PR set: Wire-Network/wire-sysio#524 (depot stamps destination endpoints), Wire-Network/wire-ethereum#152 (EVM outpost check), Wire-Network/wire-solana#381 (SVM outpost check).


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1821,6 +1841,7 @@ void msgch::buildenv(uint64_t chain_code) {
       msg.payload = std::move(payload);
 
       opp::Envelope env;
+      env.endpoints = route_endpoints;
```
huangminghuang, line 1844: [P1] Provide a safe hash-chain migration

Stamping endpoints changes the digest preimage, but buildenv chains from and replaces the latest emitted row without delivery acknowledgement. If either companion outpost enables its strict destination check while a legacy zero-endpoint envelope is pending, that envelope is rejected; subsequent stamped envelopes then chain from an unaccepted digest and are rejected permanently. Please add an activation/migration mechanism, or enforce and test a rollout that first lets every legacy envelope reach the outpost and only then enables destination validation.

contracts/sysio.chains/src/sysio.chains.cpp
```diff
@@ -54,6 +54,20 @@ void chains::regchain(opp::types::ChainKind kind,
                    "sysio.chains: a WIRE chain (depot self-row) already exists");
    }
 
+   // Enforce: EVM rows are unique per `external_chain_id`. The pair (kind, external_chain_id)
+   // is stamped into every outbound envelope's route endpoints (`sysio.msgch::buildenv`) and is
+   // what an EVM outpost verifies against its own `block.chainid`, so it must stay injective
+   // across EVM rows. SVM rows are exempt: Solana clusters have no numeric chain id (rows carry
```
huangminghuang: [P2] Fail closed on multiple SVM rows

This permits multiple SVM destinations and says code disambiguates them, but code is not carried in Envelope.endpoints and the companion Solana gate checks only the SVM kind. A second SVM row is therefore indistinguishable and recreates the misdelivery risk. Until the slug is carried and validated, please reject registration of a second SVM row.

heifner: Fixed in c8840feaf0: regchain now rejects a second SVM row outright (mirroring the single-WIRE-row check), and the misleading "code disambiguates them" comment is gone. The registry-slug-in-envelope follow-up remains the lift for onboarding a second SVM chain; until then registration fails closed. Test updated: an SVM row may still share a numeric id with EVM rows, but a second SVM registration is rejected.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: The wedge requires an in-flight zero-endpoint envelope straddling the upgrade — a pending emit built by the old depot but not yet accepted by a new, strict outpost. No supported deployment passes through that state: every deployment of this stack is genesis-fresh (production launch is chain genesis, and the platform e2e boots clean clusters with all components at head), so the first envelope any outpost ever sees is already endpoint-stamped and the destination checks are live from the first delivery. For a hypothetical non-genesis upgrade the safe ordering is depot-first: current outpost code ignores endpoints entirely, so stamping is transparently compatible, and outposts upgrade after their pending stream drains. If we ever want upgrade-order independence, the mechanism would be a one-bool generation latch on each outpost (accept the legacy zero form until the first stamped envelope is accepted, then strict forever) — deliberately not included here to keep the outpost checks config-free and stateless per the agreed scope.

huangminghuang (approved):
Approved at c8840feaf0. Before merge, please update the PR description so it reflects the complete current PR: regchain now fails closed by rejecting a second SVM row, rather than exempting SVM rows. Please also capture the launch/genesis-only deployment assumption discussed in the review thread and note the required depot-first, pending-stream-drained ordering for any hypothetical non-genesis rollout.

PR #527, opened 2026-07-17
Sigprov: split SSM/KMS signature schemes into standalone plugins

heifner:
Reorganizes `signature_provider_manager_plugin` so the built-in `KEY:`/`KIOD:` schemes stay in the manager while the AWS-backed `SSM:` and `KMS:` schemes become standalone, opt-in plugins (`signature_provider_ssm_plugin`, `signature_provider_kms_plugin`), with the shared AWS SDK glue in `libraries/signature_provider_aws`. The manager archive stays AWS-free; the provider plugins link into nodeop only, so sys-util, cranker-example, and the outpost tools stay AWS-free. (Supersedes #525, which flattened the plugin's directory layout — that change is included here.)

## Enablement

An operator turns a scheme on with `plugin = sysio::signature_provider_ssm_plugin` / `..._kms_plugin`. Provider plugins register their scheme handler from their constructor; appbase constructs every plugin before it initializes any, so the manager — which every provider consumer `APPBASE_PLUGIN_REQUIRES` — creates every configured provider at its own `plugin_initialize`, before any consumer, gated on the plugin being enabled. This is order-independent: it does not matter where the provider plugin's `plugin =` line sits relative to other plugins. A spec whose plugin is linked but not enabled fails the boot naming the exact `plugin =` line; an unknown scheme reports that no plugin provides it.

## Example config.ini

```ini
# KEY: is built in -- no plugin line needed. <name>,<chain-kind>,<key-type>,<public-key>,<provider-spec>
signature-provider = wire-01,wire,wire,SYS7AzqPxqfoEigXBefEo6efsCZszLzwv4vCdWqTt6s6zSnDELSmm,KEY:5J5LzjfChtY3LGhkxaRoAaSjKHtgNZqKyaJaw5boxuY9LNv4e1U

# SSM: fetches the key once from AWS SSM Parameter Store (a KMS-encrypted SecureString) at startup,
# then signs locally -- fast enough for every signing path including producer/finalizer keys.
plugin = sysio::signature_provider_ssm_plugin
signature-provider = bp1,wire,wire,SYS7AzqPxqfoEigXBefEo6efsCZszLzwv4vCdWqTt6s6zSnDELSmm,SSM:us-east-1:/wire/prod/bp1

# KMS: remote signing via AWS KMS (secp256k1/ethereum keys only; the key never leaves KMS).
plugin = sysio::signature_provider_kms_plugin
signature-provider = eth-01,ethereum,ethereum,0xfc5422471c9e31a6cd6632a2858eeaab39f9a7eec5f48eedecf53b8398521af1c86c9fce17312900cbb11e2e2ec1fb706598065f855c2f8f2067e1fbc1ba54c8,KMS:us-east-1:alias/wire-eth-01

# Region-less variants work anywhere the environment supplies a region (env vars, ~/.aws/config,
# or IMDS on AWS compute):  ...,SSM:/wire/prod/bp1   and   ...,KMS:alias/wire-eth-01
```

`<param-ref>` / `<key-ref>` accept the shorthand `<region>:<name-or-id-or-alias>` (as above), a region-less `<name-or-id-or-alias>` (region resolved from the environment — next section), or a full ARN. Omitting a plugin line for an `SSM:`/`KMS:` spec fails the boot with an error naming the exact line to add.

## Region resolution

A spec names its region explicitly (`<region>:` prefix or embedded in an ARN) or omits it. When omitted, the region resolves at provider creation through the same chain the AWS SDK itself uses: `AWS_DEFAULT_REGION`, then `AWS_REGION`, then the shared-config profile's `region`, then the EC2 instance-metadata service — any process on AWS compute can learn its own instance's region from IMDS, so on EC2/ECS a region-less spec works with zero extra configuration. Two deliberate differences from stock SDK behavior: resolution never silently falls back to `us-east-1` (an unresolvable region fails the boot with a precise error naming the fix), and an explicit spec region always wins over the environment. IMDS is skipped when `AWS_EC2_METADATA_DISABLED=true`.

For `SSM:` the two shorthand forms are told apart by the leading token: one shaped like an AWS region means the explicit-region form; anything else makes the whole body the parameter name, `:version`/`:label` selector colons included. The one ambiguous corner — a region-less *selector* reference to a parameter whose own name is shaped like a region (`SSM:my-param-2:label`) — parses as the explicit-region form; such a parameter stays addressable via an explicit region or an ARN. For `KMS:` there is no ambiguity: key ids and aliases can never contain `:`, so no colon = region-less. A bare region with nothing after it (`SSM:us-east-1`) is rejected at parse time as the likely "forgot the name/key" typo.

The resolution chain, region-shape classifier, and region-less client-cache normalization live in `libraries/signature_provider_aws` and are covered by a new offline unit binary (`test_signature_provider_aws`) with RAII env fixtures shared by the kms/ssm suites.

## Behavior changes

- `SSM:` no longer works without its `plugin =` line (deliberate; the boot error names it).
- `KMS:` is reachable in nodeop for the first time behind its plugin — safe because the provider hard-asserts ethereum key types, so it can never back a producer/finalizer key.
- The `signature-provider-kms-startup-check` option is removed; KMS keys are probed at startup unconditionally (a transient AWS error defers to the first sign, so a blip never blocks a boot). A config still setting the option fails to parse — delete the line.
- The manager's provider set is immutable after startup, enforced at each mutator.

## Outpost clients

The outpost client plugins resolve their signature providers and construct clients at `plugin_initialize`: the manager — which they `APPBASE_PLUGIN_REQUIRES` — creates every configured provider at its own initialize, before any consumer initializes, so `--plugin` ordering is a non-issue and clients are readable immediately after init. The standalone RPC tools and cranker-example therefore need no explicit `startup()` call; the tools' previously-added startup calls are removed.


huangminghuang (approved):

libraries/signature_provider_aws/include/sysio/signature_provider_aws/aws_common.hpp
```diff
@@ -78,24 +111,22 @@ class region_client_cache {
...
+    *               (see `resolve_default_region()`, which throws when nothing resolves)
+    * @return shared client configured for the effective region
     */
    std::shared_ptr<Client> get(const std::string& region) {
-      SYS_ASSERT(!region.empty(), chain::plugin_config_exception,
-                 "AWS client cache lookup requires a non-empty region");
-      std::scoped_lock lock{_mutex};
-      auto& slot = _by_region[region];
+      const std::string effective = region.empty() ? resolve_default_region() : region;
+      auto& slot = _by_region[effective];
```
huangminghuang, line 125: [P2] Restore synchronization around the shared client cache

This cache previously serialized lookup, and `get_kms_client` remains documented as thread-safe. Concurrent first calls to the exposed KMS/SSM helpers now mutate `_by_region` through `operator[]` without synchronization, causing a `std::map` data race and undefined behavior. Current manager initialization being single-threaded does not preserve the public helper contract. Restore the mutex around lookup/insertion, or explicitly restrict and document these APIs as initialization-only.

plugins/outpost_solana_client_plugin/README.md
```diff
@@ -389,14 +389,28 @@ activate Plugin
...
+   Plugin -> Plugin : parse + validate spec\nqueue pending client
+end
+
+Plugin --> Main : initialized
+deactivate Plugin
+
+Main -> Plugin : plugin_startup()
+activate Plugin
+note right of Plugin
+  provider resolution is deferred to
```
huangminghuang: [P3] Update the obsolete startup lifecycle

This describes an intermediate implementation that no longer exists: both outpost plugins now resolve providers and construct clients during `plugin_initialize`. The RPC tools still call `startup()` based on the obsolete assumption, and the PR description repeats it, while cranker reads clients immediately after initialization. Update the diagram and PR description and remove the unnecessary tool startup calls, unless startup-time construction is actually intended.

libraries/signature_provider_aws/include/sysio/signature_provider_aws/aws_common.hpp
heifner: Restored in 6b07b92e6d: `region_client_cache::get` now serializes lookup, insertion, and client construction behind an internal mutex, so the `get_kms_client` / `get_ssm_client` threadsafe contract holds again. The empty-region resolution deliberately runs before the lock — it reads no cache state, and its IMDS step can perform bounded network I/O that shouldn't hold the lock.

plugins/outpost_solana_client_plugin/README.md
heifner: Updated in 3dac68725a: the sequence diagram now shows provider resolution + client construction inside `plugin_initialize` (with a note on the APPBASE_PLUGIN_REQUIRES ordering that makes it safe), both RPC tools drop the now-unnecessary `startup()` calls (`plugin_startup` is a bare ilog in both plugins), the README example comment matches cranker-example's, and the PR description's Outpost clients section is corrected. Initialize-time construction is the intended design.

plugins/signature_provider_kms_plugin/test/README.md
```diff
@@ -128,7 +128,7 @@ export KMS_LIVE_SPEC='us-east-1:alias/wire-ci-test-secp256k1'
 # export KMS_LIVE_SPEC='arn:aws:kms:us-east-1:111122223333:alias/wire-ci-test-secp256k1'
 export KMS_LIVE_PUBKEY='045a87...eef3'
 
-$BUILD_DIR/plugins/signature_provider_manager_plugin/kms/test/test_sigprov_kms \
+$BUILD_DIR/plugins/signature_provider_kms_plugin/test/test_signature_provider_kms_plugin \
    --run_test=kms_signature_provider_tests/kms_live_sign_round_trip \
```
huangminghuang, line 135: [P3] Remove the deleted startup-probe flag reference

While updating this runbook, please also update its **Optional startup probe** section (around lines 59-62) and the parser comment near line 632 of `kms_signature_provider.cpp`. There is no longer a `signature-provider-kms-startup-check` flag: attaching `startup_probe` is the opt-in, and the manager always invokes each attached probe at startup. The current wording incorrectly implies that another opt-in flag gates it.

heifner: Fixed in 6f4388b849: the README's IAM section now says the probe is what the manager always runs at boot (no enable flag), with the lazy first-sign pin as the transient-deferral path; the `create_kms_provider` comment near the return, the `warm_up` block, the pinning-guard doc, and the ssm header's cross-reference all drop the opt-in-flag phrasing — attaching `startup_probe` is the opt-in and this handler always attaches one.

plugins/signature_provider_kms_plugin/test/README.md
```diff
@@ -56,12 +56,13 @@ aws kms create-alias \
 
 The principal whose credentials the test runs under needs both `kms:Sign`
 and `kms:GetPublicKey` on the key. `kms:GetPublicKey` is required at runtime,
-not just for setup: on the first sign of each key — and again from the
-optional startup probe — the plugin calls `GetPublicKey` exactly once to pin
-the operator-supplied public key against the one KMS actually holds. Without
-that grant the lazy pin and the startup probe both fail with
-`AccessDeniedException` on first use. The same call is also how you extract
-the matching pubkey hex for the spec (Prereq #4). Minimum inline policy:
+not just for setup: the plugin calls `GetPublicKey` exactly once per key to
```
huangminghuang: [P3] Don’t promise exactly one KMS call

`GetPublicKey` is not guaranteed to run exactly once per key. `ensure_kms_pubkey_pinned` leaves `pinned` false after any exception and explicitly retries on the next probe/sign, so a transient startup failure can produce multiple requests—the retry path described immediately below. The source comment near `ensure_kms_pubkey_pinned` similarly says “at most once per state” while documenting retries. Describe this as running until the first successful pin and being cached thereafter.

heifner: Fixed in fb5af1fe86 — the README and every source comment making the same promise (`ensure_kms_pubkey_pinned`'s "at most once", `verify_kms_pubkey`'s "invoked at most once", the sign-closure's "only the first", the `kms_signer`/`make_kms_signature_provider` docs' "exactly once"/"one-shot") now describe it as running until the first successful pin, cached on the shared state thereafter, with failed attempts retried by the next probe/sign.

plugins/signature_provider_kms_plugin/include/sysio/signature_provider_kms_plugin/kms_signature_provider.hpp
```diff
@@ -241,29 +254,31 @@ std::shared_ptr<Aws::KMS::KMSClient> get_kms_client(const std::string& region);
...
- * check runs at most once regardless of which one triggers it.
+ * check runs until its first success -- cached from then on -- regardless of
+ * which member triggers it.
  */
 struct kms_signer {
    /// Signing closure, usable wherever `fc::crypto::sign_fn` is expected. Each
-   /// call issues one `KMS::Sign`; the first call also runs the public-key
-   /// pinning check, unless `warm_up` has already run it. A transient KMS
-   /// failure throws `sysio::chain::signing_transient_exception` (safe to retry
+   /// call issues one `KMS::Sign`; until the pinning check has succeeded once
```
huangminghuang: [P3] Qualify the remaining call-count claims

A `sign` invocation does not always issue one `KMS::Sign`: it first calls `ensure_kms_pubkey_pinned`, and any failed `GetPublicKey` or pin validation throws before the Sign request is built. Say that each invocation which passes pinning issues one Sign. The source comment near `kms_signature_provider.cpp:615` has the same missing success qualification: after a transient startup-probe failure, the next sign performs another pin check, so only a successful probe avoids an additional check.

heifner: Done. The sign-closure doc now leads with the pinning check (it runs first until it has passed once, and a failed check throws before any `KMS::Sign` is built) and says a call that passes pinning issues one `Sign`. The `.cpp` startup-probe comment carries the same qualification: a successful probe spares the first `Sign` the check, while a transient probe failure leaves it pending so the first `Sign` runs it.

plugins/signature_provider_kms_plugin/include/sysio/signature_provider_kms_plugin/kms_signature_provider.hpp
```diff
@@ -280,14 +296,15 @@ struct kms_signer {
...
  * key's public key matches `expected_pubkey`. A mismatch
  * throws `chain::plugin_config_exception` immediately -- before any billable
  * `Sign` -- so a spec that pins the wrong `<public-key>` fails fast with a
- * direct error. The pinning check runs once on success; a transient
- * `GetPublicKey` failure is retried on the next `Sign`.
+ * direct error. The first successful check is cached for the provider's
+ * lifetime; a failed attempt (e.g. a transient `GetPublicKey` error) is
+ * retried on the next `Sign`.
  *
  * Each invocation sends an `ECDSA_SHA_256` `Sign` request with
```
huangminghuang: [P3] Qualify the duplicate Sign-request claim

This retains the unconditional wording corrected above. When `GetPublicKey` or pin validation fails, the closure throws before constructing a `KMS::Sign` request, so not every invocation sends one. Use “Each invocation that passes pinning…” or “After pinning passes…” here as well.

heifner: Fixed in 4bcb3f02bd: the paragraph now reads "Each invocation that passes pinning sends an `ECDSA_SHA_256` `Sign` request...", and the `@param expected_pubkey` doc below it -- which made the same unconditional claim -- now says the key is checked against `GetPublicKey` until pinning first passes, then used on each signature KMS produces to recover `v` and assert the match.

huangminghuang (approved):
Reviewed latest head 4bcb3f02bd8c004772e1331e666cc4851c686987. No remaining actionable findings.

PR #529, opened 2026-07-20
Opreg: bound terminate_window_ms to span the consecutive-miss run

heifner:
`sysio.opreg::setconfig` accepted any positive `terminate_window_ms`. A window narrower than the full consecutive-miss run (`terminate_max_consecutive_misses` per-epoch delivery records) makes the consecutive termination rail structurally vacuous — records age out of the rolling window before `termcheck` can observe the run — leaving only the percent rail, which a short window turns into a hair trigger. This was the residual gap flagged in the SEC-28/WSA-044 review of #493.

`setconfig` now requires `terminate_window_ms >= (terminate_max_consecutive_misses + 1) * epoch_duration_ms`, reading the live duration from `sysio.epoch::epochcfg`. Bootstrap installs opreg config before sysio.epoch is configured, so an absent epochcfg skips the check; `sysio.epoch::setconfig` performs the mirror validation against the stored opreg config — the same both-directions shape as #490's emissions-fit check — so no ordering of the two setters can accept a vacuous pair and raising the epoch duration cannot silently vacate a previously valid window.

The shared bound lives in `opreg::min_terminate_window_ms`; both contracts' wasm is rebuilt (ABIs unchanged). New tests cover the exact span boundary in both directions and the bootstrap ordering path.


contracts/sysio.opreg/include/sysio.opreg/sysio.opreg.hpp
```diff
@@ -113,6 +113,23 @@ namespace sysio {
...
+      /// structurally vacuous and only the percent rail live (SEC-28 residual).
+      /// Enforced from both `opreg::setconfig` (against the stored epoch
+      /// duration) and `sysio.epoch::setconfig` (against the stored opreg
+      /// config) so no ordering of the two setters can accept a vacuous pair.
+      /// Inputs are pre-bounded by those setters (misses <= 5, duration <= 30
+      /// days), so the product cannot overflow uint64.
+      static constexpr uint64_t min_terminate_window_ms(uint32_t consecutive_misses,
+                                                        uint32_t epoch_duration_sec) {
+         constexpr uint64_t ms_per_sec = 1000;
+         return (uint64_t{consecutive_misses} + 1) * epoch_duration_sec * ms_per_sec;
```
huangminghuang: [P1] Account for operator group rotation

This bound assumes each operator receives one delivery record per epoch, but `epoch::advance` calls `recorddel` only for the expiring group. With the default three groups and 21 operators, an operator is evaluated every three epochs; exceeding a threshold of five therefore requires six miss records spanning 15 epochs, while this helper accepts a six-epoch window and prunes earlier records first. With surplus operators, the rescheduling gap may be even longer. Base the window on a guaranteed maximum duty interval—or change the scheduling/tracking model—and add a rotation-level test with one outpost.

heifner: Right — records accrue only on duty epochs. The bound now reads `batch_op_groups` from the same epochcfg and requires `terminate_window_ms >= (terminate_max_consecutive_misses + 1) * batch_op_groups * epoch_duration_ms`; the `sysio.epoch::setconfig` mirror passes its new duration and group count, so raising either re-validates the stored window.

Added the duty-cadence test: one outpost, records spaced a full rotation apart — ACTIVE through the fifth miss, TERMINATED at the sixth, with the whole run (including the anchor a full span earlier) still un-pruned and in-window. Plus a group-raise rejection case on the mirror.

On the surplus point: an operator benched by the tail-selection accrues no records at all while benched, so no finite window can observe it — the bound governs resident operators, whose duty interval the sliding schedule pins at `batch_op_groups` epochs. That distinction is now spelled out in the helper doc.

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -284,6 +284,26 @@ void epoch::setconfig(uint32_t epoch_duration_sec,
...
+   // the same check against the stored schedule, so no ordering of the two
+   // setters can accept a vacuous pair (SEC-28 residual).
+   {
+      opreg::opconfig_t op_cfg_tbl(OPREG_ACCOUNT);
+      if (op_cfg_tbl.exists()) {
+         const auto op_cfg = op_cfg_tbl.get();
+         check(op_cfg.terminate_window_ms >=
+                  opreg::min_terminate_window_ms(op_cfg.terminate_max_consecutive_misses,
+                                                 epoch_duration_sec,
+                                                 batch_op_groups),
```
huangminghuang, line 324: [P1] Validate against the materialized schedule

This validates against the new config value, but `setconfig` does not resize `epoch_state.batch_op_groups`; `advance` preserves the existing vector length by popping and pushing one group, while `schbatchgps` is bootstrap-only. Starting with a live four-group schedule, lowering config to three passes while the old wide opreg window remains; a subsequent opreg update can then lower the window to 18 epoch durations because it reads `epochcfg`, even though operators remain on a four-epoch cadence and six misses span 20 epochs. The oldest miss ages out before termination. Reject group-count changes after scheduling, atomically migrate the live schedule, or validate against its actual size, and add a group-decrease transition test.

heifner: Right — and the drift hits more than the window: `advance()` dates each new tail group at `current_epoch_index + batch_op_groups - 1`, so a config count that diverges from the live rotation length mis-schedules groups too. The contract assumes `cfg.batch_op_groups` == the materialized schedule length throughout.

Went with option 1: `epoch::setconfig` now rejects a `batch_op_groups` change once `state.batch_op_groups` is materialized (bootstrap, before `schbatchgps`, stays free). That pins the config to the live rotation, so the opreg mirror's `epochcfg` read can't go stale and needs no change; reshaping the rotation is a fresh `schbatchgps`, a bootstrap operation. Added a transition test — against a materialized three-group schedule, `setconfig` rejects both a decrease and an increase and accepts the unchanged count.

contracts/tests/sysio.opreg_tests.cpp
```diff
@@ -408,6 +440,166 @@ BOOST_FIXTURE_TEST_CASE(setconfig_rejects_zero_min_bond, sysio_opreg_tester) { t
...
+   // (threshold + 1) * kDutyRotationMs of wall clock — the arithmetic the
+   // SEC-28 span bound protects. Drive one outpost's records at exactly that
+   // cadence under a window sized by the bound (with one rotation of margin
+   // for block-time skew) and require the run to stay observable end to end.
+   BOOST_REQUIRE_EQUAL(success(), set_epoch_config(kWindowBoundEpochDurationSec));
+   produce_blocks();
+   activate_batch_operator("batchop.a"_n,
+                           kDefaultMaxConsecutiveMisses,
+                           kMaxAcceptedPctMisses24h,
+                           kMinWindowMsAtDefaults + kDutyRotationMs);
```
huangminghuang, line 514: [P2] Test the real rotation at the exact boundary

This is not a rotation-level test: it never calls `schbatchgps`/`advance` or registers an outpost, and instead fabricates `recorddel` timing directly. It also uses `kMinWindowMsAtDefaults + kDutyRotationMs`, so it cannot catch an off-by-one at the exact minimum accepted by `setconfig`. Exercise a materialized three-group schedule with one outpost through `advance()` using exactly `kMinWindowMsAtDefaults`.

heifner: Added `terminate_at_duty_rotation_via_advance` in `sysio.msgch_chain_tests.cpp`, whose fixture already stands up the full `advance()` stack (sysio.system + emissions gate, a registered outpost, a materialized schedule). It drives a real three single-operator-group rotation with one outpost end to end: a non-bootstrapped operator delivers once (the anchor that keeps the percent rail below its ceiling) then withholds, and `advance()`'s inline recorddel/termcheck terminate it on the sixth consecutive duty-miss at exactly the minimum window the span bound accepts for that schedule -- no fabricated `recorddel`, no margin.

Because the anchor sits exactly on the window edge at that miss, an off-by-one in the span (or the window scan) would age it out and terminate early on the percent rail, so requiring the consecutive reason at the sixth miss pins the exact boundary. The fixed-cadence epoch schedule (advance sets current_epoch_start = prior next_epoch_start, never `now`) plus recorddel's one-second timestamp granularity make the exact span deterministic with no block-time skew. Left the opreg-level case as the fast arithmetic unit check.

contracts/tests/sysio.msgch_chain_tests.cpp
```diff
@@ -1006,4 +1125,61 @@ BOOST_FIXTURE_TEST_CASE(late_confirmation_after_consensus_recorded, sysio_msgch_
...
+         // otherwise: withhold delivery, so advance() records this duty epoch as a miss
+      }
+      advance_to_next_epoch();
+
+      auto op = get_operator(BATCHOP);
+      BOOST_REQUIRE(!op.is_null());
+      const auto status = op["status"].as<opp::types::OperatorStatus>();
+      if (status == opp::types::OperatorStatus::OPERATOR_STATUS_TERMINATED) {
+         terminated = true;
+         BOOST_REQUIRE_EQUAL(kTerminatingDuty, batchop_duties);
```
huangminghuang, line 1189: [P2] Assert the edge record survives

At the sixth miss, pruning the delivered anchor exactly on the window edge would not fail these assertions: the six remaining miss rows still trigger `exceeds_consecutive`, and `termcheck` prefers the consecutive reason when both rails fire. I confirmed this by changing the prune comparison from `<` to `<=`; the test still passed. Assert that the anchor `dellog` row remains after the terminating advance, or structure the rates so edge inclusion changes the outcome before the consecutive rail fires.

heifner: Good catch -- termination + reason don't distinguish the edge case (the six miss rows fire the consecutive rail regardless, and termcheck prefers that reason). Replaced that with a direct check: after the terminating advance, exactly one delivered `dellog` row must remain for the operator -- the anchor edge record, which `recorddel` erases if it ages out. Confirmed sensitive the same way you did: flipping `prune_dellog`'s `<` to `<=` (pruning the edge record) now fails the new assertion, while the old termination/reason assertions still passed.

huangminghuang (approved):
Reviewed the current head, including the rotation-cadence validation, materialized-schedule guard, exact-boundary integration coverage, and mutation-sensitive edge-record assertion. Local focused suites pass and the rebuilt epoch/opreg WASMs match the committed artifacts.

PR #530, opened 2026-07-20
SEC-130: Harden attended snapshot bootstrap downloads

huangminghuang:
## Problem

Remote snapshot bootstrap previously buffered an unbounded response in memory and lacked explicit response-size, disk-capacity, and partial-file guarantees. The first hardening pass added network deadlines, but request-write coverage was incomplete and fixed time limits do not fit the trusted, attended bootstrap model.

## Solution

- stream snapshot bodies through fixed 64 KiB parser/read buffers with one ceiling for fixed-length, chunked, and lengthless responses
- centralize decoded body increments so the bounded error diagnostic and main download loop share buffer setup, parser-read, `need_buffer` normalization, and byte accounting
- use the existing `chain-state-db-size-mb` value as the response-size ceiling; add no snapshot-specific timeout, size, or disk-reserve options
- keep the always-on 64 MiB disk-safety margin internal instead of exposing an unused per-download reserve field; deterministic disk tests inject only the free-space query through private test access
- report connection/request phases and five-second transfer updates with bytes, percentage, rate, and ETA when `Content-Length` is available
- keep network phases operator-controlled and make Ctrl+C cancel pending resolve, connect, request-write, response-header, metadata, and body-read operations
- preflight and periodically recheck disk capacity with a 64 MiB concurrent-consumer margin, then atomically rename successful downloads and remove `.downloading` files on failure or cancellation
- preserve up to 200 decoded bytes from non-200 response bodies in operator diagnostics without unbounded buffering
- reuse the metadata connection with one opt-in stale-connection retry for the idempotent download request
- require explicit download resource options and remove the unused legacy unbounded `post_to_file` overload

## Validation

- 20 focused HTTP cases covering cancellation in write/header/body/metadata phases, progressing transfers without a total deadline, fixed/chunked response ceilings, bounded non-200 diagnostics, truncated fixed-length cleanup, progress totals, stale keep-alive reuse/retry, and disk budgets
- full `test_chain_plugin` suite: 14 cases passed
- focused chain-plugin option registration/removal and response-size validation
- `snapshot_api_test` coverage for latest and block-specific snapshot bootstrap; Python syntax validation passed after removing the obsolete option
- `git diff --check`

The system-contract e2e gate is not applicable because the PR diff does not change `contracts/sysio.*`.

## Operational note

The selected snapshot endpoint is trusted and bootstrap is attended. Network phases therefore have no automatic deadline; a stalled provider waits for the operator to press Ctrl+C. Response bytes and disk usage remain bounded independently of that policy.


docs/snapshot-serving-plan.md
```diff
@@ -261,6 +261,12 @@ A single CLI-only option (not config file — single-use bootstrap):
                              https://snap.example.com/50000    → fetches block 50000
 ```
 
+The bounded download options are `--snapshot-endpoint-connect-timeout-ms`,
+`--snapshot-endpoint-header-timeout-ms`, `--snapshot-endpoint-idle-timeout-ms`,
+`--snapshot-endpoint-total-timeout-ms`, `--snapshot-endpoint-max-download-size-mb`, and
+`--snapshot-endpoint-min-disk-free-mb`. They enforce finite phase/total deadlines, fixed-length and chunked response
```
heifner: This seems like way too many options to me. I evaluated adding something similar and decided there were not really any good defaults. A user can always hit ctrl-c if it is taking too long. As for the size, any default will someday be too small. Again a user can just hit ctrl-c and try a new endpoint. I don't think we need any of these options.

heifner (changes requested):

docs/snapshot-serving-plan.md
```diff
@@ -261,6 +261,12 @@ A single CLI-only option (not config file — single-use bootstrap):
                              https://snap.example.com/50000    → fetches block 50000
 ```
 
+Snapshot bootstrap is an attended operation against an operator-selected endpoint. It reports connection/request phases
+and, every five seconds during transfer, downloaded bytes, percentage, rate, and ETA when `Content-Length` is available.
+SIGINT cancels pending resolver or socket work and retains atomic temporary-file cleanup. The existing
+`chain-state-db-size-mb` setting bounds both fixed-length and chunked response bodies; the only snapshot-specific
+resource override is `--snapshot-endpoint-min-disk-free-mb` for reserved filesystem headroom.
```
heifner: Instead of `--snapshot-endpoint-min-disk-free-mb` can we lean on resource monitor plugin? Can the monitoring there kill the download?

huangminghuang: Not as the code is structured today. The snapshot fetch runs synchronously in `chain_plugin::plugin_initialize()`, while `resource_monitor_plugin` starts its monitoring thread in `plugin_startup()`, so the monitor is not running during this bootstrap download. Also, the chain plugin currently registers only `blocks_dir` and `state_dir`; the download goes to `data-dir/snapshots` (often the same filesystem, but not guaranteed).

Once the resource monitor is active for the relevant filesystem, it can stop the download gracefully: on threshold breach it calls `app().quit()`, the HTTP client polls `app().is_quiting()` every 100 ms and cancels the active resolver/socket operation, and the `.downloading` temp file is removed while unwinding.

So we can remove the snapshot-specific free-space option and lean on the resource monitor, but it needs lifecycle integration first—for example, registering the snapshots filesystem and exposing/performing a synchronous threshold check during initialization. Moving the fetch to plugin startup would be more invasive because the snapshot is needed while initializing the chain. We should also decide how `--resource-monitor-not-shutdown-on-threshold-exceeded` applies, since enabling that option would intentionally remove the shutdown guarantee.

heifner: My vote would be to just remove the option. You set bounds on it via `chain-state-db-size-mb`. That is enough in my mind.

libraries/libfc/src/network/http/http_client.cpp
```diff
@@ -460,28 +701,66 @@ class http_client_impl {
...
 
-      http::file_body::value_type file;
-      file.open(temp_path.c_str(), boost::beast::file_mode::write, ec);
-      FC_ASSERT(!ec, "Failed to open temp file {} for writing: {}", temp_path.string(), ec.message());
-      parser.get().body() = std::move(file);
+      std::ofstream file(temp_path, std::ios::binary | std::ios::trunc);
+      FC_ASSERT(file.is_open(), "Failed to open temp file {} for writing", temp_path.string());
 
-      if (!parser.is_done()) {
+      std::vector<char> read_buffer(download_read_buffer_bytes);
```
heifner, line 745: `read_buffer` is 1 MiB, but each `async_read_some` decodes at most ~64 KiB of body: Beast's internal read-size hint caps a single socket read at 64 KiB, and `download_parser_buffer_bytes` caps the `flat_buffer` at 64 KiB regardless. Since `body().size` is reset to the full 1 MiB every iteration and only ~64 KiB is ever consumed, it never reaches 0 — so the `http::error::need_buffer` branch just below is effectively unreachable, and the buffer is ~16x larger than any iteration can use. Not a bug, but consider sizing `read_buffer` to the real per-read max (or a comment explaining the 1 MiB choice) and dropping/annotating the `need_buffer` handling.

libraries/libfc/src/network/http/http_client.cpp
```diff
@@ -411,46 +572,126 @@ class http_client_impl {
...
-         // Error responses carry small diagnostic bodies; read into memory for the message.
-         // Best effort: on a read failure the throw below reports whatever body arrived.
-         http::response_parser<http::string_body> err_parser{std::move(parser)};
-         if (!err_parser.is_done()) {
-            std::visit([&](auto& stream) {
-               return sync_read_parser_with_timeout(*stream, buffer, err_parser, deadline);
-            }, conn_iter->second);
-         }
-         FC_THROW("HTTP POST failed with status {}: {}", (int)status, err_parser.get().body().substr(0, 200));
+         FC_THROW("HTTP POST failed with status {}", parser->get().result_int());
```
heifner: This drops the response body from the diagnostic. The prior implementation read the (bounded) error body and included up to 200 chars, so an operator could see *why* the endpoint refused (`snapshot not ready`, `block pruned`, etc.). Now only the numeric status survives. Consider reading a small bounded error body before throwing to keep that signal without reintroducing unbounded buffering.

libraries/libfc/src/network/http/http_client.cpp
```diff
@@ -411,46 +572,126 @@ class http_client_impl {
...
+      }
+
+      const auto content_length = parser->content_length();
+      if (content_length) {
+         response_body_bytes = *content_length;
+      }
+      uint64_t disk_space_write_budget = 0;
+      if (content_length) {
+         FC_ASSERT(*content_length <= options.max_response_body_bytes,
```
heifner: This assertion is unreachable in practice: `sync_read_header_with_timeout` already returns `http::error::body_limit` (handled a few lines up) whenever `Content-Length` exceeds `body_limit(max_response_body_bytes)`, so reaching this point implies `*content_length <= max` already holds. Harmless as defense-in-depth, but either drop it or note that it guards against a future Beast behavior change so a later reader doesn't treat it as load-bearing.

heifner (commented):
Three low-severity cleanup notes on the bounded download path — none blocking. The size ceiling, disk-headroom checks, cancellation, and stale-connection retry all look correct and are well covered by the new tests.

docs/snapshot-serving-plan.md
huangminghuang: Addressed in `a1d7115279`. I removed `--snapshot-endpoint-min-disk-free-mb` from registration, runtime handling, docs, and integration invocation. Snapshot responses remain bounded by `chain-state-db-size-mb`; the HTTP client still checks each write against available disk space with its internal 64 MiB concurrent-consumer margin.

libraries/libfc/src/network/http/http_client.cpp
huangminghuang: Addressed in `a1d7115279` by removing the redundant post-header `Content-Length <= max_response_body_bytes` assertion. Beast’s configured `body_limit` remains the load-bearing fixed-length check; the separate overflow guard for adding the disk safety margin remains.

libraries/libfc/src/network/http/http_client.cpp
huangminghuang: Addressed in `a1d7115279`. The body buffer is now 64 KiB to match Beast’s per-read behavior. I retained and documented the `need_buffer` normalization because with the right-sized buffer it legitimately indicates that the caller-owned output buffer was filled and the decoded bytes should be processed before supplying the next buffer.

libraries/libfc/src/network/http/http_client.cpp
huangminghuang: Addressed in `a1d7115279`. Non-200 responses now read at most one 200-byte decoded body prefix and include it in the exception while retaining cancellation and bounded memory. A regression test verifies that the useful prefix is present and the suffix beyond the bound is omitted.

heifner: Thanks — the bounded read is correct. I walked the edges: absent/empty body (`is_done()` after header) throws status-only, a body already buffered with the header is parsed straight from the `flat_buffer` with nothing lost, and a mid-read cancel still surfaces because the outer `catch` converts to `interrupt_exception` when quitting.

One follow-up: the `buffer_body` increment here — set `body().data`/`size`, `sync_read_parser_some_with_timeout`, clear the benign `need_buffer`, then derive bytes decoded as `sink_size - body().size` — is now identical to the main download loop. Per the no-duplicated-helpers guideline it's worth factoring into one small member helper that returns the byte count and leaves the still-set `ec` for the caller to interpret (the loop throws on `body_limit`/transport errors; the error-body path ignores them, best-effort):

```cpp
size_t read_body_increment(auto& conn, boost::beast::flat_buffer& buffer, auto& parser,
                           char* sink, size_t sink_size, const operation_monitor& monitor, error_code& ec) {
   parser.get().body().data = sink;
   parser.get().body().size = sink_size;
   ec = std::visit([&](auto& stream) {
      return sync_read_parser_some_with_timeout(*stream, buffer, parser, deadline_type::max(), monitor);
   }, conn);
   if (ec == http::error::need_buffer) ec = {};
   return sink_size - parser.get().body().size;
}
```

Then this block becomes `error_body.resize(read_body_increment(...))` and the loop `const auto chunk_bytes = read_body_increment(...)` — same behavior, one place to maintain the parser handling. Not blocking.

huangminghuang: Addressed in `2260b339b0`. I added one typed `read_body_increment` member helper and routed both the bounded error-body prefix and the main download loop through it. The helper now owns `buffer_body` sink setup, the cancellable parser increment, benign `need_buffer` normalization, and decoded-byte accounting; callers retain their existing strict vs. best-effort error handling. All 20 focused HTTP cases and all 14 chain-plugin cases pass.

plugins/chain_plugin/src/chain_plugin.cpp
```diff
@@ -882,7 +980,26 @@ void chain_plugin_impl::plugin_initialize(const variables_map& options) {
...
                     "--snapshot-endpoint is incompatible with --snapshot; use one or the other");
-         fetch_snapshot_from_endpoint(options.at("snapshot-endpoint").as<std::string>());
+
+         const auto max_download_bytes = checked_mebibytes(
+            options.at("chain-state-db-size-mb").as<uint64_t>(), "chain-state-db-size-mb");
+         const fc::http_file_download_options download_options{
+            .max_response_body_bytes = max_download_bytes,
+            // Do not add a snapshot-specific reserve option. The HTTP client still checks that
+            // each bounded write fits on disk with its internal concurrent-consumer margin.
+            .min_free_disk_space_bytes = 0,
```
heifner: `min_free_disk_space_bytes` is dead in production — this is the only caller and it's pinned to 0, so the reserve never applies; the field's only non-zero uses are three tests simulating a full disk. Since the always-on 64 MiB `disk_space_concurrency_margin_bytes` is the real headroom guard, consider removing `min_free_disk_space_bytes` from `http_file_download_options` (and the `reserved_bytes` arg it threads through `require_disk_space`/`refresh_disk_space_write_budget`) instead of leaving an always-0 knob. Production behavior is unchanged. The catch is those three disk-headroom tests rely on this field as their only deterministic way to hit the disk-full path — they'd be dropped with it (margin protection stays, just not injection-tested). If that coverage matters, a test seam on the free-space query is cleaner than a prod field no caller sets.

huangminghuang: Addressed in `c64f6c6f73`. I removed `min_free_disk_space_bytes` from `http_file_download_options`, removed the `reserved_bytes` plumbing, and simplified the production disk check to the always-on 64 MiB safety margin. The three deterministic disk-failure cases remain covered through a private `http_client_test_access` seam that overrides only the reported available bytes. All 20 focused HTTP cases and all 14 chain-plugin cases pass.

heifner (approved):

PR #534, opened 2026-07-22
Validate outpost signer references and Ethereum chain IDs

huangminghuang:
## Summary

- require every Ethereum and Solana outpost client signer reference to match an explicitly named `--signature-provider`
- enforce the correct target chain and native key type for each outpost
- reject empty client IDs, signer names, and RPC URLs with plugin configuration errors
- validate explicitly configured Ethereum chain IDs with `eth_chainId` during startup using checked `uint32_t` parsing
- bound Ethereum client construction, DNS resolution, and explicit chain-ID validation to five seconds
- retain the three-field Solana client schema and signer-integrity checks without adding partial cluster-identity behavior
- add deterministic loopback RPC fixtures, startup validation coverage, provider-provenance tests, and updated operator documentation

### Signature-provider references

Outpost client configuration previously resolved signer references through the general provider registry. Anonymous `key-N` aliases and programmatically created providers could therefore satisfy a reference even though the operator had not supplied a stable shared name.

Both Ethereum and Solana clients now require the referenced name to come from the explicit five-field `--signature-provider` form. The provider must also use the target chain's native key type.

### Ethereum chain IDs

The optional fourth Ethereum client field is parsed as a positive 32-bit decimal integer or Ethereum hex quantity. When present, startup calls `eth_chainId`, rejects malformed/error responses, and fails on mismatch before registering the client.

The legacy three-field Ethereum format intentionally remains non-probing for compatibility: when no expected chain ID is configured, client construction still validates the URL and resolves DNS, but startup does not perform a reachability/identity RPC.

### Solana scope

`--outpost-solana-client` remains exactly:

```text
<client-id>,<sig-provider-id>,<rpc-url>
```

PR #534 validates the explicit signer reference, Solana target chain/key type, and non-empty fields. It deliberately does not add an optional genesis hash or a startup `getGenesisHash` probe. Mandatory startup pinning, runtime freshness, bypass-resistant gating, and telemetry belong to [SEC-139 / WSA-225](https://wire-network.atlassian.net/browse/SEC-139) and will be implemented as a separate complete control.

## Review follow-up

- deduplicated Ethereum exception translation and changed construction failures to neutral wording
- preserved three-field Ethereum no-probe compatibility instead of expanding the requirement to unconditional reachability checks
- retained the registered outpost chain-ID width as `uint32_t`; wider EVM chain IDs require a separate protocol/contracts migration
- removed the partial Solana cluster-identity implementation and its obsolete review concern while retaining Solana signer hardening
- added Solana empty-field coverage and explicit rejection of a fourth client field
- merged current `origin/master` (`d224ba5606`) and retained its role-aware Solana IDL/envelope-layout decoding while resolving the overlapping test include block

## Validation

- rebuilt Release targets:
  - `test_signature_provider_manager_plugin`
  - `test_outpost_ethereum_client_plugin`
  - `test_outpost_solana_client_plugin`
- focused CTest suites: 3/3 passed after the `origin/master` merge
- `git diff --check origin/master`
- full 11-file PR-only diff and 3-commit review
- independent Claude C++/security review of the feature diff: `CLEAN`

Earlier validation of the unchanged JSON-RPC transport and compatibility path also passed `json_rpc_client_tests` (10/10) and 100 repeated Ethereum three-field no-RPC runs.

## Notes

The feature branch was created directly from `origin/master` at `c3b84b93bf` and now includes merge commit `099e239bc1` from current `origin/master` (`d224ba5606`) to resolve the GitHub conflict. It does not include or depend on PR #516.


plugins/outpost_ethereum_client_plugin/src/outpost_ethereum_client_plugin.cpp
```diff
@@ -8,13 +11,114 @@ namespace sysio {
...
+         "Unable to initialize outpost Ethereum client '{}': the configured RPC endpoint is invalid or could not be resolved",
+         client_id);
+   } catch (const std::exception&) {
+      FC_THROW_EXCEPTION(
+         chain::plugin_config_exception,
+         "Unable to initialize outpost Ethereum client '{}': the configured RPC endpoint is invalid or could not be resolved",
+         client_id);
+   }
+
+   if (configured_chain_id) {
```
heifner, line 98: The legacy 3-field spec (no chain id) constructs the client but skips the `eth_chainId` call, so a dead or mistyped RPC URL still boots successfully and only fails later during relay. Solana, by contrast, always validates via `getGenesisHash`. Consider an unconditional lightweight reachability probe (e.g. `eth_chainId`/`net_version`, result unused when no chain id is configured) so the 3-field form also fails at startup, matching the PR's stated goal of failing before relay work begins.

plugins/outpost_ethereum_client_plugin/include/sysio/outpost_ethereum_client_plugin.hpp
```diff
@@ -19,7 +19,7 @@ struct ethereum_client_entry_t {
    /// batch operator auto-select the client for an outpost row by matching the
    /// row's `external_chain_id`, so multiple EVM outposts never share one
    /// remote endpoint. `nullopt` when the spec omitted the chain id.
-   std::optional<uint64_t>            chain_id;
+   std::optional<uint32_t>            chain_id;
```
heifner, line 22: Narrowing `chain_id` to `uint32_t` to match the canonical `external_chain_id` width (proto `types.proto`, `sysio.chains`, and the batch operator's `static_cast<uint32_t>`) is the right call for consistency here. Flagging the ceiling only: EVM chains with an id > 2^32 (e.g. Palm = 11297108109) can't be represented. Supporting those would mean widening `external_chain_id` to `uint64_t` across proto/contracts/plugins, which is a separate follow-up PR rather than part of this change. Noting it so the ceiling is a conscious decision.

plugins/outpost_ethereum_client_plugin/src/outpost_ethereum_client_plugin.cpp
```diff
@@ -8,13 +11,114 @@ namespace sysio {
...
+   const std::optional<uint32_t>&              configured_chain_id) {
+   fc::task::deadline_scope deadline(fc::time_point::now() + chain_id_validation_timeout);
+
+   ethereum_client_ptr client;
+   try {
+      const auto client_chain_id = configured_chain_id
+         ? std::optional<fc::uint256>{fc::uint256{*configured_chain_id}}
+         : std::nullopt;
+      client = std::make_shared<ethereum_client>(signature_provider, url, client_chain_id);
+   } catch (const fc::exception&) {
```
heifner, line 92: Two small notes on this catch pattern (both apply to all four sites: eth construction + execute, solana construction + get_genesis_hash):

1. The `catch (const fc::exception&)` / `catch (const std::exception&)` pair has byte-identical bodies. `fc::exception` doesn't derive from `std::exception`, so both arms are needed, but the duplicated message/throw could collapse into a small helper (e.g. `throw_endpoint_unreachable(client_id)`) so the wording can't drift between the four copies.

2. The wording assumes any construction failure is a bad/unresolvable URL, but `make_shared<ethereum_client>` can throw for other reasons (bad signer, internal init). A more neutral "failed to initialize" phrasing avoids sending operators to debug the URL for an unrelated failure.

plugins/outpost_solana_client_plugin/src/outpost_solana_client_plugin.cpp
```diff
@@ -10,11 +12,77 @@ namespace {
...
+         "Unable to validate outpost Solana client '{}': the configured RPC endpoint did not "
+         "return a valid getGenesisHash result",
+         client_id);
+   }
+
+   SYS_ASSERT(!expected_genesis_hash || remote_genesis_hash == *expected_genesis_hash,
+              chain::plugin_config_exception,
+              "Genesis hash mismatch for outpost Solana client '{}': configured {}, RPC endpoint reports {}",
+              client_id,
+              expected_genesis_hash.value_or("none"),
```
heifner: `expected_genesis_hash.value_or("none")` is dead code: the assertion predicate `!expected_genesis_hash || ...` is only false when `expected_genesis_hash` is engaged, so the message never renders the `"none"` branch. Can be `*expected_genesis_hash` (or drop the fallback).

heifner (commented):
Reviewed the outpost startup-validation changes. Core logic is solid: `get_client_by_chain_id` fails closed on ambiguity, `parse_chain_id` is overflow-guarded, and the `deadline_scope` is genuinely honored by the json-rpc transport via `current_deadline()`. A few notes below, mostly minor.

plugins/outpost_solana_client_plugin/src/outpost_solana_client_plugin.cpp
```diff
@@ -94,13 +95,36 @@ void outpost_solana_client_plugin::plugin_initialize(const variables_map& option
...
+      SYS_ASSERT(sig_provider->target_chain == fc::crypto::chain_kind_solana &&
+                    sig_provider->key_type == fc::crypto::chain_key_type_solana,
+                 chain::plugin_config_exception,
+                 "Outpost Solana client '{}' signer '{}' must use chain=solana and key-type=solana",
+                 id,
+                 sig_id);
+
       my->add_client(id, std::make_shared<solana_client_entry_t>(
                             id, url, sig_provider,
                             std::make_shared<solana_client>(sig_provider, url)));
```
heifner, line 130: Removing the startup `getGenesisHash` check leaves a Solana outpost with no cluster-identity verification: it will boot pointed at the wrong cluster (e.g. devnet vs mainnet) and only fail during relay. For a hardening PR that's a defense-in-depth reduction.

The asymmetry flagged earlier was that Solana validated *unconditionally* while eth only validates when a `,<chain-id>` is supplied. A symmetric, opt-in fix keeps parity without dropping the capability: eth checks `eth_chainId` only when the chain-id field is present (already the case), and Solana keeps the optional `,<genesis-hash>` exact-match while dropping only the *unconditional* probe. The code removed here already had that conditional branch, so this restores just that half — operators who want cluster pinning still get it. Not blocking; flagging for consideration.

heifner (commented):
Re-reviewed `b9a58cfdac`. The helper extraction cleanly resolves the catch-block feedback (acknowledged on the individual threads), and the removal is consistent across the plugin, README, and tests. One doc note and one design note:

The **## Why** and **## Validation** sections of the PR description still describe Solana `getGenesisHash` startup validation, which this commit removed. Worth updating so the description matches the shipped behavior. Inline note below on the cluster-identity tradeoff itself.

plugins/outpost_ethereum_client_plugin/src/outpost_ethereum_client_plugin.cpp
heifner: Resolved — `throw_client_initialization_failure` / `throw_chain_id_validation_failure` remove the duplicated bodies, and the neutral "Failed to initialize" wording no longer blames the URL for unrelated construction failures. Thanks.

plugins/outpost_solana_client_plugin/src/outpost_solana_client_plugin.cpp
heifner: Resolved — the dead `value_or("none")` branch is gone with the removal of the Solana genesis-hash validation. See the follow-up review note on the cluster-pinning tradeoff.

plugins/outpost_ethereum_client_plugin/src/outpost_ethereum_client_plugin.cpp
heifner: Acknowledged — the asymmetry is resolved by dropping Solana's startup validation rather than adding an eth probe. Defensible call; left a follow-up note weighing the defense-in-depth tradeoff and an opt-in symmetric alternative that keeps Solana cluster pinning available.

plugins/outpost_ethereum_client_plugin/include/sysio/outpost_ethereum_client_plugin.hpp
huangminghuang: Confirmed. The 32-bit ceiling is intentional because the registered outpost `external_chain_id` is currently `uint32_t` across the protocol/contracts/plugin boundary. The updated PR description now records that supporting wider EVM chain IDs requires a separate protocol and contracts migration.

plugins/outpost_solana_client_plugin/src/outpost_solana_client_plugin.cpp
huangminghuang: Thanks. We are intentionally not retaining the optional fourth field. SEC-139 requires every signing-capable Solana client to have exactly one operator-configured expected genesis identity and requires a missing identity to fail startup. Keeping an optional pin here would formalize the opposite contract and allow production clients to remain unpinned. PR #534 therefore keeps only Solana signer-reference/chain/key validation; mandatory startup pinning, runtime freshness, and bypass-resistant gating will land together in the separate SEC-139 PR. The PR description has been updated to make that boundary explicit.

heifner: Fully agree — SEC-139/WSA-225 mandates exactly one genesis identity per signing-capable Solana client with fail-closed startup and explicitly no TOFU/optional mode, so an optional pin here would formalize the wrong contract. Withdrawing the suggestion; removing the partial implementation and landing the complete gate in SEC-139 is the right sequencing. Thanks for the detail.

heifner (approved):
Approving the outpost startup-validation changes.

Reviewed through the `origin/master` merge (`099e239bc1`): the Ethereum and signature-provider-manager changes are unchanged from prior review, the Solana signer/chain/key/empty-field validation survived the merge intact, and the merge's `solana_outpost_role` threading correctly matches master's `create_outpost_client` signature.

Signer references now require an explicitly named `--signature-provider` with the correct target chain and key type; explicit Ethereum chain IDs are parsed as checked `uint32_t` and verified against `eth_chainId` within a five-second bound; `get_client_by_chain_id` fails closed on ambiguity. The `uint32_t` chain-id ceiling and the deferral of mandatory Solana cluster pinning to SEC-139/WSA-225 are both intentional and documented.

All review threads resolved. Build/test CI is still running on the merge head, so this is the code sign-off; recommend merging only once CI is green.

PR #535, opened 2026-07-24
SEC-134/SEC-136: Harden authenticated outbound HTTP transport

huangminghuang:
## Summary

- add one authenticated Boost.Beast/Boost.Asio transport core for the SEC-134 nodeop/libfc HTTP paths while retaining the existing synchronous `http_client`, JSON-RPC, Ethereum, and Solana APIs
- enforce TLS chain and DNS/IP identity verification, bounded DNS/connection/request/response work for ordinary callers, byte ceilings, cancellation, bounded/aged connection reuse, and explicit proxy/private-CA configuration
- let explicitly read-only Ethereum and Solana RPCs retry once only when a reused HTTP/1.1 connection proves stale; transaction submission, airdrops, generic calls, notifications, batches, and raw HTTP remain single-attempt
- expose generic retry-filter and one-step connection-affine continuation hooks without exposing public connection/session identifiers
- preserve the pre-refactor snapshot payload-transfer exception: no DNS/connect, upload, response-header, response-body, idle, total, or inherited task deadline
- expose fixed-cardinality process-wide outbound transport metrics

## Design and scope

The branch contains the SEC-136 authenticated Beast transport prerequisite plus SEC-134 policy hardening. Ethereum and Solana remain synchronous clients over the shared transport facade. `http_client_plugin` keeps one stable facade for its lifetime, and JSON-RPC construction performs no DNS or other network I/O.

The retry hook runs only after the transport classifies a failure as replayable and can only narrow eligibility. Attempts above one require explicit idempotence. Reusing a connection does not prove that a failed write transmitted no request bytes, so Ethereum `eth_send*`, Solana `sendTransaction`, airdrops, generic calls, notifications, batches, and raw HTTP remain conservative single-attempt operations.

The continuation hook buffers the first bounded response while retaining its exact HTTP/TLS connection outside the idle pool. It may reject or select one same-endpoint follow-up with independent limits and deadline budget. The follow-up never reconnects, retries, or falls back to another pooled connection. Validation failure, endpoint mismatch, peer close, cancellation, and follow-up failure close the retained connection. Callback re-entry into the same transport fails fast.

DNS resolution uses a fixed, bounded set of independent resolver services with event-driven admission. Linux creates resolver notification pipes atomically with `pipe2(O_CLOEXEC | O_NONBLOCK)`; macOS and other supported platforms retain the portable `pipe`/`fcntl` fallback. Idle pool residence is measured with `std::chrono::steady_clock`, capped by count and age, and probed for peer closure before reuse.

The option audit keeps trust roots, proxy selection, DNS/cache behavior, resource ceilings, deadlines, and idempotence as declarative policy. Shared transport-option and outpost-RPC helpers remove per-plugin duplication. Process-wide `--outbound-http-*` CA/proxy settings provide common fallbacks while caller-specific settings remain explicit overrides.

Snapshot metadata remains bounded by shared defaults. Only the subsequent snapshot payload transfer is intentionally unbounded, matching the pre-refactor behavior. It remains predicate-cancellable, response-size and disk-space bounded, cleaned up on failure, and atomically published after complete persistence.

`clio` remains the repository's separate libcurl consumer and is outside this nodeop/libfc transport boundary. No system-contract sources are changed, so the platform system-contract e2e gate does not apply.

## Review feedback addressed

- preserve exact KIOD Unix routes and map a bare socket URL such as `unix:///run/kiod.sock` to HTTP `/`
- add live Unix-socket regressions for both exact routes and the root route, with portable failure-path teardown
- measure idle connection age with a monotonic clock
- use atomic nonblocking, close-on-exec resolver pipes on Linux with the portable fallback elsewhere
- document why reused-connection metadata is insufficient to replay transaction submissions safely
- bound the idle pool, reject peer-closed cached sockets, make synchronous queueing deadline-aware, and fail fast on callback re-entry
- make resolver completion lifetime-safe and remove constructor-time JSON-RPC network I/O
- consolidate CA/proxy and outpost RPC option construction, document active `SSL_CERT_FILE`/`SSL_CERT_DIR` behavior, and retain caller-specific overrides
- keep snapshot payload downloads intentionally unbounded while preserving cancellation, resource checks, cleanup, and atomic publication
- make the default JSON-RPC single-attempt regression deterministic by recording the side-effect request before resetting its reused connection; any replay consumes a scripted success and fails the test

## Operator impact

- HTTPS uses platform trust plus mandatory endpoint identity verification. Caller-specific CA file/directory options add private roots without disabling verification.
- OpenSSL's default trust loading honors `SSL_CERT_FILE` and `SSL_CERT_DIR`.
- Proxy environment variables remain ignored. Proxying is configured through process-wide outbound HTTP options or caller-specific overrides.
- JSON-RPC clients establish connectivity on demand.
- Metrics omit caller and endpoint labels to retain fixed cardinality and avoid exposing endpoint details.
- Existing callers do not need to adopt the new hooks; default request behavior is unchanged.

## Validation

- affected Release `test_fc` and `nodeop` targets built successfully
- deterministic default-call single-attempt regression passed 100 consecutive runs
- complete JSON-RPC suite: 22 tests passed
- complete Release `test_fc`: 592 tests / 5,009,716 assertions passed
- focused HTTP, async transport, and file-download suites passed
- Solana outpost plugin: 48 tests / 249 assertions passed
- chain plugin: 15 tests / 1,385 assertions passed
- Ethereum outpost plugin: 24 tests / 77 assertions passed
- external debugging plugin: 7 tests / 22 assertions passed
- `git diff --check` passed
- fresh independent correctness and architecture reviews of the final CI follow-up reported no actionable findings
- current-head macOS and full Linux CI passed, including ASan, UBSan, GCC, Ubuntu 24, and assertions-on

## Residual risk

- snapshot payload transfers intentionally retain no automatic network deadline; cancellation and resource bounds remain the safety controls
- macOS-only paths were reviewed locally and remain covered by current-head CI

## Commits

1. `SEC-136: Add bounded authenticated Beast transport`
2. `SEC-136: Migrate JSON-RPC clients to authenticated transport`
3. `SEC-134: Harden synchronous outbound HTTP policies`
4. `SEC-134: Fix HTTP transport evaluation order`
5. `SEC-134: Restore unbounded snapshot downloads`
6. `SEC-134: Isolate aggregate read timeout test`
7. `SEC-134: Add generic HTTP transport hooks`
8. `Address PR review feedback`
9. `SEC-134: Address follow-up transport review`
10. `SEC-134: Stabilize single-attempt RPC test`


heifner (commented):
Review of the transport consolidation. One blocking correctness bug, several operational-robustness gaps in the new core, and a set of CLAUDE.md conformance items. The CA question is answered at the end.

## Blocking

**Unix-socket request target picks up a trailing slash, so every KIOD signing request 404s.**

In `normalize_target`, the unix-socket splitter walks parents building `request_path = socket_file.filename() / request_path` starting from an empty `request_path`. `std::filesystem::path("sign_digest") / std::filesystem::path("")` yields `"sign_digest/"`, so a provider configured as `KIOD:unix:///run/kiod.sock/v1/wallet/sign_digest` produces the target `/v1/wallet/sign_digest/`. I confirmed this with a standalone repro of the exact loop. kiod's `beast_http_session` dispatches through an exact `plugin_state_->url_handlers.find(resource)` lookup with no trailing-slash normalization, so the request comes back 404 "Unknown Endpoint" and `post_sync` throws "URL not found". The only unix-socket test added is `nonexistent_unix_socket_path_is_bounded`, so nothing exercises a live socket -- worth adding a test that stands up a real socket and signs a digest.

Secondary on the same path: the previous implementation memoized the split in `_unix_url_paths`; the new one re-runs up to N `is_socket()` stat() calls on every signing request.

## Operational robustness

**Idle-connection pool has no size cap, no age cap, and cannot detect a peer-closed socket.** `release_connection` pushes into `idle_connections[key]` unbounded with no expiry sweep, and `connection_state::open()` returns `socket().is_open()`, which is purely local fd state and stays true after the peer sends FIN. When an RPC provider closes on its keep-alive idle timeout, the next `json_rpc_client::call()` leases the dead connection, writes, and gets `end_of_stream`. That path is deliberately single-attempt (eth_sendRawTransaction, sendTransaction, notify, call_batch, send_http), so there is no retry to absorb it and submission just fails after every idle period. Sockets sitting in CLOSE_WAIT also accumulate for the process lifetime, since entries are only pruned when that key is next leased. A max-idle-age comfortably below typical server keep-alive fixes both.

**`transport_impl` serializes every request behind a blocking mutex taken before any deadline applies.** `perform`, `perform_then`, `prime_endpoint`, and `perform_to_file` all open with `std::scoped_lock lock(use_mutex)`, and the deadline is only computed inside the coroutine once the lock is held. http_client_plugin owns one `fc::http_client` shared by every KIOD signature provider, so one request stalling for its full total timeout blocks a concurrent caller carrying a 5s `fc::task::deadline_scope` for the whole duration, with unbounded queueing and no deadline-aware wait. Re-entrancy is also fatal here: a `status_callback` or `continuation_hook` that calls back into the same transport self-deadlocks on the non-recursive mutex while running on that transport's own io thread.

**All DNS in the process funnels through one semaphore slot with a 50ms busy-poll.** `platform_resolver_runtime` has `asio::thread_pool executor{1}` and `std::binary_semaphore in_flight{1}`, and `resolve()` spins on `cancellation_poll_interval` whenever `try_acquire` fails. With N ethereum clients plus N solana clients plus kiod plus external_debugging plus snapshot bootstrap, one unresponsive DNS server holds the only slot while everyone else burns their connect deadline and reports `timeout_connect`. The permit is released only when getaddrinfo actually returns -- `resolver->cancel()` does not reclaim it -- so a hung resolver stalls the whole process for the OS resolver timeout. This also makes `slow_response_headers_time_out` (10ms connect deadline) fail deterministically any time the slot is momentarily held, since the first poll wait alone is 50ms.

**Resolver completion can post to a destroyed io_context.** The handler in `start_platform_resolution` runs on the leaked global pool, does `weak.lock()`, then posts to `current->timer.get_executor()`. Cancellation is `resolver->cancel()`, which does not wait for the handler. If `resolve()` hits its deadline and throws, the coroutine unwinds and `transport_impl` (and its `asio::io_context` member) is destroyed, a pool thread that locked the shared_ptr just before the frame released it will post to an executor referencing freed storage. Rare, but a use-after-free.

**`json_rpc_client`'s constructor does a blocking DNS warm-up.** `prime_endpoint` in the constructor means `outpost_ethereum_client_plugin::plugin_initialize` (under the 5s chain_id_validation_timeout scope) and `external_debugging_plugin::plugin_startup` throw if the RPC hostname is briefly unresolvable at boot -- DNS not yet up on a restarting host, or the single resolver slot occupied per the point above. Previously construction did no I/O and the first call would simply retry. Suggest making the warm-up best-effort rather than a startup gate.

**`timeouts.idle` is armed as a whole-operation cap on the request upload.** `async_open` uses `phase_deadline(policy.timeouts.idle, failure_kind::timeout_idle, total_deadline)` for the write, but Beast's `expires_after` is a deadline for the entire logical operation, not an inactivity timer. A large request body over a slow-but-progressing link aborts and is reported as `timeout_idle`, whose documented meaning is "maximum time without upload or download progress after response headers". Operators tuning `idle` to catch stalls will instead be capping total upload duration, and a working link gets attributed to an idle stall in the metrics.

**`classify_tls_error` distinguishes verification from handshake failures by substring-matching OpenSSL prose.** It lowercases `error.message()` and looks for "certificate", "hostname", or "verify". An OpenSSL version bump that rewords a message silently reclassifies genuine chain rejections as generic `tls_handshake`, which corrupts the fixed-cardinality failure counters this PR adds and misleads anyone debugging a private-CA misconfiguration. The authoritative value is already in hand: the verify callback in `create_connection` has `X509_STORE_CTX_get_error(context.native_handle())`, and `SSL_get_verify_result(stream.native_handle())` is available post-handshake. Storing the verify code the way `identity_failure` is stored is more durable than parsing text.

**snapshot_api_plugin README contradicts the code.** The rewritten section says "The file transfer has finite connect, response-header, and no-progress deadlines", but `http_file_download_options::timeouts` initializes connect, header, read, idle, and total to `nullopt`, and chain_plugin's designated initializer for `download_options` omits `timeouts` entirely, so the default stands. The PR body itself states the opposite of the README. As written, an operator will assume a wedged snapshot provider self-recovers.

**`http_client_plugin` replaces the client in `plugin_initialize`.** The constructor does `my(new http_client())` and initialize reassigns `my`, while `get_client()` hands out `http_client&`. Today only signature_provider_manager_plugin uses it, and only lazily inside the sign lambda, so nothing dangles -- but nothing enforces that either. Constructing once (or returning a shared_ptr) removes the trap.

## CLAUDE.md conformance

**Rule 1, no duplicated helpers.** The `additional-ca-file` / `additional-ca-path` / `proxy` to `transport_options` block is written out verbatim five times (chain_plugin, external_debugging_plugin, http_client_plugin, outpost_ethereum_client_plugin, outpost_solana_client_plugin), and `solana_rpc_options` / `ethereum_rpc_options` are byte-identical apart from the option-name and byte constants. Adding a sixth transport knob means editing five files and will eventually miss one. Same rule inside libfc: `cancellation_poll_interval` is defined independently in http_client.cpp and http_client_legacy.cpp.

**Rules 2 and 3, magic literals and closed sets.** HTTP status codes are compared as bare integers throughout: `file_sink` tests `_status != 200` in four places, `request_metrics_state::finish_status` and `response_reader_impl` each open-code the `>= 200 && < 300` range, http_client_legacy tests `== 500` and `== 404`. `normalize_target` hardcodes "443"/"80", the proxy default hardcodes "80", and `build_request` passes a bare 11 for the HTTP version. Beast already provides `beast_http::status` and `to_status_class`; the remainder want file-local constexprs.

**Typed durations.** `dns_cache_timeout_seconds` is an `int64_t` with -1 meaning "cache forever" and 0 meaning "never", which costs three sentinel comparisons in `resolve()`, one in `normalize_options()`, and a range assert in the client_impl constructor, and leaves the unit discoverable only from the field name. `std::optional<fc::microseconds>` with nullopt for unlimited removes the sentinel and the assert, and matches how `timeout_options` and `retry_options` are already typed in the same header.

**Non-ASCII in docs.** docs/outbound-http-transport.md has two em-dash characters in its opening lines; these are the only non-ASCII bytes in the diff. Please use ` -- `.

## Tests

The checked-in TLS fixtures are dated time bombs. future.pem has notBefore=2030-01-01, so `not_yet_valid_certificate_is_rejected` stops testing anything and then fails outright on that date. ipv6.pem expires 2034-10-09, and ca.pem, dns.pem, ip.pem, and secondary-ca.pem expire 2036-07-20, at which point `private_ca_file_accepts_matching_dns_and_sends_sni`, `private_ca_accepts_matching_ip_without_sni`, `private_ca_accepts_matching_ipv6_without_sni`, and `concurrent_clients_do_not_share_private_ca_state` all start failing with no code change. Generating the CA and leaves in test setup relative to `time_point::now()` removes the cliff and drops nine fixture files from the diff.

## On the CA options

They are not needed for public CAs. `client_impl`'s constructor already calls `tls_context.set_default_verify_paths()`, which loads OpenSSL's compiled-in trust store (/etc/ssl/certs on Debian and Ubuntu, the system bundle on macOS) and honors the standard `SSL_CERT_FILE` and `SSL_CERT_DIR` environment variables. Infura, Alchemy, api.mainnet-beta.solana.com, and any publicly-signed snapshot host verify with no configuration at all. The new options only matter for private or internal roots.

Two things worth reconsidering there:

The env-var policy is inconsistent. The PR deliberately ignores `http_proxy` and `https_proxy` and requires explicit per-caller proxy options, but silently honors `SSL_CERT_FILE` and `SSL_CERT_DIR` through `set_default_verify_paths()`. The doc says "Proxy environment variables are ignored" and says nothing about the CA env vars that are active. Either both mechanisms are trusted or neither is, and either way it should be documented.

Five per-caller copies with no shared fallback is awkward for the common case. One corporate root typically covers the internal snapshot host, the internal ETH node, and the debugging server, and that operator has to pass the same path five times under five different flag names -- with a sixth flag needed for the next outbound consumer. A single `--outbound-http-additional-ca-file` / `--outbound-http-proxy` pair that all five callers default to, with per-caller overrides only where a caller genuinely needs a different root, covers every realistic deployment and collapses the duplication noted under rule 1. If the per-caller options stay, please at least document that `SSL_CERT_FILE` and `SSL_CERT_DIR` already work, so operators are not told the flags are the only path.

heifner (commented):
Re-reviewed against `3e33037`. Every point from the previous round is addressed, and the fixes are substantive rather than symptom patches. Notes on what I actually verified, then two small new items and two things worth confirming.

## Verified fixed

The unix path splitter now uses `lexically_relative` and restores memoization through `unix_target_cache`, so the KIOD target comes out as `/v1/wallet/sign_digest` (confirmed by compiling the expression against the three relevant input shapes).

The connection pool gained `max_idle_connections`, `max_idle_connection_age`, and `prune_expired_idle_connections`, but the part that actually closes the original hole is `healthy_for_reuse` doing a `MSG_PEEK | MSG_DONTWAIT` recv -- treating both EOF and unexpected readable bytes as unhealthy is the right polarity, and it catches the peer-FIN case that `socket().is_open()` structurally cannot. I traced `idle_connection_count` through every early return and throw path in `acquire_connection`, `release_connection`, and the prune loop; the accounting balances and it all runs on the strand.

The resolver rework holds up. Four `platform_resolver_worker` instances each with their own `io_context` is the correct shape given that Asio serializes `getaddrinfo` within one execution context, so extra threads on a shared context would not have bought parallelism. The one-shot `notified` flag looked like a starvation risk on first read -- a waiter that wakes, loses the `try_acquire` race, and re-registers with an already-notified notifier would never be woken again -- but a fresh `resolver_event` is constructed per iteration of the capacity-wait path, so it does not arise. `release()` locking the weak_ptrs under the runtime mutex and calling `notify()` outside it, against `resolver_event`'s destructor calling `disable()` under the notifier's own mutex, is safe in both interleavings. Capturing the permit in the `async_resolve` completion handler also means cancellation now reclaims the slot, which the previous version did not.

`use_guard` handles the thread-local correctly: `active_transport` is assigned last in the constructor body and restored in the destructor, so a constructor that throws on the busy, deadline, or cancel path leaves it untouched and there is nothing to unwind. The fail-fast branch for callers with neither a deadline nor a `cancel_check` is effectively unreachable in tree -- the snapshot download is the only caller that nulls its timeouts and it sets a `cancel_check` -- so it reads as a defensive guard rather than a behavior change.

`classify_tls_error`'s prose matching is gone in favor of `SSL_get_verify_result` plus verify-callback-recorded chain and identity state, the status literals are behind `status_value` and `to_status_class`, `dns_cache_timeout` is a typed optional, `timeouts.idle` is now armed per `async_read_some` chunk, the fixtures are generated with `X509_gmtime_adj` relative to now, and the snapshot README matches the code.

Thanks for taking the shared-options suggestion -- the global-then-caller layering in `read_transport_options` overriding per field rather than all-or-nothing is better than what I proposed, and documenting that `SSL_CERT_FILE` and `SSL_CERT_DIR` stay active closes the gap between the CA and proxy env-var policies.

## New, both minor

**Bare-socket unix URL yields the target `/.`** The `lexically_relative` rewrite fixes the trailing slash, but `std::filesystem::path::lexically_relative` returns `.` rather than an empty path when the two paths are identical, so the `request_path.empty()` guard never fires for a URL that names only the socket. `unix:///run/kiod.sock` produces the request target `/.` instead of `/`. Not reachable through a signature provider, since those always carry a path, but `post_sync` accepts any unix URL and addressing a server's root is the natural case. `request_path.empty() || request_path == "."` covers it.

**Idle-connection age is measured on the wall clock.** `idle_connection::idle_since` is an `fc::time_point` while the adjacent `dns_entry::expires` uses `steady_clock`, and the prune loop compares `time_point::now() - idle_since` against `max_idle_connection_age`. An NTP step backwards makes that difference negative, so nothing is pruned by age and the pool falls back to the `MSG_PEEK` check alone, reopening a narrower version of the window the age cap exists to close; a forward step evicts the whole warm pool at once. Monotonic time is the right clock for an elapsed-duration comparison, and it would match what `dns_entry` already does.

## Worth confirming, not defects

`eth_sendRawTransaction` and Solana `sendTransaction` are the only paths left on the single-attempt policy -- every read method routes through `execute_idempotent` with `stale_reused_connection_once`. Never auto-replaying a submission is a defensible default and the 15s idle cap makes the stale-connection window small now, but the `retry_context.reused_connection` predicate exists precisely to scope a replay to failures where the request provably never reached the server, and resubmitting an identical signed transaction is deduplicated by the node in both chains. If that is deliberate rather than incidental, a comment on the `execute` versus `execute_idempotent` split would keep a future reader from "fixing" it.

`resolver_event` creates its pipe with `::pipe` and then sets `FD_CLOEXEC` via `fcntl`. `pipe2(fds, O_CLOEXEC | O_NONBLOCK)` sets both atomically and removes the window where a concurrent fork/exec inherits the descriptors.

CI is still pending across all builds on this commit; I have not seen a green run for the new resolver and pool code.

heifner (dismissed):
Approving. Everything raised across both review rounds is resolved.

I verified the three fixes in `42ead0f` by compiling the exact expressions rather than reading them: the `root_request` guard yields `/` for both `unix:///run/kiod.sock` and its trailing-slash form while still producing `/v1/wallet/sign_digest` for the KIOD case, and `unix_socket_root_request_target_is_preserved` asserts `POST / HTTP/1.1` so it is a real regression test. The `idle_clock` alias moves connection residence time onto `steady_clock` end to end, and the `fc::microseconds` to `std::chrono::microseconds` bridge behaves correctly at the boundary -- a 15s cap prunes a 16s-idle connection and retains a 14s one. `create_resolver_pipe` uses `pipe2(O_CLOEXEC | O_NONBLOCK)` on Linux with the `pipe` plus `fcntl` path kept for macOS; `pipe2` needs `__USE_GNU`, and both clang-18 and g++ declare it without an explicit `_GNU_SOURCE` because the C++ drivers define it on Linux, so that branch builds.

You are right about the replay policy, and the comment you added on `stale_reused_connection_once` states the reason better than my suggestion did: reusing a connection does not prove a failed write transmitted no request bytes. A partial write before failure is exactly the case `reused_connection` cannot exclude, so keeping transaction submission on the single-attempt path is correct.

Two notes rather than requests. CI has not completed on any of the three commits I reviewed, so this approval covers the code and not a passing build -- worth holding the merge until the resolver and connection-pool changes have been through asan and ubsan in particular, since neither the pipe notifier nor the idle-pool accounting has executed under a sanitizer yet. And the PR is still marked draft.

heifner (approved):
Re-approving after the dismissal. `da648f81` is test-only and the change is the right one.

The previous form of `default_call_enforces_single_attempt` scripted a reset immediately after the first response and then asserted both a throw and `connection_count() == 2`, which raced: if the reset landed before the client's `MSG_PEEK` health check the pooled connection was rejected and a fresh one opened, and if it had not arrived the client reused the doomed connection and failed on it. Both paths throw, so only the connection count flapped. Letting the server close in response to the second request instead makes the pooled connection provably healthy at reuse time -- the server is parked in a read, so the peek returns `EAGAIN` -- and the outcome is one connection and two requests every time.

The assertions came out stronger rather than weaker. Checking `requests.front().method` and `requests.back().method` proves `wire_side_effect_probe` was transmitted exactly once, the scripted `result("replayed")` is never consumed, and the client is still built with the most permissive retry configuration a caller could supply, so the case still tests what its name claims. Stale-connection recovery remains covered by `idempotent_call_recovers_from_a_stale_cached_connection` through the other harness, and `reset_after_response` is still exercised by two other cases, so nothing became dead code.

This also clears the reservation I attached to the earlier approval: the resolver pipe notifier and the idle-pool accounting have now run under asan and ubsan, along with gcc, asserton, ubuntu24, and the Apple Silicon job.

PR #536, opened 2026-07-24
epoch/msgch: remove the zero-consumer EMISSIONS_BLOCKED broadcast; hard-fail bootstrap on missing emissions config

heifner:
## Problem

When the emissions readiness gate blocks epoch advance, `sysio.epoch` broadcast an `EMISSIONS_BLOCKED` attestation to every active outpost. **Nothing consumes it anywhere** — not the Ethereum outpost, not the Solana program, not any tooling. Worse, the queued attestation rides the *first post-recovery envelope* (envelopes only go out when the epoch finally advances), and the Ethereum outpost's inbound dispatch reverts on attestation types with no registered handler — inside the consensus-tipping delivery. Net effect: the moment the emissions gate recovers (treasury refilled, config fixed), the first envelope out permanently wedges the ETH outpost and halts OPP chain-wide, recoverable only by a governance `blackholeAttestationTypes` call. The broadcast's only real-world effect was arming that wedge.

The gate itself is sound and stays: `TREASURY_EXHAUSTED` and `BALANCE_INSUFFICIENT` are genuine operational states the depot must pause on, with `blocklog` + the console diagnostic as the ops surface.

## Change

1. **Remove the cross-chain broadcast.** `emit_emissions_block_attestation` is deleted; `record_gate_block` keeps writing/updating the `blocklog` row exactly as before (reason, amounts, `first_blocked_at`, `retry_count` semantics unchanged). The gate state is depot-local by design now: a depot pause needs no outpost cooperation — envelopes simply stop until the gate passes, which existing stall monitoring already detects.
2. **Hard-fail bootstrap on missing emissions config.** `sysio.msgch::bootstrap` (the production bootstrap entry that fires the genesis advance) now `check()`s that `sysio.system::setemitcfg` and `initt5` have run. Previously a misconfigured chain gate-blocked *silently* at its first advance and surfaced only as "epoch stuck at 0". The two set-up-once reasons (`CONFIG_MISSING` / `STATE_UNINITIALIZED`) are therefore unreachable on a properly-bootstrapped chain; the gate's never-throwing block-and-retry behavior in `advance()` is unchanged for the economic reasons (and for unit fixtures that deliberately run against a blocked gate).
3. **Proto: comment-only deprecation** of `ATTESTATION_TYPE_EMISSIONS_BLOCKED` and `message EmissionsBlocked` ("no longer emitted; do not reuse" — pretoken precedent). No structural proto change, so no bundle regeneration and no ABI delta anywhere. `EmissionsBlockReason` stays (used by the `blocklog` row). The `EMISSIONS_BLOCKED` entry in msgch's `estimate_svm_dynamic_accounts` is deliberately retained to protect the upgrade edge where an already-queued attestation exists when new code lands.

## Tests

- New `bootstrap_requires_emissions_config` (dispatch tests): the fixture deploys no `sysio.system` — exactly the misconfigured-deployment shape — and `msgch::bootstrap` now refuses with a clear message.
- All existing gate coverage passes unchanged: the emissions suite's gate-block tests assert only `blocklog` rows (dedup, reason change, clear-on-unblock), which this change preserves verbatim; `chkcons_survives_non_advancing_advance` still exercises the never-throwing block-and-retry path.
- `contracts_unit_test -- --sys-vm`: no errors detected.

## Companion

Wire-Network/wire-ethereum#171 hardens the ETH outpost's inbound dispatch for the general case (any future unconsumed type logs-and-skips instead of reverting, matching the Solana outpost and the depot's own never-throw dispatch). E2e-gate runs of this branch should set `BRANCH_WIRE_ETHEREUM` to that branch until it merges. No `BRANCH_WIRE_LIBRARIES_TS` companion is needed — no ABI delta.

## Ops notes

- Devnet/testnet depots need a `sysio.epoch` + `sysio.msgch` contract redeploy to pick this up.
- Any outpost already wedged by a previously-delivered `EMISSIONS_BLOCKED` envelope: `OutpostManager.blackholeAttestationTypes([EMISSIONS_BLOCKED])` remains the immediate unwedge (the wedged epoch's deliveries were fully unwound, so redelivery tips after the config call).


huangminghuang (commented):
Two low-priority follow-ups from the delegated code review; no blocking correctness findings.

contracts/tests/sysio.dispatch_tests.cpp
```diff
@@ -1520,6 +1520,19 @@ BOOST_FIXTURE_TEST_CASE(chkcons_survives_non_advancing_advance, sysio_dispatch_t
...
+// msgch::bootstrap refuses to start a chain whose emissions are not configured. This fixture
+// deploys no sysio.system at all (setemitcfg / initt5 never ran), which is exactly the
+// misconfigured-deployment shape: without the guard, bootstrap's inline genesis advance would
+// gate-block SILENTLY (the emissions gate never throws, by design) and the defect would surface
+// only as "epoch stuck at 0". advance()'s own soft block-and-retry behavior for the ECONOMIC gate
+// reasons is covered above (chkcons_survives_non_advancing_advance) and in emissions_tests.
+BOOST_FIXTURE_TEST_CASE(bootstrap_requires_emissions_config, sysio_dispatch_tester) { try {
+   bootstrap_for_dispatch();
+   BOOST_REQUIRE_EQUAL(
+      wasm_assert_msg("msgch::bootstrap: emissions config missing -- sysio.system::setemitcfg must run before bootstrap"),
```
huangminghuang, line 1548: The regression covers only the missing `emitcfg` branch. The new `t5state`-missing guard remains untested, so a wrong table/account lookup or accidentally shadowed second check could ship unnoticed. Please add a case where `emitcfg` exists but `t5state` does not and assert the `initt5` diagnostic; ideally also cover the fully configured 0 -> 1 bootstrap success path.

contracts/tests/emissions_tests.cpp
```diff
@@ -2299,9 +2299,9 @@ BOOST_FIXTURE_TEST_CASE( advance_gate_blocks_before_initt5, sysio_emissions_test
 
 BOOST_FIXTURE_TEST_CASE( gate_block_dedup_same_reason, sysio_emissions_tester ) try {
    // Two consecutive gate-block attempts with the same reason: blocklog row
-   // exists, retry_count increments, last_retry_at advances. No outbound
-   // queueout is attempted on the second call (reason unchanged), so the
-   // trx still succeeds even without sysio.msgch deployed.
+   // exists, retry_count increments, last_retry_at advances. The gate state
+   // is depot-local (no cross-chain broadcast), so the trx succeeds even
```
huangminghuang, line 2305: This updated comment correctly describes the depot-local behavior, but this file still has stale `EmissionsBlocked` broadcast descriptions at lines 11, 2196, and 3725-3726—including an explicit claim that an attestation is queued per outpost. Please update those so the tests do not document behavior this PR removes.

contracts/tests/sysio.dispatch_tests.cpp
heifner: Both added in b4f301d5aa.

To make the half-configured state reachable, `enable_epoch_advancement` is split into `deploy_system_with_emitcfg` (deploy sysio.system + T5 holding accounts + `setemitcfg`) and the `initt5` step on top, so the existing advancing tests are unchanged.

- `bootstrap_requires_emissions_state` — `emitcfg` present, `t5state` absent; asserts the `initt5` diagnostic. Guard 1 passes here, so a wrong table/account lookup or a shadowed copy of the first check surfaces as the wrong message rather than passing silently.
- `bootstrap_advances_genesis_epoch_when_emissions_configured` — fully configured 0 → 1: `bootstrap` succeeds, `current_epoch()` goes to 1, and the gate clears the blocklog row left behind by `bootstrap_for_dispatch`'s earlier gate-blocked advance. Without this one, guards that always threw would look identical to the two failure cases.

The three now sit under a shared banner; the "advance()'s economic reasons are covered elsewhere" note moved there instead of being repeated per case.

`contracts_unit_test`: 572 cases, no errors detected.

contracts/tests/emissions_tests.cpp
heifner: Fixed in b4f301d5aa — all three, plus one more in the same neighbourhood:

- line 11 (file header) — "enforcement via the gate's `EmissionsBlocked` path" → the gate's block path, recorded depot-locally in `sysio.epoch::blocklog`, nothing broadcast cross-chain.
- line 2196 — "bricks with `EmissionsBlocked` on the next advance" → bricks on the next advance; the gate sees treasury exhausted and refuses to advance.
- lines 3725-3726 — dropped the "an `EmissionsBlocked` attestation is queued per outpost" claim; now states the epoch index stays at 0 and a blocklog row is written, gate state depot-local.
- line 3744 (same test body, not in your list) — "no throw -- gate emits error attestation cleanly" → "the gate records the block and returns cleanly". Same stale claim, so it went with them.

`grep -n "EmissionsBlocked\|attestation" contracts/tests/emissions_tests.cpp` is now empty.

huangminghuang (approved):
Approved. The follow-up commit addresses both review comments, and the complete PR re-review found no remaining code issues. Non-blocking request before merge: please update the PR description Tests section to include `bootstrap_requires_emissions_state`, `bootstrap_advances_genesis_epoch_when_emissions_configured`, and the latest 572-case `contracts_unit_test` validation. The current description mentions only `bootstrap_requires_emissions_config`.

PR #537, opened 2026-07-27
SEC-139: Pin Solana RPC clients by genesis hash

huangminghuang:
## Summary

- add optional `--outpost-solana-cluster-identity <client-id>,<expected-genesis-hash>` configuration for a backward-compatible initial release
- preserve legacy unpinned behavior when no identities are configured; once any identity is present, require complete one-to-one coverage for every Solana client
- add a generic libfc HTTP once-per-new-connection validation hook and a generic JSON-RPC adapter; keep Solana genesis-hash policy in the Solana client
- validate every newly established persistent HTTP/TLS connection with `getGenesisHash` before ordinary RPC traffic; replacement and fallback connections validate independently
- make confirmed identity mismatches sticky, cross-check public `getGenesisHash` results before admitting queued calls, and honor both deadlines and cancellation while queued
- bind pinned signing to blockhashes obtained through this client's validated RPC connection, without adding network I/O to local signing
- submit transactions once and bound confirmation polling, including an in-flight status RPC, by the absolute retry deadline
- perform bounded all-or-nothing startup validation and expose bounded-cardinality Prometheus status, failure, cancellation, recovery, and blocked-operation telemetry

## Why

Before this change, a configured Solana RPC URL proved only that something answered JSON-RPC; nodeop had no trusted identity for the cluster behind that endpoint. A stale deployment setting, DNS or load-balancer misroute, or reconnect/fallback to a different backend could therefore move reads and transaction submission to the wrong Solana cluster without an explicit failure.

That failure can look healthy at every individual step: `getLatestBlockhash` returns a valid blockhash, local signing succeeds, submission returns a signature, and status polling returns structurally valid responses—but all of them describe the wrong ledger. A startup-only or process-wide check would still leave a gap because persistent HTTP connections are replaced over time and a replacement can reach a different backend.

This PR makes cluster identity part of connection admission. Every new persistent connection must report the independently configured genesis hash before it carries protected operations, blockhash provenance ties pinned signing to this validated client, and a confirmed mismatch fails closed. The companion tooling PR provisions and verifies the trusted hash instead of asking the runtime endpoint to attest to itself.

## Backward-compatible rollout

The initial release has two process-wide modes:

- with no identity entries, existing configurations remain usable in legacy unpinned mode; nodeop emits a security warning per client and Prometheus reports the missing identity
- once any identity entry is present, strict pinned mode requires exactly one valid canonical 32-byte genesis hash for every configured Solana client; partial, duplicate, unknown-client, and malformed configurations fail startup

There is no mixed pinned/unpinned mode. The companion tooling change in [wire-tools-ts PR #42](https://github.com/Wire-Network/wire-tools-ts/pull/42) / [SEC-142](https://wire-network.atlassian.net/browse/SEC-142) provisions trusted hashes for local validators, requires independently supplied hashes for external clusters, verifies external endpoints, persists the identity, and renders this node option.

## Architecture

[PR #535](https://github.com/Wire-Network/wire-sysio/pull/535) is merged. Its generic connection-affine continuation remains independent. This PR adds a generic connection-lifecycle validator in libfc HTTP; JSON-RPC only adapts method/parameters/result framing, while Solana owns the expected-genesis-hash predicate, sticky mismatch policy, blockhash provenance, and operation telemetry.

Validation state belongs to the live HTTP/TLS connection and is discarded when it closes. A replacement connection cannot inherit admission. The narrow Solana RPC gate makes a returned public `getGenesisHash` mismatch atomic with queued-call admission and observes the same deadline/cancellation contract as the blocking transport.

Jira: [SEC-139](https://wire-network.atlassian.net/browse/SEC-139)

## Validation

Current head: `1fe7776c6f4036379f43d5019a9c716f24a1f6ef`

- post-merge full Release build passed locally
- focused Release CTest suites passed: `test_fc`, `test_outpost_solana_client_plugin`, and `test_prometheus_plugin`
- combined JSON-RPC and Solana libfc suites passed five consecutive stress runs (124 cases per run)
- targeted regressions cover per-connection replacement validation, queued sticky mismatch, gate deadline/cancellation without network I/O, hanging confirmation RPC deadline enforcement, no transaction resubmission, startup rollback, and metrics
- `git diff --check` passed
- combined Release platform build and all 13 E2E flows passed: [run #30467121335](https://github.com/Wire-Network/wire-platform-build-system/actions/runs/30467121335)
  - `wire-sysio`: `1fe7776c6f4036379f43d5019a9c716f24a1f6ef`
  - companion `wire-tools-ts`: `39a832774f4a9310ec5207f3a3ef56bb839cc50f`
- no `contracts/sysio.*` paths are changed, so the system-contract E2E gate does not apply

## Residual risk

- legacy unpinned mode deliberately preserves compatibility but does not provide SEC-139 protection; warnings and metrics make that state observable during migration
- genesis-hash comparison detects accidental/wrong-cluster routing but does not cryptographically authenticate a malicious RPC endpoint; trusted provisioning plus TLS/trusted routing remain required
- a confirmed mismatch is intentionally sticky and requires endpoint/configuration correction plus process restart

[SEC-142]: https://wire-network.atlassian.net/browse/SEC-142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/outbound-http-transport.md
+++ b/docs/outbound-http-transport.md
@@ -55,6 +55,24 @@ closed before reuse, which keeps the reuse window below common provider keep-ali
 peer-closed sockets retained by the process. A non-blocking socket peek also rejects a cached connection when a
 peer FIN/reset or unexpected unread bytes are already observable.
 
+## Per-connection validation
+
+The asynchronous client and blocking adapter accept an optional generic validator for fixed-endpoint clients. Before
+the first ordinary request on each newly established connection, the transport sends one bounded, single-attempt
+validation request to that same endpoint. The response predicate either admits the connection or throws to reject
+and close it. An admitted connection must remain persistent; if the peer closes the validation response, the
+transport rejects it because no subsequent operation can be bound to that observation.
+
+Validation state belongs to the live HTTP/TLS connection and is discarded when it closes. A replacement or fallback
+connection therefore validates independently before carrying ordinary work. The attempt, success, and failure
+lifecycle callbacks are non-blocking telemetry hooks; they cannot change the single-attempt policy and must not
+throw from completion callbacks. The initiating ordinary request's already-started total deadline caps the
+validation deadline, so validation cannot extend that request's overall budget.
+
+For a connection-affine continuation, validation precedes the initial call. The hook and follow-up then use the same
+already-admitted connection; the continuation mechanism does not trigger another validation between its two legs.
+This keeps endpoint admission generic and independent from higher-level JSON-RPC or chain-specific policy.
+
 ## Connection-affine continuations
 
 The asynchronous client and blocking compatibility adapter expose a generic one-step continuation hook. The
--- a/docs/solana-cluster-identity.md
+++ b/docs/solana-cluster-identity.md
@@ -33,7 +33,6 @@ outpost-solana-cluster-identity = <client-id>,<expected-genesis-hash>
 Optional policy:
 
 ```ini
-outpost-solana-cluster-identity-max-age-ms = 30000
 outpost-solana-cluster-identity-probe-timeout-ms = 5000
 ```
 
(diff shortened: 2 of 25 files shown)


heifner (dismissed):
The rework addresses the full review: cluster verification is now a generic per-connection admission check in the shared transport (reusable for the Ethereum client's chainId check), the per-call preflight and its request amplification are gone, signing is local again with blockhash provenance enforcement, non-persistent endpoints are rejected up front with a clear error, and the identity telemetry now reports protected-operation failures, classifies cancellations separately, and derives a monotonic clamped verification age.

Non-blocking notes: pinned signing rejects externally sourced blockhashes by design, so durable-nonce or cross-client construction would need a follow-up if those flows arrive; lock_rpc still degrades to an untimed wait for a libfc embedder that clears the total timeout with no deadline or cancel hook; validate_returned_genesis_hash carries a vestigial rethrow-only catch; and the tight timing margins in the new concurrency tests are worth watching in CI.

PR #538, opened 2026-07-28
Packaging: wire-cdt release-asset consumption, macOS + system-contracts artifacts, Strategy-C release workflows

jglanz:
## Summary

Part 2 of the packaging effort (pairs with wire-cdt PR #100 — same merge window required: this PR consumes the renamed `wire-cdt_*` deb).

**CDT acquisition**: tag builds install the wire-cdt deb from a **pinned GitHub release asset** (`.cicd/defaults.json` → `wirecdt.release`/`wirecdt.channel`, written by `prepare-release.yaml` at version-bump time — resolve-at-prepare, pin-at-build, reproducible re-runs). The ~60-line CI artifact-resolution machinery and `wirecdt.target`/`prerelease` keys are gone; `gh` CLI baked into all five platform images. Install script asserts the `/usr/lib/cdt` home, `/usr/bin` entry-point symlinks, and rejects bare llvm/clang names. `nodeop`/`clio`/`kiod` stay directly in `/usr/bin` (unchanged, now explicitly asserted in verify-deb/rpm).

**Artifacts (8 per release)**: existing 2 deb + 2 rpm + linux tgz, plus `wire-sysio-<v>-macos-arm64.tar.gz` (service payload gated `if(NOT APPLE)`; `WIRE_ARCH_TAG` keeps Linux filenames byte-identical; verify-tgz `--no-service` mode) and `wire-sysio-system-contracts-<v>.tar.gz` (tag builds, after the contract-build assertion) + checksums.

**CI**: vcpkg binary cache monolithic→split restore/save (save-immediately-after-build) in both build workflows; explicit per-artifact tgz verification (no glob-order hazards).

**Release (Strategy C)**: `prepare-release.yaml` (sole `version` input; channel from suffix; pins the latest published wire-cdt release for the channel) → `tag-release.yaml` (`environment: release` approval, tag-exists guard, draft-first, dispatches both tag builds) → `release.yaml` (dispatch-path TAG fix, head_branch run resolver, awaits linux+macOS, 8-asset publish, draft flip). Flow diagram in `docs/release-workflow.md`.

**Evidence**: linux run 30308008876 (5-leg matrix + platform-image rebuilds with gh) and macOS run 30308010424 green, artifacts `wire-sysio-packages-{amd64,macos-arm64}` correctly named.

### DIFF
--- a/.cicd/defaults.json
+++ b/.cicd/defaults.json
@@ -1,9 +1,9 @@
 {
-    "wirecdt":{
-       "target":"master",
-       "prerelease":true
-    },
-    "wiresystemcontracts":{
-       "ref":"release/4.0"
-    }
- }
+  "wirecdt": {
+    "release": "v1.0.0-dev",
+    "channel": "prerelease"
+  },
+  "wiresystemcontracts": {
+    "ref": "release/4.0"
+  }
+}
--- a/.cicd/defaults.json
+++ b/.cicd/defaults.json
@@ -2,8 +2,5 @@
   "wirecdt": {
     "release": "v1.0.0-dev",
     "channel": "prerelease"
-  },
-  "wiresystemcontracts": {
-    "ref": "release/4.0"
   }
 }
--- a/.cicd/platforms/asan.Dockerfile
+++ b/.cicd/platforms/asan.Dockerfile
@@ -76,3 +76,15 @@ EOF
 
 ENV SYSIO_PLATFORM_HAS_EXTRAS_CMAKE=1
 ENV ASAN_OPTIONS=detect_leaks=0
+
+# GitHub CLI from the official apt repository. The release workflows use `gh`
+# inside these images (linux_amd64_build.yaml downloads the pinned wire-cdt
+# release asset with `gh release download`), and it is installed in EVERY
+# platform image rather than just the packaging one so the images stay uniform.
+RUN mkdir -p -m 755 /etc/apt/keyrings \
+    && wget -nv -O /etc/apt/keyrings/githubcli-archive-keyring.gpg https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+    && chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
+    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+       > /etc/apt/sources.list.d/github-cli.list \
+    && apt-get update \
+    && apt-get install -y gh
--- a/.cicd/platforms/asserton.Dockerfile
+++ b/.cicd/platforms/asserton.Dockerfile
@@ -79,3 +79,15 @@ COPY <<-EOF /extras.cmake
 EOF
 
 ENV SYSIO_PLATFORM_HAS_EXTRAS_CMAKE=1
+
+# GitHub CLI from the official apt repository. The release workflows use `gh`
+# inside these images (linux_amd64_build.yaml downloads the pinned wire-cdt
+# release asset with `gh release download`), and it is installed in EVERY
+# platform image rather than just the packaging one so the images stay uniform.
+RUN mkdir -p -m 755 /etc/apt/keyrings \
+    && wget -nv -O /etc/apt/keyrings/githubcli-archive-keyring.gpg https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+    && chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
+    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+       > /etc/apt/sources.list.d/github-cli.list \
+    && apt-get update \
+    && apt-get install -y gh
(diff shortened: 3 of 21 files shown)


heifner (commented):
Reviewed by running this PR's own gates against the artifacts from the two runs cited in the description (`wire-sysio-packages-amd64` from 30308008876 and `wire-sysio-packages-macos-arm64` from 30308010424), plus synthetic fixtures for the paths a branch build cannot reach.

## Verified working

- **Both `verify-tgz.sh` modes pass on the real tarballs**, and the new `--no-service` mode is not vacuous: a crafted tarball carrying a systemd unit fails with `service file present in a no-service tarball`, and one missing `bin/clio` fails with `missing wire-sysio/bin/clio`.
- **The `if(NOT APPLE)` gating holds in the shipped artifacts** — the macOS tarball contains no systemd / tmpfiles.d / logrotate entries, the Linux tarball contains all three.
- **`WIRE_ARCH_TAG` does what it claims**: the Linux artifact names are unchanged (`wire-sysio-1.0.0-dev-x86_64.tar.gz`, `wire-sysio_1.0.0-dev_amd64.deb`), and macOS is qualified as `-macos-arm64`.
- **`verify-deb.sh` passes on the real deb pair**, S6 included (the container enabled the unit and nodeop produced blocks).
- **The `$6` change in `install-wire-cdt-package.sh` fixes a real bug on master.** `dpkg-deb --contents` renders a symlink as `./usr/bin/cdt-cc -> ../lib/cdt/bin/cdt-cc`, so the old `$NF` resolver read the link TARGET rather than the path. I exercised all six new payload assertions against a synthetic-deb matrix (missing `cdt.imports`, home at `/opt/wire-cdt`, missing discoverable config, a `/usr/cdt` subtree, `/usr/bin/clang`, bare `llvm-*`) — each fails with an accurate message — and the happy path installs in a container with resolving symlinks and `CDT_ROOT` exported.
- **`cmake-version.py` round-trips** the real `CMakeLists.txt`: `""` for the stable case, the blank line after the block preserved, no lines but the `VERSION_*` ones touched.
- **The `wiresystemcontracts` removal is complete** — no consumers remain in the tree.

## The release job can give up before the build it waits for is allowed to finish

`linux_amd64_build.yaml` sets `timeout-minutes: 360`. `release.yaml` sets `350` and polls `{1..330}` with `sleep 60` — 330 minutes of sleep alone, plus two `gh api` calls per iteration — and only after that does it download ~850 MB of artifacts, run S1-S6 (which builds a container and boots nodeop), upload eight assets and flip the draft. A linux build that lands anywhere in the 330-360 minute band therefore fails the release job with `No successful linux_amd64_build run` (or gets killed at 350) while the build itself goes on to succeed, leaving the release a draft with no assets and no obvious cause.

The step comment asserts the opposite ("waits out a full build ... linux_amd64_build allows itself 360 minutes"), so the intent is clear; it is the arithmetic that does not hold. Raising `timeout-minutes` alone cannot close it, since 350 already sits against GitHub's 360-minute ceiling for hosted jobs and the post-poll work needs real time. Options: lower the build's own budget to something the release job can outlast, shorten the poll and lean on re-dispatch for the tail, or drive the asset job from `workflow_run` completion instead of polling.

## The pinned wire-cdt release does not exist yet

`.cicd/defaults.json` pins `wirecdt.release: v1.0.0-dev`, but the published releases on Wire-Network/wire-cdt currently stop at `v4.2.0-rc1`. The first tag build after this merges fails at *Download Wire CDT release package for release tag* until wire-cdt#100 lands **and** its `v1.0.0-dev` release is *published* — a draft is explicitly rejected by that same step, and wire-cdt's `tag-release.yaml` creates releases as drafts. Branch and master builds are unaffected because the step is tag-gated, which is why CI is green here.

Not a code defect, and the `not_published_hint` message is genuinely good. But it is a hard ordering prerequisite that is currently implicit; worth stating in the description and in `docs/release-workflow.md` so whoever cuts the first tag knows the wire-cdt release must be published first.

## The CDT payload gate does not assert the archiver

`install-wire-cdt-package.sh` is now the place where CDT package shape is asserted — home, entry points, banned bare toolchain names, discoverable config, `cdt.imports`. It does not assert `bin/cdt-ar` or `bin/cdt-ranlib`, and the wire-cdt package currently omits both while shipping a `CDTWasmToolchain.cmake` that sets `CMAKE_AR`/`CMAKE_RANLIB` to exactly those paths (raised on wire-cdt#100, where a static-library build under the packaged toolchain fails at the archive step). I confirmed this gate accepts a synthetic deb with no `cdt-ar`. Two more entries in the post-install `required` list would close the loop between the two PRs regardless of which one fixes the packaging side.

## Minor

**`verify-rpm.sh` fails silently when `rpm` is absent.** `l=$(rpm -qlp "$r" 2>/dev/null)` fails under `set -e` before any `fail` can print, so the script exits 127 with no output at all. CI runs it where `rpm` exists, so this is local-developer friction only — but the PR touches this file, and a `command -v rpm` guard with a real message is a one-liner.

**`cmake-version.py` reads leniently where it writes strictly.** Given `set(VERSION_SUFFIX dev) # comment`, `read` silently reports `1.3.0` (suffix dropped) while `write` correctly refuses with `could not rewrite`. `tag-release.yaml` turns the lenient read into a loud version mismatch, so nothing breaks quietly today, but a strict read would keep the two halves of that assertion honest.

## Coverage note

The runs cited as evidence are branch builds, so the system-contracts bundle, the eight-asset upload, the checksums file and the draft flip are the parts still unexercised end to end — worth a dry run on a throwaway `-rc` tag before the first real release, once the wire-cdt pin resolves. I also could not run the S5/S6 rpm gate locally (no `rpm` on this machine); the deb equivalent passed.

PR #547, opened 2026-08-06
Test: compare the renamed pre-shutdown snapshot and pin the head

heifner:
`nodeop_chainbase_allocation_test` failed in the ubsan job of run 31047433131 with `FileNotFoundError` rather than its own assertion message.

The comparison read the pre-rename path, which no longer exists once the snapshot is moved aside. When both snapshots land on the same head the two names are identical, so `filecmp` compared the post-restart snapshot with itself and the `snapshot is not identical` assertion could never fail. When the heads differ the path is gone and the test dies inside `filecmp`.

The heads differ whenever the irreversible-mode node is still short of the producer's final LIB. Killing the sole producer can leave its last block undelivered, and the fixed three second sleep was not a guarantee that the node had caught up. In the failing run the snapshot landed one block short of the schedule promotion, so `global_property_object` was not exercised either. The test now waits on the node's own LIB, then takes down the remaining peers so nothing can advance it between the two snapshots.

It also compares with `shallow=False` so matching stat signatures cannot stand in for a byte comparison, and reports a mismatched head directly instead of letting it surface as a confusing byte difference.

Exercised locally across runs that landed on different heads, which is the variability that broke the previous version.

### DIFF
--- a/tests/nodeop_chainbase_allocation_test.py
+++ b/tests/nodeop_chainbase_allocation_test.py
@@ -2,7 +2,6 @@
 
 import signal
 import json
-import time
 import os
 import filecmp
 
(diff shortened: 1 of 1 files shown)


huangminghuang (approved):
Reviewed at 367da668d45bae4a6ab65d379598751dafd635cc. No actionable findings.

PR #549, opened 2026-08-06
feat(chalg): underwriter-fault challenge — bonded, council-adjudicated slashing (WIRE-297)

heifner:
Security property #4 ("the bond is slashable") was unenforced: no production path could slash an UNDERWRITER-type operator, and the collateral-lock window ended in an unconditional release. This adds the missing trigger, mirroring the batch-operator envelope dispute everywhere the problem allows.

## sysio.chalg — the envelope dispute's sibling

- **`openuwchal(challenger, uwreq_id, underwriter, reason, detail)`** — permissionless filing against a CONFIRMED commitment, time-gated to its collateral-lock window. The challenger escrows a WIRE bond = the WIRE value of the challenged locks, each leg priced through its OWN reserve's live books (`token_to_wire`; u128-summed). One challenge EVER per (uwreq, underwriter) — a council verdict is final per commitment; other commitments are independently challengeable. The free-text `detail` is capped at 1 KiB, refused before any escrow moves.
- **`uwchalbond`** — read-only quote sharing ONE formula helper with the charge, so quote and escrow can never drift (the `split_wire_fee` principle).
- **`voteuwchal`** — record-only Tier-1 ballot (electorate + quorum snapshotted at open via `snapshot_t1_electorate`, now shared with `opendispute`), refused once `deadline_ms` passes so no late quorum can be assembled and hand-cranked into a verdict — including a slash — after the documented hard window. Three-way: UPHOLD / REJECT_REFUND / REJECT_FORFEIT — forfeiture to the wrongly-challenged underwriter happens ONLY by an explicit majority of the rejectors (tie favours refund); an honest mistake refunds.
- **`chkuwchal`** — permissionless tally, `chkdispute`-shaped. UPHELD: `slashop`, then `uwrit::sweeplocks` (inline actions run depth-first in send order, so `releaselock` takes its deferred-slash branch — locked collateral debited, outbound SLASH attestations queued). REJECTED: `freelocks`; the locks live out their window. Past the deadline with no quorum: LAPSED — bond credited back, nobody punished. Deliberately NO after-deadline relaxed tally (a challenge must be voted inside the window) and NO epoch pause (an unresolved challenge is survivable; the chain keeps advancing). Resolution also compacts the row to a fixed-width tombstone — `detail` and `electorate` cleared, the scoped ballots erased — so retained variable-length state is bounded by CONCURRENTLY OPEN challenges (each backed by a live lock set and an escrowed bond) rather than by every challenge ever filed. The per-note cap alone cannot do that: filing is permissionless and the bond comes back on every non-forfeit outcome. The full filing and every ballot stay permanently readable in the action-trace history.
- **`claimbond` / `bondcredits` (ABI addition)** — the escrow leaves a resolved challenge as a CREDIT, pulled later under the recipient's own authority (the `sysio.dclaim::claim` shape, erase-before-send). `chkuwchal` can run inline under the epoch tick, and `sysio::token::transfer` executes the recipient's code through `require_recipient(to)`: a challenger contract that allowed its outgoing escrow and asserted on the incoming refund would abort the whole `advance`, and the rollback leaves the challenge OPEN with its locks still held, so every later sweep re-pokes the same failing transfer — epoch advancement stalls chain-wide. Crediting keeps every account the crank touches system-owned; the only transaction a hostile recipient can abort is its own claim, stranding only its own funds. Credits accumulate per account and saturate at `asset::max_amount` (the `add_wire_capped` discipline).

## sysio.uwrit — the lock hold

- **`lock_entry.challenge_id` (ABI change)**: a held lock is NOT released at expiry; `chklocks` skips it and pokes `chkuwchal` instead — the epoch tick is the challenge system's only cadence (possible precisely because the chain is not paused; envelope disputes need a standalone crank because theirs is).
- **`holdlocks` / `freelocks` / `sweeplocks` (auth `sysio.chalg`)**: hold is the authoritative liveness gate for a filing (a stale open aborts whole, escrow included); free/sweep are soft past auth — they run inline from the epoch tick, where an abort would stall advancement chain-wide.
- The COMPLETED-flip / evidence-clear / retention tail is extracted to `finalize_settled_uwreqs`, shared by `chklocks` and `sweeplocks`.

The evidence-retention prerequisite flagged on WIRE-297 needed no work: CONFIRMED uwreqs carry `expires_at_epoch == 0` and `pruneuwreqs` skips them, so the stored SwapRequest / source_tx_id / UIC bytes already survive the whole lock window; the deadline is only stamped when the locks leave.

## Tests

`sysio.dispatch_tests` (reusing its CONFIRMED-uwreq machinery): escrow/hold/uniqueness; every filing guard, incl. the detail cap and whole-transaction rollback; ballot guards, incl. the door closed at the deadline; uphold end-to-end (SLASHED, locked collateral debited via the deferred-slash branch, bond credited then claimed, uwreq COMPLETED, chalg custody zero once pulled); reject-forfeit (credit to the underwriter, collateral byte-identical after the healthy release); reject-tie → refund; the resolved-row compaction and ballot erasure; `claimbond` guards; the epoch-tick skip/poke/lapse/second-sweep sequence; and chalg-only auth on the lock trio.

One case pins the pull payout specifically: a challenger that deploys `reject_all.wasm` after filing. The epoch tick still resolves the challenge, sweeps the locks and settles the uwreq, and only that account's own claim fails. Against the previous push payout it fails with "rejecting all notifications" raised out of `chklocks`.

ABI delta: chalg + uwrit only; other wasm churn is ABI-identical rebuilds from the full build.

The new `flow-underwriter-slashing` exercises both verdicts against a live cluster: REJECTED_FORFEIT credits the wrongly-challenged underwriter, UPHELD slashes the winner and sweeps the locks, and each recipient pulls its payout through `claimbond`, leaving `sysio.chalg` at zero custody.

**Companion PRs (merge this first):** Wire-Network/wire-libraries-ts#61 (regenerated `SysioContractTypes`) and Wire-Network/wire-tools-ts#56 (`flow-underwriter-slashing`). Merge order: sysio → libraries → tools.


contracts/sysio.chalg/src/sysio.chalg.cpp
```diff
@@ -227,4 +307,239 @@ void chalg::chkdispute(uint64_t dispute_id) {
...
+// ---------------------------------------------------------------------------
+void chalg::voteuwchal(name owner, uint64_t chal_id, uint8_t ballot) {
+   require_auth(owner);
+
+   const auto cast_ballot = magic_enum::enum_cast<uwchal_ballot>(ballot);
+   check(cast_ballot.has_value(), "voteuwchal: unknown ballot value");
+
+   uwchals_t chals(get_self());
+   auto c = chals.get(uwchal_key{chal_id}, "challenge not found");
+   check(c.status == DisputeStatus::DISPUTE_STATUS_OPEN, "challenge is not open");
```
huangminghuang, line 478: **[HIGH] Enforce the challenge deadline when accepting ballots.** `voteuwchal` only checks that the row is OPEN, while `chkuwchal` evaluates quorum before its lapse branch. Between `deadline_ms` and the next `chklocks` sweep, voters can therefore submit a full quorum and manually crank an UPHELD/REJECTED verdict—including slashing after the documented hard window. Reject ballots when `current_time_ms() >= c.deadline_ms`, and add a regression that advances past expiry, rejects the late vote, then verifies `chkuwchal` resolves LAPSED.

contracts/sysio.uwrit/include/sysio.uwrit/sysio.uwrit.hpp
```diff
@@ -459,7 +502,7 @@ namespace sysio {
 
          SYSLIB_SERIALIZE(lock_entry,
             (lock_id)(uwreq_id)(underwriter)(chain_code)(token_code)(reserve_code)
-            (amount)(created_at_ms)(expires_at_ms))
+            (amount)(created_at_ms)(expires_at_ms)(challenge_id))
```
huangminghuang, line 530: **[HIGH] Preserve decoding of locks written by the previous contract.** Appending a mandatory `uint64_t` makes the new `SYSLIB_SERIALIZE` decoder read eight bytes past every pre-upgrade row, which ends at `expires_at_ms`; CDT’s datastream aborts on that underflow. Because `chklocks` runs inside `sysio.epoch::advance` and `opreg::available()` also reads this table, deploying with any live lock can stall epoch advancement. Use a trailing `binary_extension<uint64_t>` interpreted as zero, a separate hold table, or an explicit migration/empty-table deployment gate, and cover an old-format row in an upgrade test.

contracts/sysio.chalg/src/sysio.chalg.cpp
```diff
@@ -227,4 +307,239 @@ void chalg::chkdispute(uint64_t dispute_id) {
...
+   const auto now     = current_time_point();
+   uint64_t   next_id = std::max<uint64_t>(1, chals.available_primary_key());
+
+   chals.emplace(ram_payer, uwchal_key{next_id}, uwchal_entry{
+      .id          = next_id,
+      .uwreq_id    = uwreq_id,
+      .underwriter = underwriter,
+      .challenger  = challenger,
+      .reason      = *fault,
+      .detail      = std::move(detail),
```
huangminghuang, line 445: **[MEDIUM] Bound permissionless, permanently retained `detail`.** The caller controls this string, the `uwchals` audit row is retained indefinitely, and RAM is billed to `sysio`. The challenge bond does not price storage and is returned on lapse, so capital can be recycled to persist near-maximum transaction payloads across confirmed requests. Enforce a named byte limit before escrow/insertion and define pruning or challenger-paid storage for retained records.

huangminghuang (commented):
Three actionable findings from the delegated code review.

contracts/sysio.chalg/src/sysio.chalg.cpp
heifner: Fixed in b1c5cd621c: `voteuwchal` now refuses ballots once `current_time_ms() >= deadline_ms` ("voteuwchal: the challenge window has expired"), so a late quorum can no longer be assembled for a manual `chkuwchal` crank in the expiry→sweep gap. The regression rides the existing lapse test (`chklocks_skips_held_lock_and_lapse_refunds`): it now advances past expiry, asserts the late ballot is rejected, then verifies the sweep poke resolves LAPSED. `contracts_unit_test`: No errors detected.

contracts/sysio.chalg/src/sysio.chalg.cpp
heifner: Fixed in b1c5cd621c: `openuwchal` now enforces `detail.size() <= max_uwchal_detail_bytes` (a named 1 KiB constant next to `dispute_deadline_sec`), refusing the filing before any escrow moves. That bounds the retained-payload amplification the recycled bond would otherwise fund; `openuwchal_guards` covers the over-limit rejection. Pruning / challenger-paid storage for the retained audit rows is a real design question, but the retention itself is deliberate (a verdict is final per commitment — the row IS the uniqueness gate), so I'd rather take that up as a follow-up than bolt it on here.

contracts/sysio.uwrit/include/sysio.uwrit/sysio.uwrit.hpp
heifner: Declining this one deliberately, per the repo's standing pre-release policy (`no-back-compat-before-release`): until 2026-09-09 the software is unreleased — there are no persistent deployments, and every cluster (local harness and the e2e gate alike) bootstraps fresh from this tree via setcode onto empty state, so a pre-upgrade lock row does not exist anywhere for the new decoder to meet. A `binary_extension` / dual-read shim would be back-compat support for a shape that was never deployed, which that policy bans outright.

The underlying concern is real for any LIVE-chain upgrade path, and it generalizes well past this field (any table-shape change has the same property). When the release posture is reassessed on 2026-09-09, contract-upgrade migration discipline — extension fields, migration actions, or empty-table deployment gates — becomes a standing requirement; this row is a good first entry for that checklist.

contracts/sysio.chalg/src/sysio.chalg.cpp
```diff
@@ -227,4 +307,247 @@ void chalg::chkdispute(uint64_t dispute_id) {
...
+         std::make_tuple(c.uwreq_id, c.underwriter)
+      ).send();
+   }
+
+   // Route the bond per the verdict: forfeiture to the wrongly-challenged underwriter ONLY on an
+   // explicit REJECT_FORFEIT majority; every other outcome returns it to the challenger.
+   if (c.bond_amount > 0) {
+      const bool forfeited = (verdict == uwchal_verdict::REJECTED_FORFEIT);
+      const name bond_recipient = forfeited ? c.underwriter : c.challenger;
+      action(
```
huangminghuang: **[HIGH] Decouple bond payout from epoch advancement.** This transfer can run under `epoch::advance -> chklocks -> chkuwchal`. `sysio.token::transfer` calls `require_recipient(to)`, so a permissionless challenger contract can allow its outgoing escrow but reject the later incoming refund. The rejection rolls back the verdict and `freelocks`; every later expiry sweep retries the same failing notification, letting the challenger hold epoch advancement hostage. Record the payout as a claimable balance and let the recipient withdraw in a separate action (or otherwise make delivery failure unable to abort the crank), and cover a recipient that rejects incoming transfers.

contracts/sysio.chalg/include/sysio.chalg/sysio.chalg.hpp
```diff
@@ -153,12 +263,83 @@ namespace sysio {
...
+      /// Challenge primary key (auto-incrementing id).
+      struct uwchal_key {
+         uint64_t id;
+         uint64_t primary_key() const { return id; }
+         SYSLIB_SERIALIZE(uwchal_key, (id))
+      };
+
+      /// An underwriter-fault challenge. Opened permissionlessly against one CONFIRMED
+      /// commitment; resolved by a Tier-1 vote or lapsed at the lock window's end. The row is
+      /// retained after resolution as the audit record AND as the guard that a commitment is
```
huangminghuang: **[MEDIUM] The 1 KiB cap does not bound cumulative retained state.** Every permissionless challenge still permanently keeps the `uwchals` row (including electorate/detail) plus its scoped ballot rows, while the bond is returned on lapse and can be recycled across confirmed requests. The uniqueness requirement only needs a compact `(uwreq, underwriter)` tombstone. Compact terminal rows and erase votes, or price/charge durable storage to the challenger.

huangminghuang (changes requested):
Requesting changes for the epoch-stall vulnerability in the bond payout path. The follow-up bounds each detail note, but cumulative terminal challenge and vote state also remains unbounded.

contracts/sysio.chalg/src/sysio.chalg.cpp
heifner: Fixed in 987a0ec883: the bond payout is now a pull.

`chkuwchal` credits a new `bondcredits` row (keyed by recipient, accumulating, saturating at `asset::max_amount`) and a new `claimbond(account)` action pays it out — `require_auth(account)`, erase-before-send, the `sysio.dclaim::claim` shape this repo already uses for exactly this. Every account the crank touches is now system-owned, so nothing in the `advance -> chklocks -> chkuwchal` tree can execute recipient-controlled code; the only transaction a hostile recipient can abort is its own claim, and only its own funds are stranded.

I checked the rest of the resolution path for the same lever while I was in there: `opreg::slash` only debits balances and emits attestations, and the UPHELD sweep's `releaselock` calls take the SLASHED branch, which also only debits and attests. The bond transfer was the last user-controlled abort in that tree.

Regression: `hostile_bond_recipient_cannot_stall_the_epoch_tick` files a challenge, then deploys `reject_all.wasm` at the challenger (nothing prevents code landing after the escrow), then drives the epoch tick. It asserts the sweep completes, the challenge LAPSES, the locks free and release on the next sweep, the uwreq settles — and that the challenger's own `claimbond` is the only thing that fails. I verified it's a real regression test rather than a passing one: against the previous push payout it fails with `assertion failure with message: rejecting all notifications` raised out of `chklocks`.

Merge-order note: the ABI gained an action and a table, so the wire-libraries-ts companion needs a regenerate, and the wire-tools-ts flow's "chalg ends at zero custody" assertion now needs an explicit `claimbond` step.

Separately — `opreg::releaselock` has an untouched sibling of this bug: its TERMINATED + WIRE-direct branch pushes an inline `token::transfer` to the operator account, also from inside `chklocks`. Pre-existing on master, not introduced here, and fixing it changes remit semantics, so I filed WIRE-327 rather than fold it in.

contracts/sysio.chalg/include/sysio.chalg/sysio.chalg.hpp
heifner: Fixed in 987a0ec883: `chkuwchal` now compacts the row on resolution and erases the scope.

`detail` and `electorate` are cleared and every ballot row for the challenge is erased, leaving the compact tombstone you described — the `byuwrequw` uniqueness gate plus the fixed-width verdict record. You're right that the 1 KiB cap only bounded one note: with permissionless filing and a bond that returns on every non-forfeit outcome, the total was still unbounded. Compaction is what moves the bound — retained variable-length state is now proportional to *concurrently open* challenges, each needing a live lock set and an escrowed bond, rather than to every challenge ever filed.

I kept the tombstone rather than pricing storage to the challenger: it is the uniqueness gate, so it has to outlive the challenge, and its count is bounded by the number of real CONFIRMED uwreqs — each costing a genuine swap to manufacture, and each challengeable once. The full filing and every ballot stay permanently readable in the action-trace history indexers consume.

Covered by `chkuwchal_compacts_resolved_row_and_erases_ballots`: a full-size note at the cap in; empty `detail`/`electorate` and null ballot rows out; and the tombstone still refusing a re-file after the reject freed the locks.

huangminghuang (dismissed):
Approved on head `987a0ec883`; the prior code findings are resolved. Non-blocking follow-up before merge: please refresh the PR description to reflect pull-based `claimbond` / `bondcredits`, terminal challenge and ballot compaction, the updated validation count, and the required companion SDK/flow changes. It still describes inline bond payouts and immediate zero custody.

huangminghuang (commented):
One blocking bond-range issue found.

contracts/sysio.chalg/src/sysio.chalg.cpp
```diff
@@ -1,19 +1,132 @@
...
+                                                        rit->reserve_wire_amount,
+                                                        rit->connector_weight_bps,
+                                                        it->amount);
+      if (leg_wire == 0) return uwchal_bond_quote{};
+      total += leg_wire;
+      quote.deadline_ms = it->expires_at_ms;
+      ++quote.live_locks;
+   }
+   if (quote.live_locks == 0) return uwchal_bond_quote{};
+   if (total > std::numeric_limits<uint64_t>::max()) return uwchal_bond_quote{};
```
huangminghuang: `total` is only bounded by `uint64_t`, but `openuwchal` immediately converts this result into `asset(static_cast<int64_t>(quote.bond), WIRE_SYMBOL)`, whose maximum is `asset::max_amount` (2^62-1). A two-leg request can produce a summed WIRE quote above that lower limit even while each lock/collateral bucket is valid (in particular, source and destination may use the same imbalanced reserve, so the two independent `token_to_wire` quotes can exceed one asset-sized WIRE balance). `uwchalbond` then advertises a nonzero quote, but filing fails before the locks are held, leaving that commitment unchallengeable. Bound the computed bond to the transferable asset range (or redesign escrow/claims to represent larger bonds) and add the boundary regression test.

huangminghuang (approved):
Re-review complete. The prior P1 is addressed; no further actionable findings.

PR #550, opened 2026-08-07
security: settle swaps on the AMM quote; gate operator prune on settlement (WNS-02, WNS-01)

heifner:
Closes the two **Critical** findings from CertiK "Wire Network - Sysio Audit 1" — WIRE-316 (WNS-02) and WIRE-317 (WNS-01). Both were verified still present on `master`.

## WNS-02 / WIRE-316 — a swap settled on the caller's `target_amount`

`createuwreq` stored `dst_amount = sr.target_amount` — a number the caller puts in the SwapRequest, which crosses the OPP boundary unauthenticated — and `sysio.reserv::applyswap` / `applyfromwire` / `paywire` paid it out verbatim.

The only guard computed its allowance from that **same** caller-supplied target:

```cpp
allowed = target * sr.target_tolerance_bps / 10000;   // scales with what it constrains
```

At `tolerance_bps = 10000` this reduces to `|quote - target| <= target`, which every target above the quote satisfies. A negligible source deposit could name an arbitrary destination amount and drain the reserve.

**Fix**

- `dst_amount` is the AMM quote — in `createuwreq` and `drainfwq` alike.
- The allowance derives from the **quote** via a shared `variance_allowance()`, with `tolerance_bps` clamped to 100%.
- `try_select_winner` re-prices on the live curve and settles on *that*. The ingestion quote alone would still be stale by up to the caller-supplied (previously unbounded) tolerance, and the books move in this transaction. This moves the quote/variance step **ahead of the bond check**, since the bond must cover what the winner will actually deliver.
- `sysio.reserv` enforces its own floor: the destination debit may not exceed the curve's output for the post-fee WIRE it receives. On the live path equality holds (same kernel, same pre-mutation rows, same transaction), so these are unreachable defense-in-depth — but the audited bug *was* a caller-chosen amount paid out verbatim, and a reserve that trusts its caller for a payout size has no floor of its own.
- The zero-quote fail-closed gate is hoisted out of its `target_amount != 0` nesting so its correctness no longer depends on an unrelated upstream guard.

### Behavior changes worth a reviewer's eye

1. ~~Variance rejection can now fire for an under-bonded candidate.~~ **Reverted in review** — the terminal verdicts now run after the bond check. See review comment 3.
2. **An unpriceable request no longer reaches the bond check at all** (it returns early), which is why `swap_same_token_legs_overcommit_is_disqualified` now needs ACTIVE reserves.
3. **The to-WIRE `asset::max_amount` guard is now unreachable.** An oversized target is refused at ingress by the variance check instead. The guard stays as defense-in-depth; its test was rewritten to assert the (stronger) ingress refusal.

## WNS-01 / WIRE-317 — `prune` stranded terminated operators' collateral

`terminate_inline` remits only the immediately-unlocked portion and deliberately leaves the locked remainder on the row, for `releaselock` to settle when `chklocks` frees each lock. `prune` erased on `terminated_at` + the prune delay alone, and `releaselock` returns at its missing-operator check — so the retained collateral was stranded with no attestation that could ever release it. `prune` is permissionless, so any caller could trigger it.

`prune` now additionally requires `is_fully_settled()`: every balance bucket drained **and** no `sysio.uwrit` lock still naming the account. Balances are checked by value, not `balances.empty()` — `subtract_balance` zeroes a bucket in place. The magic `20` bound became `MAX_OPERATOR_PRUNE_PER_CRANK`.

A row that never settles is never pruned, and that is the intended outcome: an unprunable row costs some RAM, an erased one costs the operator their bond. The wait is bounded in correct operation — `chklocks` releases every lock at its wall-clock expiry.

## Tests

New coverage:

- `swap_settles_on_amm_quote_not_caller_target` — the exploit verbatim (refused), a >100% tolerance (still refused), and a target the tolerance legitimately admits (accepted, still settles on the quote).
- `swap_settlement_debits_destination_reserve_by_the_quote` — full race to CONFIRMED, asserted against the reserve books.
- `applyswap_rejects_debit_above_curve_output`, `applyfromwire_rejects_debit_above_curve_output`, `paywire_rejects_payout_above_curve_output` — boundary (curve output + 1) and drain-shaped.
- `prune_keeps_terminated_operator_until_collateral_settles` — win a race, terminate, prove the row survives past the delay, release the locks, prove it is then erased.

Three existing cases encoded the old behavior and were rewritten rather than patched around: `swap_to_wire_oversized_target_is_rejected_not_aborted` (→ `..._refused_at_ingress`), `swap_same_token_legs_overcommit_is_disqualified`, `swap_same_token_legs_exact_balance_wins`.

Assertions derive the expected quote from `opp::amm::quote_swap` over the live reserve rows, so they pin the invariant rather than one fixture's arithmetic.

## Verification

| Suite | Result |
|---|---|
| `contracts_unit_test --sys-vm` | **No errors detected** |
| `unit_test --sys-vm` | **No errors detected** (1515 cases) |
| `plugin_test` | **No errors detected** |
| `test_fc` | 4 failures, all `test_http_client.cpp` DNS/connect — this diff touches zero host code |

## ABI / codegen — needs sequencing with #549

Addressing review comment 1 required retaining the caller's original `target_amount` on the `uwreqs` row, so **`sysio.uwrit`'s ABI gains one additive field**. `sysio.opreg` and `sysio.reserv` remain byte-identical.

Regenerating `SysioContractTypes.ts` against this branch produces exactly one new line:

```ts
target_amount: number | string   // in the uwreqs row interface
```

No consumer reads it, so nothing breaks if the regen lands later — but **it must land after #549 merges**, not now: `sdk-core`'s copy is currently *ahead* of `wire-sysio` master (it already carries #549's `sysio.chalg` types), so regenerating from this branch today would delete them. Sequence is: merge #549 → merge this → regenerate once → companion `wire-libraries-ts` PR.

Until then the e2e gate needs only the `wire-sysio` branch override; the stale generated type is harmless for an additive field.

### Behavior changes since the first review pass

- The variance / unpriceable terminal verdicts moved **after** the bond check (review 3), so the "variance rejection can now fire for an under-bonded candidate" note from the original description no longer applies — that was the bug being reported.
- Race-time slippage is measured against `target_amount`, not the previous quote (review 1).
- `regoperator`'s re-registration replace now carries the settlement gate (review 2).
- `prune`'s budget counts rows examined rather than erased (review 4).


contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -1187,6 +1224,73 @@ void try_select_winner(name self, uint64_t uwreq_id, name candidate) {
...
+                      uwreq_id, " (unprovisioned LP), skipping\n");
+         return;
+      }
+      // Variance — the user's slippage bound, measured against the live quote
+      // (see `variance_allowance`). `req.dst_amount` is the price quoted when
+      // the row was created; drift beyond tolerance refunds the source side and
+      // REJECTs rather than settling at a price the user never accepted.
+      // Non-throwing throughout (`feedback_opp_handlers_never_throw`).
+      if (req.dst_amount != 0) {
+         const uint64_t  quoted  = req.dst_amount;
```
huangminghuang: **[High][bug] Preserve the original target for the race-time slippage check.** `req.dst_amount` now holds the ingestion quote, so this compares the settlement quote with the previous quote rather than with the original `target_amount`, compounding the tolerance. For example, target 100 with 10% tolerance accepts an ingestion quote of 91 (`9 <= 9`) and then a settlement quote of 83 (`8 <= 8`), even though the final result is 17% below the target. The unprovisioned path is worse: it stores `dst_amount == 0`, so this guard is skipped entirely if a reserve becomes priceable later. Please preserve or decode the original target and compare the final live quote directly against it.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -317,6 +317,43 @@ uint64_t sum_locks_inline(name account, sysio::slug_name chain_code, sysio::slug
...
+///
+/// Balance rows are checked by VALUE, not by `balances.empty()` —
+/// `subtract_balance` zeroes a bucket in place and never erases the entry.
+///
+/// A row that never settles is never pruned, and that is the intended outcome:
+/// an unprunable row costs some RAM, while an erased one costs the operator
+/// their bond. In correct operation the wait is bounded — `chklocks` releases
+/// every lock at its wall-clock expiry, and each `releaselock` drains the
+/// matching bucket, so the predicate goes true on its own.
+bool is_fully_settled(const opreg::operator_entry& op) {
```
huangminghuang: **[High][bug] Apply this settlement gate to re-registration as well.** This predicate protects `prune`, but `regoperator` still erases any existing TERMINATED row at lines 246–251 without checking settlement. A terminated underwriter can therefore re-register while locks remain, replacing the retained balance row with an empty healthy row; when `chklocks` later calls `releaselock`, it sees that new non-terminated row and no-ops, stranding the old collateral exactly as WNS-01 describes. Every terminated-row erase path should require full settlement, or re-registration should remain blocked until the old row is safely pruned.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
```diff
@@ -1187,6 +1224,73 @@ void try_select_winner(name self, uint64_t uwreq_id, name candidate) {
       return;
    }
 
+   // ── Settlement quote — re-price on the live curve, then variance ─────
+   // Runs BEFORE the bond check because it decides `dst_amount`, and the bond
+   // must cover what the winner will actually deliver. Ordering it after the
+   // candidate-specific gates above (eligibility, signature) is deliberate: a
+   // bogus candidate poking this resolver must not be able to terminally close
+   // a healthy request. A candidate that clears those gates but is under-bonded
```
huangminghuang: **[Medium][security] Do not let an under-bonded candidate terminally reject the request.** This request-wide variance or unpriceable decision now runs before per-request bond sufficiency. An ACTIVE underwriter can satisfy the role minimum yet lack enough collateral for this swap; with valid UICs it can trigger a terminal refund during transient drift before reaching the bond disqualification below. Compute the live quote first so the required bond is known, but only allow the request-wide terminal mutation after the candidate proves sufficient collateral.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -1588,14 +1625,24 @@ void opreg::prune() {
...
            magic_enum::enum_integer(OperatorStatus::OPERATOR_STATUS_TERMINATED));
         it != status_idx.end() &&
         it->status == OperatorStatus::OPERATOR_STATUS_TERMINATED;) {
-      if (it->terminated_at > 0 && now - it->terminated_at >= cfg.terminate_prune_delay_ms) {
+      const bool delay_elapsed = it->terminated_at > 0
+                                 && now - it->terminated_at >= cfg.terminate_prune_delay_ms;
+      if (delay_elapsed && is_fully_settled(*it)) {
          it = status_idx.erase(std::move(it));
-         if (++removed >= 20) break; // Bound CPU
+         if (++removed >= MAX_OPERATOR_PRUNE_PER_CRANK) break; // Bound CPU
```
huangminghuang: **[Medium][performance] Bound examined rows, not only successful erases.** `removed` increments only on this branch, while every delay-elapsed but unsettled TERMINATED row executes `is_fully_settled()` and its lock-index probe without consuming the cap. A sufficiently large prefix of unsettled rows therefore makes this permissionless crank unbounded and can exceed its CPU deadline; each retry also restarts at the same lower bound. This needs a bounded continuation or cursor, or an ordering/index that can skip non-prunable rows.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Agreed, and fixed in 8219091a — this was a real gap, not a tradeoff I should have taken.

`uw_request_t` now carries `target_amount`: the amount the caller asked for, retained as the fixed reference the slippage check measures against and never paid out. `try_select_winner` compares the live settlement quote directly against it.

Both halves of your example are closed. The compounding one, verbatim as a test precondition — the new case asserts the scenario actually sits in the window before asserting the verdict, so it fails loudly rather than silently passing if the fixture arithmetic ever drifts:

```cpp
BOOST_REQUIRE_LE(ingestion_quote - settle_quote, allowed);   // old rule: would pass
BOOST_REQUIRE_GT(target - settle_quote,          allowed);   // new rule: must reject
```

The unprovisioned path is closed by construction: `target_amount` is guaranteed positive by `createuwreq`'s positive-amounts gate and `swapfromwire's `check`, so the bound no longer depends on `dst_amount` being non-zero and cannot be skipped on a row created before an LP existed.

I went with preserving rather than decoding: it is O(1) with no decode-failure path, and the row carrying the user's declared bound is worth having for audit anyway. Cost is an additive `sysio.uwrit` ABI field — see the PR description for the sdk-core sequencing that implies.

contracts/sysio.opreg/src/sysio.opreg.cpp
heifner: Confirmed and fixed in 8219091a. I gated the symptom and missed that `prune` was not the only erase path — `regoperator:251` replaces a TERMINATED row unconditionally, and your trace is exactly right: `releaselock` settles only SLASHED / TERMINATED operators, so it sees the fresh healthy row and no-ops, stranding the old collateral.

`is_fully_settled` moved into the earlier anonymous namespace so it sits above `regoperator`, and both erase paths now share it. Re-registration stays blocked until settlement lands:

```cpp
check(is_fully_settled(existing),
      "operator has unsettled collateral: a terminated operator may only "
      "re-register once its balances are drained and no underwriting locks remain");
```

The wait is bounded — `chklocks` releases every lock at its wall-clock expiry — so this delays re-registration rather than blocking it indefinitely. New test `reregistration_blocked_until_collateral_settles` drives the whole arc: win a race, terminate, prove re-registration is refused with the balance intact, let the locks expire, prove it then succeeds.

contracts/sysio.uwrit/src/sysio.uwrit.cpp
heifner: Fixed in 8219091a, and your framing is the right one — I had documented this reordering as acceptable in a code comment, which was me arguing for it rather than fixing it. The DoS shape you describe is real: role-minimum ACTIVE is a much weaker property than can-settle-this-swap.

Split exactly as you suggested. The quote is still computed first, because it is what sizes the obligation the bond has to cover:

```cpp
const uint64_t settle_quote = swap_quote(...);
if (settle_quote != 0) { req.dst_amount = settle_quote; }   // bond sized on the live obligation
// ... bond check -> disqualify_candidate on shortfall ...
// ── Request-level terminal decisions ── (unpriceable / variance) run only here
```

So a candidate reaching a terminal verdict has proved it could actually settle, and the verdict it triggers is the request's own rather than its own inability. When the curve cannot price the swap, the stored ingestion quote stands in for the bond check and the request is terminated (or skipped) immediately after.

New test `swap_underbonded_candidate_cannot_terminally_reject`: price drifted far outside tolerance so the request is genuinely doomed, candidate holds 10 units against a ~2e6 aggregate — asserts the row stays PENDING with the candidate DISQUALIFIED on bond, not REJECTED.

contracts/sysio.opreg/src/sysio.opreg.cpp
heifner: Fixed in 8219091a. You are right that the budget was measuring the wrong thing — I added a lock-index probe per candidate and left the counter on the erase branch, so the scan cost became unbounded in exactly the case the gate creates (a prefix of delay-elapsed but unsettled rows).

The budget now counts rows EXAMINED, matching the convention already used by `flushwtdw`'s `flushed` and `drainfwq`'s `drained` (both increment before any per-row `continue`):

```cpp
uint32_t examined = 0;
for (auto it = ...; it != status_idx.end() && ... && examined < MAX_OPERATOR_PRUNE_PER_CRANK;) {
   ++examined;
   ...
}
```

On the cursor: I did not add one, and I want to be explicit about what that leaves rather than imply it is fully solved. Rows beyond the budget do wait for a later crank, and each crank restarts at the same lower bound as you note. I believe that converges rather than starves — every unsettled row is settled by `chklocks` at its locks' wall-clock expiry and then erased, so the unprunable prefix drains on its own and the scan advances behind it. The bound is the lock window, not unbounded.

That reasoning is in the code comment so it can be challenged rather than assumed. If you would rather have a real cursor (a stored resume key, or an index that can skip non-prunable rows), say so and I will add it — the convergence argument holds for the current single-digit terminated-operator population but it is an argument, not a guarantee.

contracts/sysio.uwrit/include/sysio.uwrit/sysio.uwrit.hpp
```diff
@@ -546,11 +557,30 @@ namespace sysio {
...
+         ///
+         /// It is retained on the row precisely so `try_select_winner` can
+         /// compare the LIVE settlement quote against the user's ORIGINAL bound.
+         /// Comparing instead against the previous quote (what `dst_amount`
+         /// holds before the race re-prices) would compound the tolerance across
+         /// the two checkpoints — a 10% tolerance accepting a 91 quote at
+         /// ingestion and an 83 quote at settlement, 17% below a target of 100 —
+         /// and would skip the bound entirely on a row created while no LP was
+         /// provisioned (`dst_amount == 0`).
+         uint64_t                                target_amount     = 0;
```
huangminghuang, line 575: [P1] Preserve compatibility with existing `uwreqs` rows

`uwreqs` already exists on the base branch, and `SYSLIB_SERIALIZE` encodes and decodes its members positionally. Inserting a plain `uint64_t` here before `variance_tolerance_bps` causes rows written by the current contract to be misdecoded after upgrade: the old variance and source bytes are consumed as `target_amount`, shifting every following field and potentially aborting deserialization. Because `sysio.epoch::advance` invokes `pruneuwreqs` inline and that sweep reads these rows, a legacy row can stall epoch advancement. Please provide a migration or verified empty-table deployment gate, or use a backward-compatible trailing extension with explicit legacy-row handling; the original target is also available in the stored inbound payload. Fresh-state tests do not cover this upgrade path.

huangminghuang (dismissed):
Requesting changes for one upgrade-safety blocker: the new persisted field is not backward-compatible with existing uwreqs rows.

contracts/sysio.uwrit/include/sysio.uwrit/sysio.uwrit.hpp
huangminghuang: Retracting this finding: the project is still pre-launch and deployments start from fresh state, so there are no persisted legacy `uwreqs` rows that need to survive this schema change. The review has been dismissed.

huangminghuang (approved):
Approved after re-review. The four earlier findings are addressed in 8219091a, and the persisted-row compatibility concern is not applicable during the project pre-launch stage.

PR #551, opened 2026-08-07
security: validate contract inputs and bound registry metadata (WNS-03, WNS-09, WNS-10, WNS-11, WNS-12)

heifner:
The remaining input-validation findings from CertiK "Wire Network - Sysio Audit 1", closing WIRE-318, WIRE-323, WIRE-324, WIRE-325 and WIRE-326.

**No ABI changes** — every fix is an added guard, so no `SysioContractTypes.ts` regeneration and no downstream `wire-libraries-ts` / `wire-tools-ts` rebuild is required. Only the five `.wasm` artifacts changed; no `.abi` did.

| Finding | Severity | Ticket |
|---|---|---|
| WNS-03 — `roa::reducepolicy` accepts negative weights and expands quotas | Major | WIRE-318 |
| WNS-09 — `token::create` stores the issuer without validating it | Minor | WIRE-323 |
| WNS-10 — `tokens::regtoken` does not bound `symbol_name`/`description` | Minor | WIRE-324 |
| WNS-11 — `regchain` does not enforce the canonical `WIRE` code | Informational | WIRE-325 |
| WNS-12 — unused `sysiosystem::system_contract` forward declaration | Optimization | WIRE-326 |

## WNS-03 — negative reduction weights inflate quota

`reducepolicy` bounded the request only from above (`w <= stored`), which any negative amount satisfies whenever the stored weight is positive. The weight is then applied as a subtraction, so a negative one *increased* the account's quota:

```cpp
new_net = (net_limit < 0) ? -1 : std::max(0, net_limit - net_weight.amount);
```

A node owner could inflate an account's NET/CPU past the issuer's ROA budget — bypassing `expandpolicy`'s free-allocation check — and desynchronise the `reslimit` row and the issuer's `nodeowners` accounting from the policy weights. `addpolicy` and `expandpolicy` both reject negatives already; `reducepolicy` was the outlier.

CertiK's PoC (a 10.0000 SYS policy reduced by −5.0000 SYS ending at 15.0000 SYS) is now `reducepolicy_negative_weight_rejected`, covering all three weights and asserting the policy **and** the account's on-chain resource limits are untouched after the rejected attempts.

## WNS-09 — unvalidated issuer

`issuer` was trusted input after `require_auth(get_self())`, but `issue` gates on `to == st.issuer` + `require_auth(st.issuer)`. A null or non-existent issuer produced a token nobody could ever issue while permanently burning the symbol, since `create` rejects duplicates. Now checks `issuer.value != 0` and `is_account(issuer)`, placed after the supply checks so existing error ordering is preserved.

## WNS-10 — unbounded metadata in system-paid state

CertiK raised this on `tokens::regtoken`, but the **identical unbounded `name`/`description` pair, with the same `ram_payer = sysio` billing, was also on `chains::regchain` and `reserv::regreserve`**. Rather than fix one instance of a three-instance defect, the limits live once in a new shared header:

`contracts/sysio.opp.common/include/sysio.opp.common/registry_metadata.hpp` — `label_max_bytes = 128`, `description_max_bytes = 256` (the latter matching the established `token::issue` memo bound).

All three privileged registrations call `check_metadata()` before `emplace`.

### The label bound started at 32 and was wrong

The e2e gate caught it: run 31258456527 came back 6/13, two failures mine. `flow-reserve-lifecycle` and `flow-swap-private-reserves` create reserves through `create_reserve` → `oncrtreserve`, and four of their labels exceed 32 bytes (longest 42). Over-bound metadata routes into the CANCELLED/refund path, so a reserve those flows expect PENDING came back CANCELLED.

32 was wrong, not the names. WNS-10 is about *unbounded* strings consuming up to the KV/action ceiling of system-paid state; it says nothing about how terse a label should be, and a reserve naming its full leg is what a reader wants in the registry. The bound is 128 — still a real cap, with room for descriptive names rather than sized to the current longest. The header says so explicitly, so the next person who hits it questions the bound before shortening a legitimate label.

### `reserv::oncrtreserve` needed different handling

It carries the same two strings from an outpost-side creator, but it is an OPP inbound dispatch handler — a `check()` there rolls back the consensus-tipping delivery and stalls epoch advancement chain-wide (`feedback_opp_handlers_never_throw`, `epoch-stall-is-fatal`). It instead uses the non-throwing `metadata_exceeds_bounds()` and joins the **existing** reject predicate alongside `invalid_amount` and the unlinked-creator case, releasing the creator's escrow through the `RESERVE_CREATE_CANCELLED` flow that path already emits.

The CANCELLED tombstone is itself a `sysio`-billed row storing `name`/`description`, so writing the oversized strings onto it would persist exactly the state the bound prevents. **The rejected row stores a fixed `<rejected>` marker as its name and an empty description — nothing is truncated.** Truncation would have had to cut at a byte offset rather than a UTF-8 code-point boundary, and the salvaged text buys nothing: nothing reads a tombstone's metadata (the reclaim path overwrites every field) and the creator's originals are preserved in the inbound OPP envelope artifact regardless. The earlier `truncate_label` / `truncate_description` helpers were removed.

`oncrtreserve_oversized_metadata_is_cancelled` covers three cases — an over-bound ASCII label (129 bytes), an over-bound multibyte label (127 ASCII + `é` = 129 bytes, which a byte-wise clamp would have split), and an over-bound description (257 bytes) behind an in-bound name so each half of `metadata_exceeds_bounds` is driven independently. All three assert CANCELLED, the `<rejected>` marker, and — for the description case — an empty stored description.

The header makes the split explicit: privileged abort-safe registrations `check`; dispatch handlers ask and route.

## WNS-11 — depot identity not pinned to its code

Bootstrap invariant V3 (`docs/platform-bootstrap-config.md`) is "exactly one `CHAIN_KIND_WIRE` chain, code `WIRE`". Only the cardinality half was enforced on-chain, and `is_depot` is derived from the kind alone — so a registration could claim depot identity under any code (`FAKE`), with the code's validity resting entirely on the off-chain config validator.

The guard is bidirectional: `kind == CHAIN_KIND_WIRE` requires `code == "WIRE"`, and the `else` branch rejects `code == "WIRE"` under any other kind. The forward check alone still admitted `regchain(EVM, "WIRE", ...)`, which — with code uniqueness and no erase action — would have permanently bricked registration of the canonical depot row. Both orderings are covered in `sysio.epoch_tests.cpp`.

## WNS-12

Removed the unused `sysiosystem::system_contract` forward declaration from `sysio.token.hpp`; it implied a `sysio.token` → `sysio.system` dependency that does not exist.

## Verification

Re-run at the final head (`f8ba67a`):

| Binary | Result |
|---|---|
| `contracts_unit_test` (full, `--sys-vm`) | 589 cases, no errors |
| `unit_test` (full, `--sys-vm`) | 1515 cases, no errors |
| `plugin_test` | no errors |

`unit_test` matters here beyond the usual sweep: `unittests/test_contracts.hpp.in` loads `sysio.token.wasm` from the same build path, so the new `is_account(issuer)` check is live in those tests too. Every `token::create` call site in `unittests/` uses `alice` or `sysio.token`, both created in their fixtures.

Harness compatibility was re-verified against the actual data rather than a sample: every `*Name` / `*Description` literal across all `wire-tools-ts` packages fits, with the longest label at 42 bytes (3× headroom under 128) and the longest description at 137. The original check looked only at `RegistrySteps.ts` — the bootstrap registrations, max 23 bytes — and missed the flow-level `create_reserve` names, which is exactly where the 32-byte bound broke.

## Reviewer note

The WNS-10 fix extends past CertiK's literal `regtoken` scope to the three sibling registries plus the `oncrtreserve` dispatch path. That was a deliberate call — same defect, same billing account — but it is the one part of this PR that is wider than the finding, so it is the part worth a second opinion.


huangminghuang (commented):
Two actionable findings from the delegated OCR review.

contracts/sysio.chains/src/sysio.chains.cpp
```diff
@@ -40,14 +48,24 @@ void chains::regchain(opp::types::ChainKind kind,
...
 
-   // Enforce: at most one row with kind == WIRE (the depot self-row).
+   // Enforce: the depot self-row is unique AND canonically coded.
+   //
+   // `is_depot` is derived from the KIND alone below, so the code half of bootstrap
+   // invariant V3 has to be enforced here or a registration could claim depot identity
+   // under any code (e.g. `FAKE`) -- previously only the cardinality half was on-chain
+   // and the code depended entirely on the off-chain config validator.
    if (kind == opp::types::CHAIN_KIND_WIRE) {
+      sysio::check(code == WIRE_CHAIN_CODE,
```
huangminghuang, line 72: **[P2] Reserve the `WIRE` code bidirectionally**

This guard only handles `kind == CHAIN_KIND_WIRE`. During bootstrap, `regchain(CHAIN_KIND_EVM, "WIRE", ...)` can still succeed first; code uniqueness then permanently prevents registering the canonical depot row, and there is no erase action. Enforce the inverse implication too: `code == WIRE_CHAIN_CODE` must require `kind == CHAIN_KIND_WIRE`, with a regression test for this ordering.

contracts/sysio.opp.common/include/sysio.opp.common/registry_metadata.hpp
```diff
@@ -0,0 +1,96 @@
...
+/// verbatim — otherwise rejecting an oversized registration would persist exactly the
+/// state the bound exists to prevent.
+inline std::string truncate_label(std::string label) {
+   if (label.size() > label_max_bytes) label.resize(label_max_bytes);
+   return label;
+}
+
+/// Clamp a description for storage on a reject-path tombstone row. See `truncate_label`.
+inline std::string truncate_description(std::string description) {
+   if (description.size() > description_max_bytes) description.resize(description_max_bytes);
```
huangminghuang: **[P3] Preserve UTF-8 when truncating metadata**

`resize()` truncates bytes rather than UTF-8 code points. A valid 33-byte label consisting of 31 ASCII bytes plus `é` is cut midway through the final character, persisting malformed human-readable metadata in the `CANCELLED` row. Clamp at a valid UTF-8 boundary or store a safe fixed/empty tombstone value, and add a multibyte boundary test.

heifner:
Both findings were real and are fixed in 537d436. Thanks — P2 in particular was a genuine gap, not just a hardening nit.

**[P2] Reserve the `WIRE` code bidirectionally** — confirmed and fixed.

You're right that the forward implication alone left the hole open, and the consequence is worse than a mis-registration: chain codes are unique and there is no erase action, so an EVM/SVM row registered under `WIRE` permanently squats the depot's identity and leaves the canonical self-row unregisterable — bootstrap bricked with no on-chain recovery path. The inverse is now enforced on the `else` branch with its own message, so "wrong code for a WIRE chain" and "reserved code on a non-WIRE chain" stay separately diagnosable.

`regchain_wire_code_reserved_from_other_kinds` attempts the squat with both EVM and SVM **before** the depot row exists — the ordering being the whole point of the finding — then asserts the canonical row still registers afterwards and that non-`WIRE` codes are unaffected.

**[P3] Preserve UTF-8 when truncating** — confirmed and fixed.

Your example is exactly what it did: 31 ASCII + `é` (0xC3 0xA9) at a 32-byte clamp kept the lone 0xC3 lead byte. I went with clamping to a valid boundary rather than a fixed/empty tombstone value, since the tombstone's metadata has no consumer in logic — the reclaim branch overwrites every field — so its only remaining value is forensic ("what did the creator try to name it"), and an empty value throws that away.

`clamp_utf8` walks back off continuation bytes (`10xxxxxx`) so a straddling character is dropped whole; it's in the shared header, so both `truncate_label` and `truncate_description` get it. The bound itself deliberately stays a **byte** bound — it exists to cap state size, and the input is operator-relayed bytes with no guarantee of being well-formed UTF-8 in the first place. `oncrtreserve_metadata_clamp_respects_utf8` asserts the stored label is the 31 ASCII bytes, not 32.

Verification: `contracts_unit_test` 590 cases, no errors (588 before these two tests). `unit_test` and `plugin_test` are unchanged from the earlier green run — this round touched only `sysio.chains`, `sysio.reserv` and the shared header, none of which those binaries load.

contracts/sysio.opp.common/include/sysio.opp.common/registry_metadata.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * persist malformed text in state. Backing off to the previous boundary drops the straddling
+ * character whole. The BOUND itself stays a byte bound -- it exists to cap state size.
+ */
+inline std::string clamp_utf8(std::string s, std::size_t max_bytes) {
+   if (s.size() <= max_bytes) return s;
+   // `s[cut]` is the first dropped byte. While it is a continuation byte (10xxxxxx) the
+   // character straddling the boundary is being split, so walk back onto its lead byte and
+   // drop the whole sequence. Terminates at 0 in the worst case.
+   std::size_t cut = max_bytes;
+   while (cut > 0 && (static_cast<unsigned char>(s[cut]) & 0xC0) == 0x80) --cut;
```
huangminghuang: **[P3] Name the UTF-8 bit masks** — The repository’s “No magic literals” rule requires nontrivial numeric values behind named constants. These masks define the correctness of the UTF-8 boundary check, so introduce named `inline constexpr` values for `0xC0` and `0x80` and use them here.

huangminghuang (commented):
One low-severity project-rule finding from the follow-up review.

contracts/tests/sysio.reserv_tests.cpp
```diff
@@ -601,6 +624,61 @@ BOOST_FIXTURE_TEST_CASE(oncrtreserve_invalid_amount_is_cancelled, sysio_reserve_
...
+   BOOST_REQUIRE_EQUAL(success(),
+      recordlink_em("alice"_n, ChainKind::CHAIN_KIND_EVM, creator_pub));
+
+   auto create_with_name = [&](std::string_view reserve_code, const std::string& name) {
+      return push_action(MSGCH_ACCOUNT, "oncrtreserve"_n, mvo()
+         ("chain_code",            codename_mvo("ETH"))
+         ("token_code",            codename_mvo("ETH"))
+         ("reserve_code",          codename_mvo(reserve_code))
+         ("name",                  name)
+         ("description",           "")
```
huangminghuang: **[P3] Cover the oversized-description path** — This regression test always passes an empty description, so it never exercises the `description.size() > description_max_bytes` half of `metadata_exceeds_bounds` or verifies that an over-bound description cannot be persisted. Add a description-only case with an in-bound name and a 257-byte description, then assert CANCELLED status, the `<rejected>` marker, and an empty stored description.

huangminghuang (commented):
One low-severity test-coverage finding from the latest follow-up review.

huangminghuang (approved):
Approved — the code review is clean.

Before merge, please refresh the PR description so it matches the final implementation: `label_max_bytes` is 128 (not 32); oversized tombstones now store `<rejected>` and clear the description (the truncation helpers were removed); the compatibility note should reflect the 42-byte real labels/e2e follow-up; and the verification counts should be updated for the final head.

PR #552, opened 2026-08-07
feat(opp-solana): drive resumable inbound dispatch; delete the depot-side terminal budget

valthon:
## Summary

The depot-side half of resumable OPP dispatch (paired with wire-solana#419), plus a proto-generator fix that closes a decoder-panic wedge.

Previously the depot delivered an inbound envelope to the Solana outpost and settled it in one terminal `epoch_in` call, bounded by a fixed **terminal budget** the depot computed and enforced so the whole envelope fit one transaction. That cap made large envelopes undeliverable and coupled the depot to the outpost's per-transaction limits.

## Commits

1. **Drive resumable Solana dispatch from the depot** — delete the depot-side terminal budget (`solana_terminal_budget.hpp` and its `sysio.msgch` enforcement) and drive the outpost's two-phase crank instead: `sysio.msgch` stages chunks and tips consensus, and the `outpost_solana_client` plugin settles via the permissionless `dispatch_attestations` instruction, reading the on-chain cursor and resuming until every attestation has dispatched. The relay's drain is ordered **consensus-first** (a cheap no-op before any per-attestation manifest work when consensus has not tipped), carries the **remaining** deadline budget into the drain so the dispatch-phase guard is reachable, and memoizes the delivered envelope **only after** the buffer is fully staged so a partial upload re-uploads from zero rather than cranking an incomplete buffer. The stub IDL tracks the outpost's slimmed seven-account `epoch_in` surface and the now read-only dispatch `chunk_buffer`; `docs/opp-two-phase-dispatch-design.md` records the design.

2. **Guard nested decodes in the Solana proto generator** — the generated decoders sliced a nested/length-delimited field without checking the length, so a malformed length prefix **panicked** instead of returning `DecodeError` — on-chain, a permanently wedged outpost epoch reachable by any depot that emits a structurally malformed field. Every length-delimited decode the generator emits (nested-message, repeated, and map templates **and** the emitted `decode_bytes` / `skip_field` runtime helpers) is now guarded with `checked_add(...).ok_or(BufferOverflow)?` before the bounds check. The checked add also closes an overflow route a bare bounds check misses: `new_pos + len` overflows `usize` before any comparison for a `u64::MAX`-scale varint length, and the program builds release with overflow-checks on. Regenerated decoders return `DecodeError`, and the outpost handlers' existing log-and-skip discipline does the rest. Runtime unit tests cover both routes.

## Testing

- `sysio.msgch` contract tests (21 cases) and the `outpost_solana_client` plugin suite green after rebasing onto current `master`.
- Proto-generator jest + Rust runtime suites green (both decode routes under overflow-checks).
- End-to-end paired with wire-solana#419: the decoder guard was confirmed live by the wire-solana panic pins flipping before their retirement.

Addresses the review feedback on the prior revision (consensus-first drain ordering, remaining-budget deadline, inert-seed removal, stub-IDL account surface); the session-scaffolding docs flagged in review have been removed.


plugins/outpost_solana_client_plugin/include/sysio/outpost_solana_client_plugin/outpost_solana_client.hpp
```diff
@@ -19,14 +19,12 @@ namespace sysio {
...
-/// Finalization happens in a separate zero-data terminal call, so data chunks
-/// do not reserve packet space for dynamic effect accounts. If the static
-/// `epoch_in` account list changes, the SEC-94 fixture/parity test must be
-/// regenerated and will catch any packet overflow.
-inline constexpr size_t SOLANA_MAX_CHUNK_BYTES = 672;
+/// Must equal `MAX_CHUNK_BYTES` on the Solana side — non-final chunks must be
+/// exactly this size. 668 keeps a full data chunk under Solana's 1 232 B packet
+/// limit (562 B tx overhead, `dispatch_limit` included). Any change to the
+/// `epoch_in` account list or argument tuple moves that overhead;
+/// `epoch_in_full_data_chunk_fits_packet_limit` is what catches it.
```
heifner: **Lockstep break: the program on the manifest-pinned wire-solana `next` has none of this contract.**

Checked wire-solana `next` (local == `origin/next`, and `wire-platform-manifest/default.xml` pins `next`):

- `programs/liqsol-core/src/states/opp_states.rs:54` — `MAX_CHUNK_BYTES = 672`, and `epoch_in.rs:122-134` requires non-final chunks to be **exactly** 672 → every envelope > 668 bytes from this relay is rejected with `ChunkSizeMismatch` (6027); the epoch never assembles and SOL OPP stalls in both directions. Envelopes ≤ 668 bytes are single-chunk and still pass, so this breaks intermittently under load.
- `epoch_in.rs:17` / `lib.rs:1031-1047` — no `dispatch_limit` arg. anchor 0.31.1 uses `BorshDeserialize::deserialize` (trailing bytes tolerated), so on a single-chunk terminal call the 4 extra bytes are **silently dropped** and dispatch runs unbounded — no error, no signal, while the relay believes it throttled.
- `opp_states.rs:704-710` — `EpochDeliveries` has no `dispatched_count`; no resumability exists.
- `errors/opp_errors.rs` tops out at 6063 — no `InvalidDispatchLimit` / `EffectAccountMissing`; missing effect accounts still **log-and-skip** (`inbound.rs:78-90`, `:710-717`, `:803-825`), so on the pinned program a missing account is a silently lost remit with the epoch then closed — the exact failure this PR exists to close.

No commit on any wire-solana ref implements the new contract (`git log --all -S"dispatch_limit"` is empty). This PR must land/deploy in lock-step with the companion program change, and the two new error codes should be appended at 6064/6065 (the numeric table in `opp-outpost-technical-spec.md:72-105` and `scripts/opp/patch-idl-errors.js` both key on stable numbering).

heifner (commented):
Review of the resumable-dispatch change: 10 findings, ranked by severity. Nine are anchored inline on the diff; one sits in a file this PR does not touch and is included here.

---

**Retry idempotence / fee burn — `plugins/batch_operator_plugin/src/outpost_opp_job.cpp:120` (file not in this diff)**

Nothing records that chunks were already staged, and `_last_outbound_epoch` only advances on a clean return (`outpost_opp_job.cpp:120-134`). Every throw path in the new, much longer delivery (deadline, program abort, decode failure, stale cursor read) leaves it unset, so the cron re-enters `deliver_outbound_envelope` every 15 s, which unconditionally re-sends **all** data chunks from offset 0 (`outpost_solana_client.cpp:908-923`) before reaching the terminal loop. For a wedged epoch that is an unbounded fee burn and RPC load — 4 tx/tick for the dev-026 envelope, 99 tx/tick at the 64 KiB cap — with no backoff, and the re-upload itself eats into the same 15 s budget, making the deadline-throw wedge self-reinforcing. Suggest probing the chunk_buffer PDA (or a local memo) before re-uploading.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -924,27 +915,76 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
-      {},
-      std::move(terminal_accounts));
-   ilog("outpost_solana_client[{}]: epoch_in terminal finalize sent epoch={} chunk={}/{} bytes=0 extras={} sig={}",
-        to_string(), epoch_index, total_chunks, total_chunks, terminal_extra_count, last_sig);
+   // Terminal calls, looped until the on-chain cursor drains. Each call
+   // carries the account union for exactly the attestations it claims via
+   // `dispatch_limit` -- the relay is the only party that can size the batch,
+   // because it supplies the accounts and therefore knows the transaction's
+   // real byte cost.
+   uint32_t settled = 0;
```
heifner: **`settled` is never seeded from the on-chain cursor — any re-drive or late delivery against a partially-drained cursor wedges the epoch permanently.**

`read_epoch_dispatch_progress` is first called only at line 963, *after* the first terminal `epoch_in` has been submitted. So the first call of every invocation packs accounts for `[0, batch_end)` and claims `dispatch_limit = batch_end`, while the program settles `[dispatched_count, dispatched_count + dispatch_limit)`.

Concrete wedge: an epoch needs 2 terminal rounds; round 1 settles `[0,5)`, then the 15 s delivery deadline trips at line 932. `outpost_opp_job.cpp:120-144` catches, leaves `_last_outbound_epoch` unset, and retries next tick — which re-enters with `settled = 0`, packs accounts for `[0,5)`, and sends `dispatch_limit = 5`; the program dispatches `[5,10)` from its cursor and aborts with `EffectAccountMissing` per this PR's own contract. `execute_tx_and_confirm` **throws** on a failed tx (`libfc solana_client.cpp:1737-1741`), so the self-heal read at 963 is never reached, and every retry repeats identically: attestations 5-9 (remits/slashes) never dispatch. Same wedge via a late operator delivering after another operator partially drained (its first call sees cursor K>0), `MAX_TERMINAL_ROUNDS` exhaustion + the `retry_pending` redelivery, or a nodeop restart. Note the single-round case is benign (`[K,total) ⊂` accounts for `[0,total)`), so this bites exactly the multi-round case the feature exists for.

Fix: seed `settled = read_epoch_dispatch_progress(epoch_index).dispatched_count` before the loop, and skip the loop when `settled >= total_attestations` (still sending one terminal call to record delivery if needed).

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -924,27 +915,76 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+         auto candidate = batch_accounts;
+         for (const auto& meta : per_attestation[batch_end]) {
+            outpost_solana_client_detail::record_terminal_account(candidate, meta.key, meta.is_writable);
+         }
+         if (candidate.size() > MAX_TERMINAL_DYNAMIC_ACCOUNTS && batch_end > settled) break;
+         batch_accounts = std::move(candidate);
+         ++batch_end;
+      }
+      const uint32_t dispatch_limit = batch_end - settled;
+      FC_ASSERT(dispatch_limit > 0,
```
heifner: **Zero-attestation path: on envelope decode failure this asserts *before any terminal call* — the delivery is never recorded. Regression vs. the base.**

`count_inbound_attestations` returns 0 on `ParseFromArray` failure (lines 411-416); the packing `while` then never runs, `dispatch_limit = 0`, and this `FC_ASSERT` throws. All data chunks were already staged (line 917), but no terminal `epoch_in` is ever sent → `EpochDeliveries` never records this operator → consensus can never tip → the depot row stays `PENDING_DELIVERY` and the job re-uploads the whole envelope every 15 s forever. The merge base (`2b74bed5a0`) sent the terminal finalize unconditionally with zero extras and let effects log-and-skip on-chain.

The PR contradicts itself on this path: the `extract_inbound_effects` wlog (lines 429-435) still says *"submitting epoch_in with no extras"*, and the new test's comment (`test_outpost_solana_client_plugin.cpp:880-882`) states *"the relay submits a terminal call claiming nothing"* — behavior the loop no longer performs (the test only asserts on the extractors, so nothing catches it).

A genuinely zero-attestation envelope is not producible by the depot (`sysio.msgch.cpp` `check(included_count > 0, ...)`), so decode failure is the live trigger — guard `total_attestations == 0` and still send the delivery-recording terminal call.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -924,27 +915,76 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+      const size_t batch_extra_count = batch_accounts.size();
+      last_sig = _program_client->epoch_in(
+         epoch_index, total_chunks, total_chunks, static_cast<uint32_t>(total), {},
+         std::move(batch_accounts), dispatch_limit);
+      ilog("outpost_solana_client[{}]: epoch_in terminal finalize sent epoch={} "
+           "round={} window=[{},{}) of {} extras={} sig={}",
+           to_string(), epoch_index, round, settled, batch_end, total_attestations,
+           batch_extra_count, last_sig);
+
+      const auto progress = read_epoch_dispatch_progress(epoch_index);
```
heifner: **Commitment mismatch: the write confirms at `processed`, this read defaults to `confirmed`, no re-poll — a stale cursor read produces false breaks.**

`execute_tx_and_confirm` returns once the tx is `processed` (`libfc solana_client.hpp:45`, `solana_client.cpp:1707-1718`); `get_account_info` defaults to `confirmed` (`solana_client.hpp:586`) and `read_epoch_dispatch_progress` calls it immediately with no commitment argument and no retry (line 687). When the confirmed bank lags the processed slot, the read returns the pre-tx `EpochDeliveries` — or `nullopt` on the epoch's first delivery, i.e. zero-valued progress — so the loop breaks at 969 ("consensus not yet reached" when this very tx reached it) or 979 ("another operator is draining" when *this* relay advanced the cursor). Result: only the first batch dispatches per invocation and the remainder waits for a re-drive, which then hits the `settled = 0` wedge.

`read_inbound_envelope` already pins its read commitment deliberately (lines 1006-1010); this read should pin `processed` (or poll until the read catches up to the write) to match.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -924,27 +915,76 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
-      std::move(terminal_accounts));
-   ilog("outpost_solana_client[{}]: epoch_in terminal finalize sent epoch={} chunk={}/{} bytes=0 extras={} sig={}",
-        to_string(), epoch_index, total_chunks, total_chunks, terminal_extra_count, last_sig);
+   // Terminal calls, looped until the on-chain cursor drains. Each call
+   // carries the account union for exactly the attestations it claims via
+   // `dispatch_limit` -- the relay is the only party that can size the batch,
+   // because it supplies the accounts and therefore knows the transaction's
+   // real byte cost.
+   uint32_t settled = 0;
+   for (uint32_t round = 0; round < MAX_TERMINAL_ROUNDS; ++round) {
```
heifner: **Loop budgets don't fit the feature: the 15 s delivery deadline was sized for one terminal call, and `MAX_TERMINAL_ROUNDS = 64` can be exceeded by a legal envelope — with exhaustion returning silently as success.**

- `batch_operator_plugin.cpp:38-39` gives the entire delivery (N confirmed chunk txs + 2 `get_account_info` per unique reserve + M rounds of confirmed tx + read) 15 s — equal to the cron period. Any envelope needing ≥ 3 rounds makes the deadline throw at line 932 the *normal* exit path, feeding the unseeded-cursor wedge. The deadline was not raised in this PR.
- With the depot's 16-account SVM cap deleted, `MAX_ENVELOPE_BYTES = 65'536` permits ~500 small SPL swap-remits across distinct reserves (~5 attestations per 16-account round → 100+ rounds). After 64 rounds the loop falls through with **no log and no error**, returns `last_sig` as success, and strands the cursor mid-envelope with nothing scheduled to resume it (`_last_outbound_epoch` advances, so this operator never re-drives).

At minimum: log/throw on round exhaustion, and size `MAX_TERMINAL_ROUNDS` + the delivery deadline against the depot-side worst case now that the depot no longer bounds it.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -924,27 +915,76 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+      // effect makes progress rather than wedging the epoch (its surplus
+      // accounts log-and-skip on-chain, which the depot re-drives).
+      std::vector<fc::network::solana::account_meta> batch_accounts;
+      uint32_t batch_end = settled;
+      while (batch_end < total_attestations) {
+         auto candidate = batch_accounts;
+         for (const auto& meta : per_attestation[batch_end]) {
+            outpost_solana_client_detail::record_terminal_account(candidate, meta.key, meta.is_writable);
+         }
+         if (candidate.size() > MAX_TERMINAL_DYNAMIC_ACCOUNTS && batch_end > settled) break;
```
heifner: **The always-take-one rule's justification is wrong, and the 16-account budget lost its measurement test.**

The comment above ("its surplus accounts log-and-skip on-chain") is false twice: (a) an oversized single attestation's accounts are all *included* — nothing is surplus/omitted; the real failure is `transaction::serialize`'s 1232-byte packet assert (`libfc solana_types.cpp:309-313`), which throws locally on every attempt — the deterministic packer rebuilds the identical oversized tx each 15 s retry, i.e. exactly the "wedging the epoch" the comment claims is avoided; (b) per this PR's own contract, a genuinely missing account now *aborts* the terminal call, it does not log-and-skip.

Compounding: this PR deletes the SEC-94 fixture test that measured a terminal `epoch_in` with 16 dynamic accounts against the packet limit, so `MAX_TERMINAL_DYNAMIC_ACCOUNTS = 16` (line 48) is now an unmeasured hand estimate — the budget counts only the extras, never the ~13 IDL-declared accounts or the compute-budget pre-ix. The new `epoch_in_full_data_chunk_fits_packet_limit` test covers only the data-chunk shape; please add the terminal-shape equivalent (static accounts + 16 extras ≤ 1232), which is the same failure class that test was added to prevent.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -924,27 +915,76 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
-   last_sig = _program_client->epoch_in(
-      epoch_index,
-      total_chunks,
-      total_chunks,
-      static_cast<uint32_t>(total),
-      {},
-      std::move(terminal_accounts));
-   ilog("outpost_solana_client[{}]: epoch_in terminal finalize sent epoch={} chunk={}/{} bytes=0 extras={} sig={}",
-        to_string(), epoch_index, total_chunks, total_chunks, terminal_extra_count, last_sig);
+   // Terminal calls, looped until the on-chain cursor drains. Each call
```
heifner: **Design conflicts with the standing `outpost-remit-is-immediate` rule — the rule/spec must change in the same coordinated landing.**

`wire-platform-manifest/.claude/rules/outpost-remit-is-immediate.md`: *"every remit-class attestation in that envelope MUST trigger its on-chain transfer inline, in the same tx that processes the envelope. There is no concept of 'pending remit,' 'queued release,' 'delayed payout,' or any kind of waiting state on the outpost... the dispatch loop IS the release."*

The resumable cursor is precisely a queued release: after the consensus-tipping tx, undispatched remits sit in a pending state until later terminal calls that may never come (see the unseeded-cursor and round-exhaustion findings on this file). If the cursor design is the intended direction — and the depot-side rationale in `buildenv` reads that way — the rule doc and wire-solana's normative `opp-outpost-technical-spec.md` (lines 150, 838-842, 882, and 915, which still codifies log-and-skip, 672-byte chunks, and the 5-arg `epoch_in`) need updating in the same landing, or every diagnosis playbook built on them will misread production.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1636,16 +1568,6 @@ void msgch::queueout(uint64_t chain_code,
          has_auth(RESERV_ACCOUNT) || has_auth(get_self()),
          "queueout: caller not authorized to queue outbound attestations");
 
-   {
-      sysio::chains::chains_t chains_tbl(CHAINS_ACCOUNT);
```
heifner, line 1654: **`queueout` lost its only chain_code registration check — an authorized caller's typo now creates a permanent, unreachable READY row.**

All 8 production call sites derive their code from `sysio.chains` (epoch:686,729; opreg:634,673,745; uwrit:349,561; reserv:153), so this is not an attack vector — but it is an ops footgun: a governance msig holding one of the authorized authorities pushing `queueout` directly with an unregistered code used to abort with `"chain_code is not registered"`; now it succeeds silently. `epoch::advance` only fans `buildenv` to registered active outposts, the PROCESSED-drain (lines 1853-1860) skips mismatched chain codes, and no prune action exists — the row is permanent `sysio`-pool RAM, re-walked by the Phase-1 `bystatus` scan on every `buildenv` for every outpost, every epoch.

Related cleanup if the removal stands: `sysio.reserv.cpp:93-95` ("`sysio.msgch::queueout` fails loudly for direct callers"), `sysio.msgch_tests.cpp:164-170`, and `sysio.reserv_tests.cpp:411-414` all still describe the deleted guard — a future reader will conclude msgch is the backstop and relax the reserv-side guards. Also `buildenv`'s chains fetch (new lines 1648-1651) now asserts with the bare default `"key not found"` — worth passing the message overload.

plugins/outpost_solana_client_plugin/test/test_outpost_solana_client_plugin.cpp
```diff
@@ -946,6 +837,56 @@ BOOST_AUTO_TEST_CASE(extract_pubkeys_deposit_revert) try {
...
+/// order, counting attestations that need no effect accounts.
+///
+/// This is what makes resumable dispatch safe: the program settles
+/// `[dispatched_count, dispatched_count + dispatch_limit)` over that same
+/// sequence, so the relay sizes each terminal call's `dispatch_limit` to the
+/// attestations whose accounts it is carrying. If an index drifted, a batch
+/// would claim attestations whose accounts are absent -- and a handler with a
+/// missing account log-and-skips (a SUCCESSFUL no-op) while the cursor
+/// advances past it, silently dropping a remit or a slash with no retry path.
+BOOST_AUTO_TEST_CASE(extract_effects_indices_track_dispatch_order) try {
```
heifner, line 947: **The resumable-dispatch state machine itself ships untested, and the five whole-envelope extractors are now dead public surface.**

The new tests cover `extract_inbound_effects` / `count_inbound_attestations` only. Nothing drives the packing loop, `dispatch_limit` sizing, cursor resume, the `consensus_reached`/stall breaks, or the 16-account budget — the two confirmed bugs in that loop (the zero-attestation `FC_ASSERT` and the unseeded cursor) both live in the untested region and both contradict comments in the same files, which is why they survived.

Separately: `extract_inbound_recipient_pubkeys`, `extract_inbound_swap_remit_spl_targets`, `extract_inbound_swap_revert_spl_targets`, `extract_inbound_swap_remit_reserve_seeds`, and `extract_inbound_reserve_create_cancelled_seeds` now have zero production callers (`deliver_outbound_envelope` uses `extract_inbound_effects` directly) but remain exported from the public header purely for tests — and the comments at `outpost_solana_client.cpp:525/539` mislabel the two SPL-target views as "deduplicating", which they are not. Either test the loop through a seam and delete/relocate the views, or the dead surface plus wrong docs will drift.

heifner (commented):
Re-review at `b349fe1414` (rebased onto `dade24c613`).

This is not a patch-up of the version I reviewed — it is a different architecture. `epoch_in` is now delivery-only (the `dispatch_limit` argument is gone entirely) and settlement moved to a new permissionless `dispatch_attestations` instruction driven from the on-chain cursor. That restructuring is what closes the prior findings, and it closes them at the root rather than at the symptom. All ten are addressed.

---

## Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | Retry re-uploads all chunks on every drain failure | **Resolved by design.** The terminal `epoch_in` is fixed-size with no manifest, and the inline drain is try/caught (`outpost_solana_client.cpp:1095-1107`) so settlement can no longer fail the delivery. Re-upload now requires an actual chunk / terminal-call failure — the same surface as the base. |
| 2 | `settled` never seeded from the cursor → multi-round wedge | **Fixed.** `drive_dispatch_rounds` reads the cursor before packing (`:597`, `:612`). Pinned by `drive_dispatch_resumes_from_nonzero_cursor`. |
| 3 | Zero-attestation / decode-failure path never records the delivery | **Fixed, and better than asked.** Delivery records regardless of decode; decode failure now throws inside `drain_dispatch` where it belongs (`:993-1001`); an empty envelope gets one clamped close crank, since the program's completion block is the only place `next_epoch_index` advances. The matching program-side regression test (`dispatch_window(0, 0, 5) == (0, 0)`) is the right place for that invariant to live. |
| 4 | Cursor read at `confirmed` against a write confirmed at `processed` | **Fixed.** The read is pinned to `commitment_t::processed` (`:748`). |
| 5 | Silent round exhaustion; budget not sized against the depot's worst case | **Fixed.** Exhaustion `elog`s (`:689-692`), the budget went 64 → 128 with a stated sizing rationale, and resumption is now real via drain-then-read. |
| 6 | `MAX_TERMINAL_DYNAMIC_ACCOUNTS = 16` left unmeasured after the SEC-94 fixture was deleted | **Fixed.** `dispatch_attestations_full_manifest_fits_packet_limit` measures the real shape — IDL statics, fee payer, the heap-frame pre-instruction, and 16 distinct writable extras. |
| 7 | Conflicts with the standing `outpost-remit-is-immediate` rule | **Addressed; one item outstanding** — see below. |
| 8 | `queueout` lost its only `chain_code` registration check | **Fixed.** The check is back with its original message, and `buildenv` / `dispatch_attestation` now carry explicit assert messages instead of the bare `key not found`. |
| 9 | The dispatch state machine shipped untested; five dead extractor views | **Fixed.** The loop is factored over its RPC seams with 12 dedicated tests, and the dead views are deleted. |
| 10 | Lockstep break against the manifest-pinned program | **Fixed.** wire-solana#419 now carries `MAX_CHUNK_BYTES = 668` and `dispatch_attestations`. |

The e2e evidence checks out independently: run `31550041513` succeeded with `BRANCH_WIRE_SYSIO` **and** `BRANCH_WIRE_SOLANA` both set to `feat/resumable-opp-dispatch` and was dispatched after the head commit — the correct paired combination for a coupled change.

---

## New findings

Four on the reworked design, anchored inline. Two of them (the memoize placement and the eager manifest build) are worth fixing before merge; the other two are cleanups.

---

## Please remove `docs/superpowers/plans/2026-08-11-contained-resumable-dispatch-relay.md`

That file is session scaffolding rather than repository documentation: a checkbox task list addressed *"For agentic workers"*, pinning an absolute local path (`/home/valthon/wire-platform/wire-sysio`) and a commit-by-commit workflow that is already history. It will read as normative to the next person who greps `docs/` and finds it.

`docs/opp-two-phase-dispatch-design.md` is the genuine design doc and should absolutely stay — it is the artifact that makes this change reviewable, and its `outpost-remit-is-immediate` section is doing real work. `docs/superpowers/specs/2026-08-11-contained-resumable-dispatch-relay-design.md` is a placement redirect for the plan; if the plan goes, it is worth asking whether the redirect still has a reader.

---

## Outstanding landing item

`wire-platform-manifest/.claude/rules/outpost-remit-is-immediate.md` is still unedited, and continues to state: *"There is no concept of 'pending remit,' 'queued release,' … the dispatch loop IS the release."*

The design doc handles the tension honestly — there is an explicit *"Relationship to `outpost-remit-is-immediate`"* section that concedes the design fails the rule's letter and argues its intent, with the rule update listed as landing step 4 — and wire-solana#419 has already updated `opp-outpost-technical-spec.md` (+106/−57). Only the manifest rule is left. Worth landing in the same window, since every diagnosis playbook built on that rule will otherwise misread production.

---

**Verdict:** the architecture change is the right call, and the tests now pin the exact region where both confirmed bugs lived. Nothing here blocks the coordinated merge with wire-solana#419.

heifner (dismissed):
Approving. All fourteen findings from the previous rounds are addressed, and the two-phase split is the right resolution rather than a patch — the instruction reads its own cursor and `dispatch_limit` bounds only how far, which makes the unseeded-cursor defect unrepresentable. `drive_dispatch_rounds` factored over its RPC touchpoints turns the previously untestable region into the best-covered part of the file, and `dispatch_attestations_full_manifest_fits_packet_limit` makes `MAX_TERMINAL_DYNAMIC_ACCOUNTS` a measured budget again.

Non-blocking follow-ups:

**Generator templates ship untested.** The new commit's payload is four length-delimited templates in `field.ts`; the tests added cover only the emitted runtime helpers, which were already bounds-checked (just the `checked_add` route is new there). `Field.test.ts` touches nothing that emits a decode arm, and no fixture proto declares a `map<`, so `genMapDecode` — two of the four edits — has no coverage at all, jest or cargo. A map field in `sample_types.proto` plus a truncated-prefix and a `u64::MAX`-prefix case in the fixture crate would close it.

**The guard reaches no deployed program yet.** wire-solana pins `wire-opp-solana-models` at exactly 1.0.38; only the platform e2e overrides it with the generated tree, which is why the panic-pin evidence is real. The published crate still has the unguarded decoders, so the landing sequence needs a republish and pin bump alongside wire-solana#419.

**`outpost-remit-is-immediate` is unreconciled.** The companion updated `opp-outpost-technical-spec.md`, but the standing rule in wire-platform-manifest still says there is no queued release and that the dispatch loop is the release. The design doc names the file; it needs to land in the same sequence.

**Three envelope parses per drain, one above the consensus gate.** The decode probe, `count_inbound_attestations` and `extract_inbound_effects` each parse up to 64 KiB, on every inbound tick while the outbound envelope is missing. The probe's placement is deliberate, but consensus-not-tipped means the same bytes fail identically next tick, so every non-tipping operator pays a parse for a result that cannot matter yet. One parse can feed both the count and the effects.

**The already-drained early return did not land.** `drain_dispatch` reads progress at the gate but uses only `consensus_reached`, so a drained epoch still walks every effect and does two account reads per unique reserve before `drive_dispatch_rounds` returns on its cursor check. Composes with the above — the count plus the progress already read gives the early return for free.

**Minor.** `SWAP_REMIT_ATTESTATION_TYPE` and `SOL_OUTPOST_ID` are now unused in `sysio.msgch_tests.cpp`, with a stray double blank line where `emitted_attestation_count` was. `drive_dispatch_rounds` does two progress reads per round where the next round's could serve the previous round's stall check. The `.gitignore` commit says in its own message that it is unrelated to the branch, which argues for its own PR.

heifner (approved):
Approving at `6db0483ee`.

Re-reviewed the two commits added since the last round — `d1159239` (harden the relay against epoch churn) and `6db0483e` (gate the stub IDL fixture against the generated `liqsol_core` IDL) — plus the rebase onto current master. The dispatch state machine reads correctly: no defect in `drive_dispatch_rounds` packing or cursor resume, in the degrade/throw split in `reserve_info_for_codes`, or in the new boot asserts. On the generator commit, the `checked_add` guards plus the `end > data.len()` bounds check now present on all four templates close the slice-panic route.

Two things I raised in re-review and am withdrawing:

- **`sysio.msgch.wasm` grew 155,599 → 184,678 B on a net-deleting source commit.** Different CDT options produce different wasm sizes; the larger artifact is intended. No action.
- **Boot asserts hard-fail against the currently deployed program.** `has_idl("dispatch_attestations")` and the `dispatched_count` shape check both fail against `wire-solana@next`, but fail-loud is the right design here and the merge order (wire-solana#419 first, then this) is understood.

Nothing else blocks. The items below are follow-ups, not change requests.

## `dispatch_limit` is bounded by accounts, never by compute

The greedy packer's only exit is the account union exceeding `MAX_TERMINAL_DYNAMIC_ACCOUNTS`:

```cpp
if (candidate.size() > MAX_TERMINAL_DYNAMIC_ACCOUNTS && batch_end > settled) break;
```

Only four attestation types produce an `inbound_effect` — `OPERATOR_ACTION` (deposit/withdraw), `DEPOSIT_REVERT`, `SWAP_REMIT`, `SWAP_REVERT`. Everything else contributes zero accounts, so `candidate.size()` never grows and the window extends to the end of the envelope. `send_dispatch_attestations` requests a heap frame but deliberately no `set_compute_unit_limit`, so the transaction runs under the runtime default. Nor does packet size bound it: a window of N no-account attestations serializes to the same small instruction regardless of N.

In practice this is hard to reach — the zero-account types are state mirrors (`OPERATORS`, `BATCH_OPERATOR_GROUPS`) that ride one per envelope, and the effect-yielding types bound the window normally. The realistic shape is an epoch dominated by `OperatorAction(SLASH)`, which falls in the no-effect arm. Worth a ceiling anyway, because the failure mode has no self-heal: on CU exhaustion `send_dispatch` throws, the cursor does not advance, and every later tick repacks the byte-identical window from the same cursor. A constant beside `MAX_TERMINAL_DYNAMIC_ACCOUNTS`, or a `set_compute_unit_limit`, closes it. This one had an implicit bound from the depot-side terminal budget that this PR removes, so it is worth not leaving open indefinitely.

## `solana_idl_stub_consistency_test` fails rather than skips on a stale sibling

The script's skip path covers an *absent* `target/idl/liqsol_core.json`, but a *present but older* one is a hard exit 1. Against `wire-solana@next` (`5eac77cb`) there are three divergences: no `dispatch_attestations` instruction, no `EpochDeliveries.dispatched_count`, and `epoch_in` declares 12 accounts against the stub's 7. This resolves itself once #419 is the manifest default, and CI without a built sibling skips — so the only cost is a false red for anyone holding a stale build. Gating the comparison on the sibling declaring `dispatch_attestations` (skip-with-reason otherwise) would remove that window.

Related: `tests/CMakeLists.txt:290` always passes `--solana-path ${CMAKE_SOURCE_DIR}/../wire-solana`, and `resolve_solana_path` gives the argument precedence, so the documented `WIRE_SOLANA_PATH` override never fires under ctest. In any workspace where wire-solana is not the literal sibling the gate takes the skip path and is permanently green — the silent no-op the docstring's precedence rule exists to prevent.

## Carried over, still open

- **`drain_dispatch` has no already-drained early return.** It reads progress only to gate on `consensus_reached`, discards `dispatched_count`, then builds the full manifest — one `get_account_info` per distinct reserve — before `drive_dispatch_rounds` returns immediately on its cursor check. Re-entered on every inbound tick while the delivered envelope matches the epoch.
- **Three full parses of the same envelope per drain** — the decode probe, `extract_inbound_effects`, and `count_inbound_attestations`. One decode can feed all three.
- **Two cursor reads per dispatch round.** The post-send `read_progress()` is overwritten by the next round's top-of-round read with the same value; only the consensus re-check needs it, and consensus cannot un-tip.
- **Dead constants in `sysio.msgch_tests.cpp`** — `SOL_OUTPOST_ID` and `SWAP_REMIT_ATTESTATION_TYPE` lost their last users when the two SVM budget tests were deleted. `UNCOVERED_TEST_ATTESTATION_TYPE` and `emitted_attestation_count` were removed correctly; these two were missed. Unused internal-linkage `constexpr` trips `-Wunused-const-variable` under `-Wall`.

## Landing checklist

`wire-platform-manifest/.claude/rules/outpost-remit-is-immediate.md` is still unedited across three rounds now. It states remits fire "inline, in the same tx that processes the envelope" and explicitly bans a two-step shape; this PR makes settlement a separate permissionless instruction. `docs/opp-two-phase-dispatch-design.md` argues the invariant is preserved in spirit and names the rule update as a landing step, and the companion already updated `opp-outpost-technical-spec.md` — only the manifest rule is left. It lives in another repo, so it is the item most likely to be dropped once this merges, and every stall-diagnosis playbook is built on the sentence it contradicts.

PR #553, opened 2026-08-08
feat(outpost_solana): declare collateral-settlement accounts on the terminal manifest (SOL-375/379/380)

valthon:
> [!IMPORTANT]
> **Part of a three-repo set — all must merge together:**
> - wire-solana **https://github.com/Wire-Network/wire-solana/pull/409** — the program change (`CollateralPosition` PDA, asset-identity enforcement)
> - wire-sysio **this PR** — the relay's terminal account manifest
> - wire-tools-ts **https://github.com/Wire-Network/wire-tools-ts/pull/57** — harness + flow coverage
>
> `dispatch_attestations` matches `remaining_accounts` by pubkey. If the program lands
> without this relay change, every collateral-settling handler aborts for want of an
> account the relay never declared — and because the envelope bytes are consensus-pinned,
> that abort recurs on every retry and wedges the epoch.

Relay side of the SOL-375/379/380 collateral work.

wire-solana #409 replaced the bounded collateral `Vec` with a per-`(operator, token_code)`
`CollateralPosition` PDA and made the terminal handlers settle in the asset actually
escrowed. Both mean the zero-data terminal `dispatch_attestations` call must declare
accounts it never carried before, for every `WITHDRAW_REMIT` / `SLASH` / `DEPOSIT_REVERT`.

## What this does

- Extends `build_dispatch_manifests` to declare each collateral-settling attestation's
  accounts: the `CollateralPosition` PDA, and for SPL custody the `collateral_vault` PDA,
  the destination ATA, and the SPL token program. `SLASH` is surfaced deliberately (a
  dropped seizure has no return attestation to re-drive it), settling into the
  `reserve_aggregate`'s ATA; `WITHDRAW_REMIT` into the operator's; the SPL `DEPOSIT_REVERT`
  refunds the depositor's ATA.
- Byte-exact PDA derivations for `collateral_position` and `collateral_vault`, mirroring
  `opp_states.rs`.

## Custody comes from the position, never the config

The on-chain handlers branch on the `custody_mint` **pinned into each `CollateralPosition`
at first deposit**. The relay resolves custody the same way — reading it off the
`CollateralPosition` account, keyed by `(operator, token_code)` — never from the mutable
`OutpostConfig` token maps. Resolving from config would let one `token_code` back positions
of disagreeing custody (operator A native, operator B SPL), or an admin re-point between
deposit and settlement, ship a manifest missing the SPL accounts → a permanent, correlated
epoch wedge. This mirrors the reserve path exactly: `get_account_info` sits outside the
decode `try` so an RPC/deadline error propagates; an absent position degrades to a partial
manifest; a present-but-unreadable one fails the tick loudly with the cursor untouched.

A boot-time `assert_collateral_position_shape`, under the batch-operator role gate, fails
loudly on a drifted `CollateralPosition` IDL rather than silently misreading a live
position's custody on the first drain.

## Verification

- `test_outpost_solana_client_plugin` — green (includes a regression test proving two
  operators sharing one `token_code` with disagreeing custody get different manifests).
- `plugin_test` builds clean.
- Reviewed adversarially (correctness + security + rules) after a full rebase onto
  `origin/master`; findings addressed.

https://claude.ai/code/session_01VqAoaLdCBqgV2vXgDCGhgY

### DIFF
--- a/plugins/outpost_solana_client_plugin/include/sysio/outpost_solana_client_plugin/outpost_solana_client.hpp
+++ b/plugins/outpost_solana_client_plugin/include/sysio/outpost_solana_client_plugin/outpost_solana_client.hpp
@@ -185,6 +185,27 @@ std::vector<char> decode_latest_envelope_account(opp_solana_outpost_client&  pro
 ///         is absent or disagrees with what the cursor read decodes.
 void assert_epoch_deliveries_shape(const fc::network::solana::idl::program& program);
 
+/// Assert that the loaded IDL's `CollateralPosition` declaration carries the
+/// four fields the on-chain settlement path binds together: `operator` and
+/// `custody_mint` (pubkeys), plus `token_code` and `amount` (u64), in either
+/// IDL field home.
+///
+/// The relay itself decodes only `custody_mint`; the other three are a
+/// DELIBERATE drift canary — a program-side rename of any of the four fails
+/// loudly at boot rather than surfacing later as a live position this relay
+/// silently half-understands.
+///
+/// Called at construction for the batch-operator role. IDL drift is the
+/// realistic reason a live, program-readable position becomes unreadable to
+/// THIS relay; without its pinned custody mint the manifest cannot include the
+/// effect accounts the program's real branch requires, so the dispatch window
+/// would abort and every retry would wedge on the same cursor position.
+///
+/// @param program  the program's loaded Anchor IDL.
+/// @throws fc::exception if the account or any of the four fields is absent,
+///         or a field has a type the manifest builder cannot interpret.
+void assert_collateral_position_shape(const fc::network::solana::idl::program& program);
+
 /// Assert that the loaded IDL's `Reserve` declaration carries the three fields
 /// the terminal manifest resolves from it: `creator` and `custody_mint`
 /// (pubkeys) and `custody_decimals` (u8), in either IDL field home.
@@ -323,6 +344,26 @@ std::string drive_dispatch_rounds(
                                                                       send_dispatch,
    const std::string&                                                 log_label);
 
+/// Custody binding read from the account the on-chain settlement handler
+/// branches on. Deliberately mint-only: `CollateralPosition` carries no
+/// decimals, and the manifest path branches on the mint alone — a decimals
+/// field here could only ever hold a made-up value (contrast
+/// `reserve_terminal_info`, whose `custody_decimals` is a real pinned fact).
+struct token_custody_info {
+   /// SPL mint for the token, or the all-zero system-program key when the
+   /// token is native lamports (the on-chain zero-marker convention).
+   fc::network::solana::solana_public_key mint;
+};
+
+/// Reads ONE `(operator, token_code)` position's custody binding from its
+/// `CollateralPosition` PDA, exactly where the on-chain handler resolves it.
+/// Absent positions degrade because the program log-and-skips them; a present
+/// but unreadable position throws so the relay never submits a manifest that
+/// is guaranteed to abort.
+using collateral_custody_reader =
+   std::function<std::optional<token_custody_info>(
+      const fc::network::solana::solana_public_key& operator_key, uint64_t token_code)>;
+
 } // namespace outpost_solana_client_detail
 
 /**
@@ -434,6 +475,19 @@ private:
       uint32_t                                       dispatch_limit,
       std::vector<fc::network::solana::account_meta> extra_remaining_accounts);
 
+   /// Resolve one per-`(operator, token_code)` collateral position's pinned
+   /// custody mint from the `CollateralPosition` PDA the on-chain handlers
+   /// themselves branch on.
+   ///
+   /// An ABSENT or EMPTY position returns empty because the program
+   /// log-and-skips an uninitialized account. A position that EXISTS but this
+   /// relay cannot decode THROWS after logging: the program takes its real
+   /// custody branch, so degrading would omit required effect accounts and
+   /// permanently wedge the unadvanced dispatch cursor.
+   std::optional<outpost_solana_client_detail::token_custody_info>
+   collateral_position_custody(
+      const fc::network::solana::solana_public_key& operator_key, uint64_t token_code);
+
    solana_client_entry_ptr                       _entry;
    fc::network::solana::solana_public_key        _program_id;
    std::shared_ptr<opp_solana_outpost_client>    _program_client;
@@ -491,12 +545,51 @@ derive_reserve_vault_pda(const fc::network::solana::solana_public_key& program_i
                          uint64_t token_code,
                          uint64_t reserve_code);
 
+/// Derive the per-`(operator, token_code)` `CollateralPosition` PDA (SOL-379):
+/// seeds `["collateral_position", operator.as_ref(), token_code.to_le_bytes()]`.
+/// Byte-exact mirror of the program's `#[account(seeds = ...)]` declaration.
+///
+/// Exported so the manifest builder and its tests derive through ONE
+/// implementation rather than re-spelling the seed list.
+fc::network::solana::solana_public_key
+derive_collateral_position_pda(const fc::network::solana::solana_public_key& program_id,
+                               const fc::network::solana::solana_public_key& operator_key,
+                               uint64_t token_code);
+
+/// Derive the per-`token_code` `collateral_vault` PDA: seeds
+/// `["collateral_vault", token_code.to_le_bytes()]`. The vault holds the
+/// token's SPL collateral custody; native collateral settles straight out of
+/// the named `vault`, so this account only rides an SPL manifest.
+fc::network::solana::solana_public_key
+derive_collateral_vault_pda(const fc::network::solana::solana_public_key& program_id,
+                            uint64_t token_code);
+
 /// Which family of effect accounts one inbound attestation needs. The relay
 /// derives the concrete metas per shape; the on-chain handler resolves them
 /// out of `remaining_accounts` by pubkey.
+///
+/// The collateral-settling shapes (`withdraw_remit`, `slash`,
+/// `deposit_revert`) exist because SOL-379 replaced the bounded collateral
+/// `Vec` on `OperatorRegistry` with a per-`(operator, token_code)`
+/// `CollateralPosition` PDA, and SOL-380 made the handlers settle in the
+/// asset the position actually escrows — so their manifests must declare the
+/// position PDA (seeds `[COLLATERAL_POSITION_SEED, operator.as_ref(),
+/// &token_code.to_le_bytes()]`) and, for SPL custody, the
+/// `[COLLATERAL_VAULT_SEED, &token_code.to_le_bytes()]` vault PDA, the
+/// destination ATA and the SPL token program.
 enum class effect_shape {
-   /// Pays a native-SOL recipient directly (WITHDRAW_REMIT, DEPOSIT_REVERT).
-   native_payee,
+   /// OPERATOR_ACTION(WITHDRAW_REMIT): pays the operator (natively out of the
+   /// named `vault`, or into their canonical ATA under SPL custody) and
+   /// debits their `CollateralPosition` PDA.
+   withdraw_remit,
+   /// OPERATOR_ACTION(SLASH): debits the operator's `CollateralPosition` PDA
+   /// and routes the seizure into the named `reserve_aggregate` (native) or
+   /// its canonical ATA via the collateral vault + token program (SPL).
+   slash,
+   /// OPERATOR_ACTION(DEPOSIT_REVERT): refunds the depositor (natively out of
+   /// the named `vault`, or into their canonical ATA under SPL custody) and
+   /// debits their `CollateralPosition` PDA.
+   deposit_revert,
    /// SWAP_REMIT: Reserve PDA, plus vault + recipient ATA + token program
    /// when the reserve's custody mint is SPL rather than native.
    swap_remit,
@@ -521,10 +614,17 @@ enum class effect_shape {
 struct inbound_effect {
    size_t                                                attestation_index;
    effect_shape                                          shape;
-   /// Native payee / SWAP_REMIT recipient / SWAP_REVERT depositor.
+   /// WITHDRAW_REMIT operator / DEPOSIT_REVERT depositor / SWAP_REMIT
+   /// recipient / SWAP_REVERT depositor. For `slash` it is the SLASHED
+   /// operator — it keys the `CollateralPosition` PDA and the destination
+   /// ATA owner lookup but is never itself paid.
    std::optional<fc::network::solana::solana_public_key> recipient;
    /// Set for every reserve-backed shape.
    std::optional<reserve_pda_seeds>                      reserve;
+   /// Set for every collateral-settling shape (`withdraw_remit`, `slash`,
+   /// `deposit_revert`): the `token_code` keying the `CollateralPosition`
+   /// PDA and its pinned custody lookup.
+   std::optional<uint64_t>                               collateral_token_code;
 };
 
 /// Walk `envelope_bytes` ONCE and return every attestation that needs effect
@@ -578,12 +678,26 @@ uint32_t count_inbound_attestations(const std::vector<char>& envelope_bytes);
 ///     read propagates untouched — it means a manifest the program would abort
 ///     on, and shipping one would wedge the epoch rather than delay it.
 ///
+///   * Collateral-settling shapes (`withdraw_remit`, `slash`, `deposit_revert`)
+///     resolve custody through `read_collateral_custody` — called at most ONCE
+///     per distinct `(operator, token_code)` position, memoised like the
+///     reserve reads. A degraded (empty) custody read costs that attestation
+///     only its SPL extras; the `CollateralPosition` PDA and (where owed) the
+///     recipient are still declared, matching the program's log-and-skip /
+///     abort-and-retry gates.
+///
 /// @param program_id            outpost program id, for PDA derivation.
 /// @param effects               account-needing attestations, in dispatch order.
 /// @param total_attestations    the envelope's attestation total (the cursor's
 ///                              denominator) -- the size of the result.
 /// @param throw_if_past_deadline  throws once the caller's deadline passes.
 /// @param read_reserve_info     reads one `Reserve` record (may degrade).
+/// @param read_collateral_custody  reads one `(operator, token_code)`
+///                              position's pinned custody (may degrade only
+///                              when the position is absent or empty).
+/// @param reserve_aggregate     the named `reserve_aggregate` account — the
+///                              destination whose ATA receives an SPL slash
+///                              seizure.
 /// @param log_label             client identity for log lines.
 /// @return one manifest per attestation, in dispatch order.
 std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manifests(
@@ -592,6 +706,8 @@ std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manif
    uint32_t                                      total_attestations,
    const std::function<void()>&                  throw_if_past_deadline,
    const reserve_info_reader&                    read_reserve_info,
+   const collateral_custody_reader&              read_collateral_custody,
+   const fc::network::solana::solana_public_key& reserve_aggregate,
    const std::string&                            log_label);
 
 } // namespace outpost_solana_client_detail
--- a/plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
+++ b/plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
@@ -41,6 +41,14 @@ constexpr std::string_view OP_UW_COMMIT   = "uw_commit:commit_underwrite";
 constexpr std::string_view EPOCH_DELIVERIES_SEED = "epoch_deliveries";
 constexpr std::string_view ENVELOPE_CHUNKS_SEED  = "envelope_chunks";
 
+/// Anchor seed literals for the collateral-settlement PDAs. Byte-exact
+/// mirrors of the program's `COLLATERAL_POSITION_SEED` /
+/// `COLLATERAL_VAULT_SEED` constants (wire-solana `opp_states.rs`) -- a
+/// one-character drift derives a well-formed WRONG PDA that only surfaces at
+/// runtime as `EffectAccountMissing`, holding the dispatch cursor.
+constexpr std::string_view COLLATERAL_POSITION_SEED = "collateral_position";
+constexpr std::string_view COLLATERAL_VAULT_SEED    = "collateral_vault";
+
 /// The 4-byte little-endian seed encoding of a WIRE epoch index -- the exact
 /// bytes the program's `epoch_index.to_le_bytes()` seed component uses.
 std::vector<uint8_t> epoch_index_le_seed(uint32_t epoch_index) {
@@ -124,6 +132,17 @@ namespace reserve_account {
    constexpr auto field_custody_mint     = "custody_mint";
    constexpr auto field_custody_decimals = "custody_decimals";
 } // namespace reserve_account
+
+/// Field identifiers of the per-`(operator, token_code)` collateral position.
+/// Its custody mint is pinned at first deposit and is what the on-chain
+/// settlement handlers branch on.
+namespace collateral_position {
+   constexpr auto account_name       = "CollateralPosition";
+   constexpr auto field_operator     = "operator";
+   constexpr auto field_token_code   = "token_code";
+   constexpr auto field_custody_mint = "custody_mint";
+   constexpr auto field_amount       = "amount";
+} // namespace collateral_position
 
 } // anonymous namespace
 
@@ -267,6 +286,49 @@ void assert_epoch_deliveries_shape(const fc::network::solana::idl::program& prog
              "EpochDeliveries IDL missing '{}' field; the dispatch crank would resume from 0 on "
              "every tick and re-send settled windows forever",
              epoch_deliveries::field_dispatched_count);
+}
+
+/// Assert the loaded IDL declares `CollateralPosition` with the four fields
+/// its on-chain settlement path binds together. Full contract on the header
+/// declaration.
+void assert_collateral_position_shape(const fc::network::solana::idl::program& program) {
+   namespace idl = fc::network::solana::idl;
+   const auto& fields = declared_account_fields(program, collateral_position::account_name);
+
+   bool has_operator   = false;
+   bool has_token_code = false;
+   bool has_mint       = false;
+   bool has_amount     = false;
+   for (const auto& field : fields) {
+      const bool is_pubkey =
+         field.type.is_primitive() && field.type.primitive == idl::primitive_type::pubkey;
+      const bool is_u64 =
+         field.type.is_primitive() && field.type.primitive == idl::primitive_type::u64;
+      if (field.name == collateral_position::field_operator) {
+         FC_ASSERT(is_pubkey, "CollateralPosition '{}' must be declared pubkey, got '{}'",
+                   collateral_position::field_operator, describe_idl_type(field.type));
+         has_operator = true;
+      } else if (field.name == collateral_position::field_token_code) {
+         FC_ASSERT(is_u64, "CollateralPosition '{}' must be declared u64, got '{}'",
+                   collateral_position::field_token_code, describe_idl_type(field.type));
+         has_token_code = true;
+      } else if (field.name == collateral_position::field_custody_mint) {
+         FC_ASSERT(is_pubkey, "CollateralPosition '{}' must be declared pubkey, got '{}'",
+                   collateral_position::field_custody_mint, describe_idl_type(field.type));
+         has_mint = true;
+      } else if (field.name == collateral_position::field_amount) {
+         FC_ASSERT(is_u64, "CollateralPosition '{}' must be declared u64, got '{}'",
+                   collateral_position::field_amount, describe_idl_type(field.type));
+         has_amount = true;
+      }
+   }
+   FC_ASSERT(has_operator && has_token_code && has_mint && has_amount,
+             "CollateralPosition IDL missing '{}' / '{}' / '{}' / '{}' "
+             "(found {} / {} / {} / {}); a live collateral position's pinned custody cannot be "
+             "resolved safely without this declaration",
+             collateral_position::field_operator, collateral_position::field_token_code,
+             collateral_position::field_custody_mint, collateral_position::field_amount,
+             has_operator, has_token_code, has_mint, has_amount);
 }
 
 /// Assert the loaded IDL declares `Reserve` with the three fields the terminal
@@ -518,6 +580,11 @@ std::vector<uint8_t> u64_seed(uint64_t value) {
    return out;
 }
 
+/// Seed bytes of a `solana_public_key` for Anchor PDA derivation.
+std::vector<uint8_t> pubkey_seed(const fc::network::solana::solana_public_key& key) {
+   return std::vector<uint8_t>(key._data.begin(), key._data.end());
+}
+
 } // anonymous namespace (within outpost_solana_client_detail)
 
 /// Derive the per-reserve `Reserve` PDA. Full contract on the header
@@ -543,6 +610,30 @@ fc::network::solana::solana_public_key derive_reserve_vault_pda(
       {std::vector<uint8_t>{'r','e','s','e','r','v','e','_','v','a','u','l','t'},
        u64_seed(token_code),
        u64_seed(reserve_code)},
+      program_id).first;
+}
+
+/// Derive the per-`(operator, token_code)` `CollateralPosition` PDA. Full
+/// contract on the header declaration.
+fc::network::solana::solana_public_key derive_collateral_position_pda(
+   const fc::network::solana::solana_public_key& program_id,
+   const fc::network::solana::solana_public_key& operator_key,
+   uint64_t token_code) {
+   return fc::network::solana::system::find_program_address(
+      {std::vector<uint8_t>(COLLATERAL_POSITION_SEED.begin(), COLLATERAL_POSITION_SEED.end()),
+       pubkey_seed(operator_key),
+       u64_seed(token_code)},
+      program_id).first;
+}
+
+/// Derive the per-`token_code` `collateral_vault` PDA. Full contract on the
+/// header declaration.
+fc::network::solana::solana_public_key derive_collateral_vault_pda(
+   const fc::network::solana::solana_public_key& program_id,
+   uint64_t token_code) {
+   return fc::network::solana::system::find_program_address(
+      {std::vector<uint8_t>(COLLATERAL_VAULT_SEED.begin(), COLLATERAL_VAULT_SEED.end()),
+       u64_seed(token_code)},
       program_id).first;
 }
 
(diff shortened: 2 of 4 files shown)


plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -865,6 +996,67 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+      // The SPL deposit-revert branch is refused on-chain by design (the
+      // lamport-denominated penalty stub has no meaning in token units), and
+      // the native branch's depositor is already declared as a recipient —
+      // so this destination never needs vault / ATA / token-program extras.
+      if (target.destination
+          == outpost_solana_client_detail::collateral_settlement_destination::depositor_native_only) {
+         continue;
+      }
+
+      const auto custody_opt = collateral_custody(target.token_code);
```
heifner, line 1033: **Custody branch decided from the mutable `OutpostConfig` maps, but the on-chain handlers branch on the `CollateralPosition`'s *pinned* `custody_mint`.**

A position opened while `token_addresses_by_code[T]` held SPL mint `M` pins `custody_mint = M`. If the config row for `T` is later re-pointed (to native/zero or a new mint) — exactly the drift the program's pinning exists to survive (`opp_states.rs`: *"an admin re-point ... can never change how an already-open position settles"*) — the relay here resolves the NEW config row: it either takes the native `continue` branch (declares no vault/ATA/token-program) or derives the destination ATA for the wrong mint. `process_withdraw_remit` / `process_slash_action` then take the SPL branch for mint `M` and `require_remaining_account(collateral_vault / ATA(owner, M))` returns `Err(EffectAccountMissing)`, aborting the ENTIRE terminal `epoch_in`. Every redelivery rebuilds the same wrong manifest, so the envelope never lands and the epoch stalls.

Deeper fix: the relay already derives the position PDA — fetch the `CollateralPosition` account and use its pinned `custody_mint` (the same pinned-facts pattern the reserve path adopted per the comment above `terminal_accounts`), falling back to config only when the position does not exist yet.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -747,6 +852,32 @@ outpost_solana_client::reserve_info_for_codes(uint64_t token_code, uint64_t rese
    };
 }
 
+std::optional<outpost_solana_client_detail::token_custody_info>
+outpost_solana_client::collateral_custody_for_code(uint64_t token_code) {
+   // Deliberately non-throwing: a collateral token_code missing from the
+   // config maps must degrade to "omit the SPL extras" (the handler then
+   // log-and-skips that one settlement), never abort the envelope delivery.
+   try {
```
heifner, line 860: **The "degrade to on-chain log-and-skip" fallback is unachievable for an existing SPL position.**

If this config read fails transiently (RPC timeout/5xx, deadline), the `catch` swallows it, `custody_opt` is nullopt, and the manifest loop `continue`s without the vault/ATA/token-program extras. On-chain, `load_collateral_position` succeeds (position exists, `custody_mint != NATIVE`), so `resolve_collateral_vault_transfer`'s `require_remaining_account(collateral_vault)` returns `Err(EffectAccountMissing)` and the terminal `epoch_in` tx fails — the exact "abort the whole envelope delivery" outcome this function's contract says it avoids, only surfaced as an on-chain custom-program error instead of a diagnosable relay-side error.

Compare `reserve_info_for_codes`, which deliberately lets RPC exceptions propagate so the delivery fails locally and retries cleanly.

plugins/outpost_solana_client_plugin/include/sysio/outpost_solana_client_plugin/outpost_solana_client.hpp
```diff
@@ -315,6 +329,57 @@ extract_inbound_swap_remit_spl_targets(const std::vector<char>& envelope_bytes);
...
+/// A collateral-settling terminal effect: the `(operator, token_code)` pair
+/// whose `CollateralPosition` PDA the on-chain handler resolves out of
+/// `remaining_accounts`, plus where that settlement sends the funds.
+///
+/// SOL-379 replaced the bounded collateral `Vec` on `OperatorRegistry` with
+/// a per-`(operator, token_code)` PDA, and SOL-380 made the terminal
+/// handlers settle in the asset actually escrowed. Both changes mean the
+/// terminal manifest must now declare accounts it never carried before.
+///
+/// Omitting the position PDA is NOT merely a liveness problem: `SLASH`
```
heifner, line 359: **This rationale is inverted versus the companion program.** Omitting the position PDA does NOT "flip the operator's status and then silently SKIP the seizure (log-and-skip, never a failed envelope)": `require_remaining_account` ABORTS the dispatch (`mod.rs`: *"Log-and-skip is NOT an option here ... aborting dispatch"*), and Solana tx atomicity rolls the status flip back with it.

Same inversion appears in the extractor's decode-failure `wlog` ("a SLASH would flip status WITHOUT seizing collateral") and the manifest-block comment in `deliver_outbound_envelope`. During a real epoch stall, an operator following these comments will look for a slashed-but-unseized operator and a succeeding envelope, when the actual signature is a failing terminal `epoch_in` with `EffectAccountMissing` — misdirecting forensics on exactly the incident class the epoch-stall runbooks cover. The manifest declaration is still correct (it is what keeps the abort from happening) — only the stated why needs fixing.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -865,6 +996,67 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+      const auto settlement_owner =
+         target.destination
+               == outpost_solana_client_detail::collateral_settlement_destination::reserve_aggregate
+            ? reserve_aggregate_pda
+            : target.operator_key;
+      const auto before = terminal_accounts.size();
+      add_terminal_account(
+         outpost_solana_client_detail::derive_collateral_vault_pda(_program_id, target.token_code),
+         true);
+      add_terminal_account(
```
heifner, line 1052: **Nothing creates the `reserve_aggregate` ATA this line declares — every SPL slash seizure will silently stick.**

The on-chain SLASH SPL branch requires the destination ATA to be pre-created (`resolve_collateral_vault_transfer`: uninitialized dest → `Ok(None)` → "status flipped, funds stuck"). This PR declares `ATA(reserve_aggregate_pda, mint)`, but no deploy/bootstrap step in either repo creates it, and unlike the swap-revert block above the manifest carries no mint/ATA-program/system-program for on-chain creation (which the handler does not implement for this path anyway). SLASH rides one envelope with no retry attestation, so the depot's `sysio.reserv` credit and the outpost escrow diverge permanently until manual intervention. Needs a provisioning story (create the reserve_aggregate ATA per SPL collateral mint at deploy/config time) or an on-chain creation path.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -747,6 +852,32 @@ outpost_solana_client::reserve_info_for_codes(uint64_t token_code, uint64_t rese
...
+              "terminal manifest will omit SPL collateral accounts for token_code {}",
+              to_string(),
+              _program_client->config_pda.to_string(fc::yield_function_t{}),
+              token_code);
+         return std::nullopt;
+      }
+      const auto config_v =
+         _program_client->decode_account_info_data("OutpostConfig", config_info->data);
+      return outpost_solana_client_detail::resolve_token_custody(config_v.get_object(), token_code);
+   } catch (const fc::exception& e) {
```
heifner, line 873: **"Never throws" contract only catches `fc::exception`.** The RPC path (`execute_idempotent` → HTTP client → variant decode) can surface non-fc exceptions on rare paths (`std::exception` from the transport stack, `std::bad_alloc`); one such throw escapes this function, escapes the collateral loop, and fails `deliver_outbound_envelope` before the terminal call — the exact abort-the-delivery outcome the function was written to prevent. Add a `catch (const std::exception&)` arm (mirroring this handler) if the no-throw contract is kept.

heifner (commented):
Reviewed the collateral-settlement terminal-manifest change against the companion program branch (`wire-solana` `fix/SOL-375-379-380-opp-collateral`: `inbound.rs`, `mod.rs`, `opp_states.rs`, `epoch_in.rs`).

The account inventory itself checks out: I traced every branch (SLASH native/SPL, WITHDRAW_REMIT native/SPL including the rent-refund close, DEPOSIT_REVERT native and SPL-refused) — seed strings, `token_code` endianness, writable flags, and ATA owner/mint inputs all match the on-chain handlers; the position PDA is correctly declared unconditionally, and dedup/merge via `record_terminal_account` is sound.

10 findings inline. The top three are one cluster: the PR's degrade-gracefully premise ("omission degrades to on-chain log-and-skip") is inverted relative to the program it targets — `require_remaining_account` makes a missing-account manifest abort the terminal `epoch_in` atomically (`EffectAccountMissing`), which is the epoch-stall failure class. The remainder: an unprovisioned `reserve_aggregate` ATA that leaves SPL slash seizures stuck, an exception-type gap in the no-throw contract, a cross-repo lock-step hazard on the SPL deposit-revert skip, a per-token repeated config fetch, a duplicate `reserve_aggregate` PDA derivation, a comment-only cross-extractor coupling, and magic-literal seed strings.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -865,6 +996,67 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+      add_terminal_account(
+         outpost_solana_client_detail::derive_collateral_position_pda(
+            _program_id, target.operator_key, target.token_code),
+         true);
+
+      // The SPL deposit-revert branch is refused on-chain by design (the
+      // lamport-denominated penalty stub has no meaning in token units), and
+      // the native branch's depositor is already declared as a recipient —
+      // so this destination never needs vault / ATA / token-program extras.
+      if (target.destination
```
heifner, line 1028: **Cross-repo lock-step hazard.** This skip hard-codes today's on-chain behavior ("SPL deposit-revert is refused by design") with no version coupling. The program marks that refusal as temporary (*"re-attest once policy lands"* / a Wire-team coordination item): when the SPL revert-penalty policy lands on-chain, the upgraded handler's SPL branch will `require_remaining_account(collateral_vault/ATA)` while a not-yet-upgraded relay still `continue`s here — `EffectAccountMissing` aborts every terminal `epoch_in` carrying an SPL DEPOSIT_REVERT and stalls the epoch until the relay is upgraded in lock-step. At minimum, name the lock-step requirement here so the program-side change picks the relay up with it.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -747,6 +852,32 @@ outpost_solana_client::reserve_info_for_codes(uint64_t token_code, uint64_t rese
    };
 }
 
+std::optional<outpost_solana_client_detail::token_custody_info>
+outpost_solana_client::collateral_custody_for_code(uint64_t token_code) {
+   // Deliberately non-throwing: a collateral token_code missing from the
+   // config maps must degrade to "omit the SPL extras" (the handler then
+   // log-and-skips that one settlement), never abort the envelope delivery.
+   try {
+      const auto config_info = _entry->client->get_account_info(_program_client->config_pda);
```
heifner, line 861: **Same `OutpostConfig` account is fetched + IDL-decoded once per distinct collateral `token_code` per delivery** (and separately once per reserve in `reserve_info_for_codes`). An envelope settling across 4 token codes issues 4 identical `getAccountInfo` RPCs + 4 full decodes on the deadline-bounded delivery path; each extra RPC also widens the window for the transient-failure fallback above, and two mid-delivery reads can observe *different* config versions. Fetch + decode the config once per delivery and resolve every token code (collateral and reserve) from that single snapshot.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -378,6 +378,45 @@ fc::network::solana::solana_public_key derive_reserve_vault_pda(
...
+   return fc::network::solana::system::find_program_address(
+      {std::vector<uint8_t>{'c','o','l','l','a','t','e','r','a','l','_','v','a','u','l','t'},
+       u64_seed(token_code)},
+      program_id).first;
+}
+
+/// The singleton `reserve_aggregate` PDA — seeds `[RESERVE_AGGREGATE_SEED]`.
+/// It is a NAMED account on `epoch_in`, but the SPL slash branch settles into
+/// its canonical ATA, which the manifest must derive and declare.
+fc::network::solana::solana_public_key derive_reserve_aggregate_pda(
```
heifner, line 413: **Duplicate derivation:** `opp_solana_outpost_client::reserve_pda` is already derived from the identical `"reserve_aggregate"` seed and cached (see `outpost_solana_client_plugin.hpp`), and `_program_client` is in hand at the only call site. Two independent spellings of the same seed re-create the incident documented on that member (the `outpost_reserve` → `reserve_aggregate` mis-seed that made `epoch_in` account validation fail): a future seed rename now has two sites, and a miss here derives the slash-destination ATA for a wrong owner — a silently-stuck SPL slash rather than a loud failure. Use `_program_client->reserve_pda` and drop this helper.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -865,6 +996,67 @@ std::string outpost_solana_client::deliver_outbound_envelope(
...
+      if (target.destination
+          == outpost_solana_client_detail::collateral_settlement_destination::depositor_native_only) {
+         continue;
+      }
+
+      const auto custody_opt = collateral_custody(target.token_code);
+      if (!custody_opt.has_value()) continue;
+      const auto& custody = *custody_opt;
+      if (is_native_custody(custody.mint)) {
+         // Native settles out of the named `vault` account; WITHDRAW_REMIT's
```
heifner, line 1037: **Hidden cross-extractor coupling, asserted only in comments.** The native branch (and the SPL full-drain rent-refund path) rely on the *recipient sweep* (`extract_inbound_recipient_pubkeys`) having independently declared the operator/depositor wallet. Nothing ties the two account sources together — no shared code, and no test pins "manifest contains the op wallet for a WITHDRAW_REMIT collateral target". A future edit narrowing the recipient sweep (whose own doc already carves out SLASH) breaks collateral settlement with no compile-time or test signal. Either have this loop declare the wallet for its own destinations (`record_terminal_account` dedups) or add a test pinning the cross-extractor invariant.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -378,6 +378,45 @@ fc::network::solana::solana_public_key derive_reserve_vault_pda(
...
+}
+
+/// Per-`(operator, token_code)` CollateralPosition PDA (SOL-379) — seeds
+/// `[COLLATERAL_POSITION_SEED, operator, token_code_le]`.
+fc::network::solana::solana_public_key derive_collateral_position_pda(
+   const fc::network::solana::solana_public_key& program_id,
+   const fc::network::solana::solana_public_key& operator_key,
+   uint64_t token_code) {
+   return fc::network::solana::system::find_program_address(
+      {std::vector<uint8_t>{'c','o','l','l','a','t','e','r','a','l','_','p','o','s','i','t','i','o','n'},
```
heifner, line 393: **Magic-literal seed strings** (CLAUDE.md "No magic literals": every non-trivial string lives behind a named `constexpr`). The three new seeds — `collateral_position`, `collateral_vault`, `reserve_aggregate` — are wire-format constants that must byte-match `COLLATERAL_POSITION_SEED` / `COLLATERAL_VAULT_SEED` / `RESERVE_AGGREGATE_SEED` on the program side, yet are spelled as inline `{'c','o','l',...}` char sequences: unsearchable by name, and a one-character typo derives a wrong PDA that only fails at runtime as `EffectAccountMissing`. The pre-existing seeds follow the same pattern, but new code is where the rule applies — a file-local named constant per seed also gives the cross-repo grep anchor.

heifner (commented):
Re-review at `7b83b644ae`.

Before the findings, one structural note that changes how this PR should be read and merged.

## This PR is stacked on #552, but its base says `master`

`7b83b644ae`'s ancestry is `8f12916a1a` → `7c4c825599` → `b349fe1414` (all three of #552) → `7b83b644ae`. The base branch is still `master`, so:

- The diff GitHub renders is **2811 / 1084 across 13 files**. The change this PR actually makes is **326 / 75 across 3 files** — one commit, `7b83b644ae`. Everything else is #552, which I reviewed separately.
- Merging this PR into `master` would silently bring all of #552 with it, regardless of #552's own review state.

The same applies to the companion: wire-solana#409's base is `next`, while its head `51ac5e8d` sits directly on `0dbd6493` — wire-solana#419's head. So both halves of this change are stacked on both halves of the resumable-dispatch change, which is coherent and, I assume, deliberate; only the two base settings are out of step with it.

Retargeting wire-sysio#553 → `feat/resumable-opp-dispatch` and wire-solana#409 → `feat/resumable-opp-dispatch` makes each diff show only its own work and makes the merge order explicit. Everything below is scoped to `7b83b644ae` alone.

---

## Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | Custody branch read from the mutable `OutpostConfig` maps rather than the position's pinned `custody_mint` | **Not addressed** — and materially worse on this stack. See inline. |
| 2 | The "degrade to on-chain log-and-skip" fallback was unachievable for an existing SPL position | **Resolved by #552's architecture.** Delivery and settlement are now separate instructions and the drain resumes from the on-chain cursor, so a transient config-read failure genuinely does re-drive on a later tick. The rewritten rationale on `collateral_custody_for_code` is now accurate. |
| 3 | Inverted "log-and-skip / never a failed envelope" rationale in the header and extractor comments | **Fixed** in the source. Still live in the PR description — see below. |
| 4 | Nothing creates the `reserve_aggregate` ATA, so SPL slash seizures stick | **Not addressed.** Re-confirmed against wire-solana#409. See inline. |
| 5 | The "never throws" contract only catches `fc::exception` | **Not addressed**, and now the odd one out in its own file. See inline. |
| 6 | SPL deposit-revert skip hard-codes today's on-chain refusal with no lock-step marker | **Not addressed.** The comment at `:970-973` now explains *why* the refusal exists (the lamport-denominated penalty stub), which is a real improvement, but still does not say that a program-side policy change requires a matching relay change. When the SPL revert penalty lands, a not-yet-upgraded relay will `return metas` here while the upgraded handler requires the vault and ATA — `EffectAccountMissing`, aborting every batch that packs an SPL `DEPOSIT_REVERT`. One sentence naming the coupling is enough. |
| 7 | `OutpostConfig` re-fetched and re-decoded per token code | **Partially.** See inline. |
| 8 | Duplicate `reserve_aggregate` PDA derivation | **Fixed** — `_program_client->reserve_pda` is used and the local helper is gone. |
| 9 | Hidden cross-extractor coupling on the recipient sweep | **Fixed.** The manifest loop now declares the destination wallet itself (`:964`), and #552 deleted the sweep it used to depend on, so the coupling no longer exists. |
| 10 | Magic-literal seed strings | **Not addressed**, and now inconsistent with its own stack. See inline. |

---

## New findings

**The manifest derivation — the entire subject of this PR — is untested.**

The new tests cover the extractor half well: `extract_effects_slash_carries_collateral_position_key`, `extract_effects_keeps_distinct_collateral_token_codes`, and `extract_effects_skips_non_settling_operator_actions` pin which effects come out and which `token_code` rides each one. Nothing exercises the other half — `derive_collateral_position_pda`, `derive_collateral_vault_pda`, or the settlement-owner selection at `:985` (`reserve_pda` for SLASH vs the recipient for WITHDRAW_REMIT).

That is the half where a defect is silent locally and loud only in production: a one-character seed typo or a swapped ATA owner derives a well-formed PDA that simply is not the one the program requires, which surfaces as `EffectAccountMissing` at runtime, and on this stack that aborts the batch and holds the dispatch cursor. `accounts_for_effect` is a lambda inside `drain_dispatch` today, so there is no seam to test through — the same shape of problem #552 solved for the drive loop by factoring it over its RPC touchpoints. Golden-vector assertions on the three derivations against known program-side values would also work and are cheaper.

**The PR description no longer describes the code.**

Three symbols it documents are absent at head: `extract_inbound_collateral_settlement_targets`, `collateral_settlement_destination`, and the `extract_inbound_recipient_pubkeys` sweep it defers the native depositor to (deleted by #552). The description also still carries the exact inverted rationale that was correctly fixed in the source per finding 3 — *"an unreadable config degrades to the handler's log-and-skip rather than aborting the whole envelope delivery"*, and *"SLASH still flips the operator's status and then silently skips the seizure"*. A missing vault or destination ATA reaches `require_remaining_account` and returns `Err(EffectAccountMissing)`; it does not skip. Worth refreshing, since the description outlives the branch and this is the failure class the epoch-stall runbooks cover.

Also: *"Rebased onto current `master` (0 behind)"* is no longer true given the stacking above, and *"E2E flow run for the set pending"* — worth landing that run before merge, paired the way #552's was (`BRANCH_WIRE_SYSIO` + `BRANCH_WIRE_SOLANA` both on their stacked branches), since the account inventory is exactly what a flow exercises and unit tests do not.

---

**Verdict:** the account inventory itself still checks out — I re-traced the branches against wire-solana#409 and the seeds, endianness, writable flags and ATA inputs match the handlers. Findings 1 and 4 are the ones I would want resolved before this merges: both produce silent or permanently-wedged outcomes rather than loud ones, and neither is reachable by retry. 5, 6, 7 and 10 are small. The stacking/base issue is worth fixing first simply so the rest of the review has a diff that matches the change.

valthon:
## 14/14 e2e flows ✅

Cross-repo E2E integration run with all three branches pinned to `fix/SOL-375-379-380-opp-collateral` (wire-sysio, wire-solana, wire-tools-ts; other repos on manifest defaults, Release build): **[14/14 flows passed](https://github.com/Wire-Network/wire-platform-build-system/actions/runs/31853124095)** — full platform build + every discovered flow green, no epoch stalls, OPP circulation clean.

Directly exercising this branch's surfaces:
- **operator-collateral-deposit** ✅ / **reserve-lifecycle** ✅ — F1 per-position custody read + F3 SPL reserve-ATA provisioning
- **underwriter-slashing** ✅ / **batch-operator-slashing** ✅ — SPL slash prechecks-before-debit
- full **swap** suite ✅ (incl. swap-private-reserves, swap-non-native-tokens) — SPL terminal-settlement paths (SOL-380)

Run: https://github.com/Wire-Network/wire-platform-build-system/actions/runs/31853124095

heifner (commented):
Re-review at `af32d3d098`, traced against wire-solana#409 at `70f229ed` and wire-tools-ts#57 at its current head.

**The stacking is resolved** — #552 is on `master`, this PR is one commit on `101847d4c6`, and the rendered diff (652 / 84 across 3 files) is now exactly the change. The description has been rewritten and matches the code; the inverted "log-and-skip" rationale is gone from it.

**The custody rework is the right fix, done well.** Reading `custody_mint` off the `CollateralPosition` PDA rather than the `OutpostConfig` maps closes finding 1 at the source, the cache is correctly keyed on `(operator, token_code)` rather than the token code alone, and `collateral_position_custody` mirrors `reserve_info_for_codes` exactly — read outside the decode `try`, absent/empty degrades, present-but-unreadable throws loudly with the cursor untouched. `build_manifests_follows_collateral_custody_per_position` pins the property that matters (two operators, one token code, disagreeing custody → different manifests). Extracting `build_dispatch_manifests` over `read_collateral_custody` closes most of the "manifest derivation is untested" gap from last time.

I re-traced the account inventory branch by branch against `70f229ed`. Five of the six settlement branches match the handlers exactly:

| Branch | Relay declares | Program requires | |
|---|---|---|---|
| WITHDRAW_REMIT native | operator (w), position (w) | both, via `assert_writable_remaining_account` | ✅ |
| WITHDRAW_REMIT SPL | operator (w), position (w), vault (w), ATA(operator, mint) (w), token program | same + operator as lenient rent recipient | ✅ |
| SLASH native | position (w) | position; `vault` / `reserve_aggregate` are named accounts | ✅ |
| SLASH SPL | position (w), vault (w), ATA(reserve_aggregate, mint) (w), token program | same | ✅ |
| DEPOSIT_REVERT native | depositor (w), position (w) | both | ✅ |
| DEPOSIT_REVERT SPL | depositor (w), position (w) | **+ vault, + ATA(depositor, mint), + token program** | ❌ |

That last row is the one blocking finding.

---

## 1. The SPL DEPOSIT_REVERT skip is no longer true against the companion, and the outcome is a wedged epoch

`outpost_solana_client.cpp:868-871`:

```cpp
// The SPL deposit-revert branch is refused on-chain by design (the
// lamport-denominated penalty stub has no meaning in token units),
// so it never needs vault / ATA / token-program extras.
if (effect.shape == effect_shape::deposit_revert) continue;
```

That premise held at wire-solana `51ac5e8d` — the head I traced on 08-12 — where the branch read *"SPL DEPOSIT_REVERT is DELIBERATELY NOT IMPLEMENTED … the handler refuses rather than settling on an assumption"* and required no accounts.

It does not hold at `70f229ed`. `handle_deposit_revert`'s non-native branch (`inbound.rs:1561-1579`) now settles for real:

```rust
let destination_ata =
    associated_token::get_associated_token_address(&depositor, &position.custody_mint);
let Some((vault_ai, dest_ai, token_prog_ai, vault_bump_spl)) =
    resolve_collateral_vault_transfer(remaining, token_code, position.custody_mint,
                                      destination_ata, "DEPOSIT_REVERT")?
```

`resolve_collateral_vault_transfer` `require_remaining_account`s three accounts the relay never declares for this shape — the `collateral_vault` PDA (`:329`), the depositor's destination ATA (`:362`), and the SPL token program (`:403`) — and each returns `Err(EffectAccountMissing)` on absence (`mod.rs:326-342`, *"aborting dispatch"*). The `?` propagates through `dispatch_attestation(...)?` in the settle loop, so the whole `dispatch_attestations` transaction aborts, `dispatched_count` is never written, and `drive_dispatch_rounds` repacks the identical window from the identical cursor on every subsequent tick. That is a permanent stall, not a retry — precisely the failure class this PR's own **IMPORTANT** block describes.

It is reachable on ordinary paths, not exceptional ones. `deposit_non_native` opens SPL-custody positions, and the depot emits `DEPOSIT_REVERT` as its routine deposit-rejection mechanism (`sysio.opreg.cpp:1125` / `:1133` / `:1141` / `:1153` — unknown operator, type not permitted, etc.), parameterised by `chain_code` / `token_code` with nothing restricting it to native tokens. Any rejected SPL collateral deposit on SOL trips it.

The 14/14 e2e run does not cover this: `grep -rl "DEPOSIT_REVERT\|DepositRevert" packages/flow-*` returns nothing, so no flow exercises the attestation type at all, in either custody mode.

The fix is small, because the existing code already computes the right owner. Deleting the early `continue` lets `deposit_revert` fall through to the same custody read and SPL block as `withdraw_remit`; `settlement_owner` (`:881`) is `*effect.recipient` for every non-`slash` shape, and the depositor wallet is already declared at `:864`. That yields exactly the five accounts the handler requires, and the rent-refund recipient the full-drain close looks for. The comment goes with it.

This is finding 6 from both prior reviews, no longer hypothetical. Worth adding the one sentence it asked for in the other direction too: this manifest and `handle_deposit_revert` must move together, because the relay's shape decisions are unversioned against the program.

## 2. Neither `slash` nor `deposit_revert` is exercised at the manifest level

`effect_shape::slash` and `effect_shape::deposit_revert` appear only in extractor tests (`:1032`, `:1076`, `:1132`). Every `build_dispatch_manifests` case goes through `withdraw_remit_effect`. The harness even carries a `reserve_aggregate` member (`:1361`) that no assertion ever reads.

Those two shapes are the ones with branch-specific manifest logic — the SLASH owner swap (`reserve_aggregate` instead of the recipient) and its deliberate omission of the operator wallet, and the `deposit_revert` early return. Finding 1 is exactly the defect a `deposit_revert` manifest test would have caught, and a SLASH SPL test would pin the one place a wrong ATA owner produces a silently-stuck seizure rather than a loud failure. Two cases in the shape of the one you already wrote.

## 3. Seed strings are still inline char arrays — third review

`:615` `{'c','o','l','l','a','t','e','r','a','l','_','p','o','s','i','t','i','o','n'}` and `:627` `{'c','o','l','l','a','t','e','r','a','l','_','v','a','u','l','t'}`, against `opp_states.rs:56` `COLLATERAL_POSITION_SEED` and `:103` `COLLATERAL_VAULT_SEED`.

Same file, `:41-42`:

```cpp
constexpr std::string_view EPOCH_DELIVERIES_SEED = "epoch_deliveries";
constexpr std::string_view ENVELOPE_CHUNKS_SEED  = "envelope_chunks";
```

— with a comment stating they are byte-exact mirrors that must agree or every seeds-validated call fails. The new seeds are the same kind of constant with the same consequence, spelled the unsearchable way, 570 lines below the pattern. There are now five inline seed literals in this plugin (`reserve`, `reserve_vault`, the two new ones, and `reserve_aggregate` in `outpost_solana_client_plugin.hpp:205`). CLAUDE.md's "no magic literals" aside, a one-character typo here derives a well-formed wrong PDA that only fails at runtime as `EffectAccountMissing`.

## 4. `token_custody_info::decimals` is always zero, and nothing needs it

`collateral_position_custody` hardcodes `0` (`:1404`) — correctly, since `CollateralPosition` carries no decimals — and the header documents the field as *"Chain-native decimals when a caller needs them"*. No caller needs them. A future one that reads `.decimals` gets `0` silently, which is wrong for every SPL token, and the sibling `reserve_terminal_info` in the same header carries a real `custody_decimals` alongside a `custody_mint`, so the two structs are one rename away from being confused. Either drop the field (leaving a struct that is just the mint) or make it `std::optional<uint8_t>` so an unset value cannot be read as a valid one.

## 5. `assert_collateral_position_shape` asserts four fields; the relay reads one

`assert_reserve_shape` asserts exactly the three fields the manifest resolves. This one requires `operator`, `token_code` and `amount` in addition to `custody_mint`, none of which the relay decodes, so a benign program-side rename of `amount` becomes a hard boot failure for the batch-operator role. If the extra three are a deliberate drift canary — which the header's *"the four fields the on-chain settlement path binds together"* implies — say so in one line; otherwise narrow it to what is read.

## 6. `reserve_pda`'s doc names the wrong seed

`outpost_solana_client_plugin.hpp:119` says *"Pre-derived from seed `outpost_reserve`"*; the derivation twenty lines down (`:205`) uses `reserve_aggregate`, and a comment at `:199` explains why. Pre-existing, but this PR is the first consumer where `reserve_pda` selects an ATA owner, and a wrong owner there yields `Ok(None)` → *"status flipped, funds stuck"* rather than a loud abort. Worth correcting the comment while the path is being added.

Also cosmetic: no blank line between `derive_collateral_vault_pda` and `count_inbound_attestations` (`:629`).

---

## Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | Custody read from the mutable `OutpostConfig` rather than the position's pinned `custody_mint` | **Fixed**, with the regression test |
| 2 | Unachievable "degrade to log-and-skip" fallback | Resolved by #552 |
| 3 | Inverted log-and-skip rationale in source and description | **Fixed** in both |
| 4 | Nothing creates the `reserve_aggregate` ATA | **Addressed** — `SolanaOutpostBootstrapper.ts:399-409` pre-creates it per registered SPL mint, and the program now documents it as an ops precondition. Production provisioning is a runbook item, not a relay concern. The failure mode is still a silent `Ok(None)` "funds stuck"; an operator alarm on it is worth a follow-up, not this PR |
| 5 | "Never throws" contract catching only `fc::exception` | **Moot** — that function is gone. `collateral_position_custody` throws by design and mirrors `reserve_info_for_codes`'s single-arm log-and-rethrow exactly; a non-`fc` throw loses only the diagnostic `elog` |
| 6 | SPL deposit-revert skip hard-codes today's on-chain refusal | **Materialised — see finding 1** |
| 7 | `OutpostConfig` re-fetched per token code | **Gone** — collateral no longer reads the config at all |
| 8 | Duplicate `reserve_aggregate` derivation | Fixed |
| 9 | Cross-extractor coupling on the recipient sweep | Fixed |
| 10 | Magic-literal seed strings | **Still open — see finding 3** |
| — | Base says `master` while stacked on #552 | **Resolved** |
| — | Manifest derivation untested | **Mostly closed** — see finding 2 for what remains |

## Two things I checked and am not raising as findings

**Transaction sizing.** Collateral settlements grow a manifest from one account to as many as five, but `drive_dispatch_rounds` packs greedily against the account union with `MAX_TERMINAL_DYNAMIC_ACCOUNTS`, always taking at least one attestation, so denser manifests shrink the window rather than overflowing the packet. Round-budget exhaustion resumes from the on-chain cursor on the next tick.

**Read volume.** One `getAccountInfo` per distinct `(operator, token_code)` is inherent to per-position custody and is the correct trade for what finding 1 bought. A 9-operator slash envelope pays nine sequential reads on the deadline-bounded path; the per-effect deadline probe plus #552's resumable drain make that a delay rather than a failure.

There is also a narrow TOCTOU: a position that is absent when the relay reads it and opened with SPL custody before the terminal transaction executes produces an abort. It self-heals, because the next tick re-reads and declares the extras — worth knowing when reading a one-off `EffectAccountMissing` in a log, not worth defending against.

---

**Verdict:** the custody rework, the test extraction and the description rewrite all landed well, and the inventory is right on five of six branches. Finding 1 blocks: the companion implemented SPL `DEPOSIT_REVERT` in the same window this branch was rebased, the relay still declares the manifest for a handler that refuses, and the outcome is a held cursor with no retry path. It is a small change here plus the manifest test in finding 2, and it wants a paired e2e — with no flow covering `DEPOSIT_REVERT` in either custody mode, this specific regression is invisible to the gate that just went 14/14.

valthon:
Re-review findings addressed at `61d273b1c9` (amended into the single PR commit; base unchanged, `master` has not moved).

**1 — SPL `DEPOSIT_REVERT` skip (blocking):** the early `continue` and its stale "refused on-chain by design" comment are gone. `deposit_revert` now falls through the same per-position custody read and SPL block as `withdraw_remit`: native custody stays depositor + position; SPL custody declares depositor wallet (the full-drain close's rent recipient) + position + `collateral_vault` + the **depositor's** canonical ATA + token program — exactly the five accounts `handle_deposit_revert`'s `resolve_collateral_vault_transfer` requires at `70f229ed`. The lock-step sentence you asked for (both directions) now sits on the SPL settlement block: which shapes settle SPL here must match the program's `resolve_collateral_vault_transfer` callers exactly, and a program-side settlement-policy change and this manifest must move together.

**2 — SLASH / DEPOSIT_REVERT untested at the manifest level:** two new `build_dispatch_manifests` cases in the shape of the existing custody-per-position test:
- `build_manifests_slash_settles_into_reserve_aggregate_ata` — pins the owner swap (ATA owner is `reserve_aggregate`, **not** the operator, asserted both ways) and the deliberate absence of the operator wallet, native and SPL.
- `build_manifests_deposit_revert_declares_spl_refund_accounts` — pins the five-account SPL refund manifest and the two-account native one. This case fails on the pre-fix early-return, so the finding-1 regression is now discriminated.

**3 — seed strings:** `COLLATERAL_POSITION_SEED` / `COLLATERAL_VAULT_SEED` are now named `constexpr std::string_view` constants in the same block as `EPOCH_DELIVERIES_SEED` / `ENVELOPE_CHUNKS_SEED`, with the same byte-exact-mirror doc; the derivations consume them the same way. (Blank line before `count_inbound_attestations` restored too.)

**4 — `token_custody_info::decimals`:** dropped; the struct is mint-only, and its doc now says why (the position carries no decimals; a decimals field here could only hold a made-up value) and contrasts `reserve_terminal_info.custody_decimals` as the real pinned fact.

**5 — `assert_collateral_position_shape`:** the four-field check is a deliberate drift canary — now stated explicitly in the header contract ("the relay itself decodes only `custody_mint`; the other three fail loudly at boot rather than surfacing later as a live position this relay silently half-understands").

**6 — `reserve_pda` doc:** corrected to seed `reserve_aggregate`, pointing at the derivation's `outpost_reserve` mis-seed incident note, and now mentions its role as the SPL SLASH seizure ATA owner.

Validation: `test_outpost_solana_client_plugin` rebuilt and green (`*** No errors detected`, including the two new cases); `nodeop` links clean. A paired e2e (`BRANCH_WIRE_SYSIO` + `BRANCH_WIRE_SOLANA` + `BRANCH_WIRE_TOOLS_TS` on their stacked branches) is being dispatched against this head — noting your caveat that no flow exercises `DEPOSIT_REVERT` in either custody mode, so the finding-1 regression itself is pinned by the new unit case rather than the gate.

(The prior head's Apple Silicon failure was `checktime_intrinsic`'s `is_code_cached` assertion — unrelated JIT-timing test; the identical content passed at `af32d3d098` and the identical base passes on `master`. This push re-runs it.)

heifner (dismissed):
Approving at `61d273b1c9`.

Finding 1 is resolved at the source rather than patched around: all three collateral shapes now run one uniform path, so no per-shape early return is left that can go stale against the program again. I re-traced every branch against wire-solana#409 at `70f229ed` — all six match the handlers now, including the five-account SPL `DEPOSIT_REVERT` manifest (`resolve_collateral_vault_transfer` at `inbound.rs:1565`, plus the depositor wallet the full-drain close resolves leniently).

The two new manifest cases pin the right properties: the SLASH owner swap is asserted both ways, and the deposit-revert case requires five accounts where the pre-fix early return produced two — so it discriminates the regression rather than just describing it. Findings 3–6 all addressed.

One non-blocking leftover: `outpost_solana_client.hpp:589-592`, the `effect_shape::deposit_revert` enum doc, still says the SPL branch is *"REFUSED on-chain by design … so this shape never carries SPL extras."* Both clauses are now false and the shape carries exactly those extras. It is the last copy of the premise that produced the finding, sitting where a reader looks up what the shape means — worth deleting on the way in.

Land it once CI is green here and the paired e2e comes back. Standing caveat on that gate: no flow exercises `DEPOSIT_REVERT` in either custody mode, so the new unit case is what holds that line.

heifner (approved):

PR #543, opened 2026-07-30
WIRE-291: prevalidate underwriter UIC signatures

huangminghuang:
## Summary

- Prevalidate each relayed underwrite-intent commit (UIC) before it can create or alter candidate evidence, locks, reserves, or settlement state.
- Require canonical UIC encoding and fixed packed K1, R1, EM, or ED signatures that satisfy the claimed underwriter's current `active` or `owner` permission threshold. Ethereum and Solana bind that claimed identity to their authenticated caller before relaying the unchanged bytes.
- Reject swaps whose source and destination are the identical `(chain, token, reserve)` triple before custody or request mutation.
- Preserve a complete candidate while mutable prerequisites (operator activation, collateral, destination authex link, reserve activation, or liquidity) are unavailable. An outpost may trigger re-evaluation only by relaying the exact stored UIC in a new transaction; stored evidence and timestamps remain immutable.
- Remove the public `retrycommit` action and autonomous underwriter retry scheduling/submission. Bounded, cancellable autonomous recovery is tracked in [WIRE-330](https://wire-network.atlassian.net/browse/WIRE-330).
- Regenerate the `sysio.uwrit` ABI/WASM artifacts.
- Remove the obsolete CTest registration for a deleted test case, restoring the Linux and macOS CI matrix.

## Deployment scope

Pre-launch behavior for disposable redeployment. This changes system-contract and node/plugin behavior and requires the regenerated `sysio.uwrit` artifact; no data migration or mixed-version compatibility path is provided.

## Validation

- Local Release Phase A and Phase B previously passed for the coordinated companion heads; Phase B passed all 13 flows at concurrency 16, including `swap-with-underwriting`.
- After merging `master`, rebuilt the resolved `sysio.uwrit` artifact with CDT and verified its generated ABI matched the checked-in ABI exactly.
- Merged-head focused contract validation passed: `amm_math_tests`, `sysio_dispatch_tests`, `sysio_emissions_tests`, `t5_emissions_tests`, and `sysio_reserve_tests` (225 cases, 13,530 assertions).
- The CTest-registration correction was validated by regenerating the Release build, confirming the stale target is absent from the generated test manifest, and running `test_chain_plugin_chain_plugin_default_tests` successfully.
- Current-head Linux, macOS, and OPP CI is running and monitored.

## Companion heads

- `wire-sysio`: `febcedb8f78d56bc2ac8fad4f14c5a2fb662088c`
[...]

PR #556, opened 2026-08-10
WIRE-316: Bind source-deposit hash to target_amount

heifner:
## Jira ownership

Follow-up to [WIRE-316](https://wire-network.atlassian.net/browse/WIRE-316). PR #550 intentionally split the caller's immutable `target_amount` from the depot AMM quote; this PR fixes the resulting verifier regression. It is independent of logging-only WIRE-295 and terminal-payout semantics WIRE-332.

Split out of #543, which fixes this as part of a broader UIC-prevalidation refactor. It is isolated here because it is a live defect on `master` with a small, independently-reviewable fix, and #543 still has open design discussion on unrelated `sysio.uwrit` reject semantics.

## The defect

Both outposts fold the user's accepted `target_amount` into the `SwapDeposit` correlation hash:

- EVM — `ReserveManagerLib.hashSwapDeposit(sender, sourceAmountDepot, args)` packs `args.targetAmount`
- SVM — `swap_correlation_hash(..., target_amount, target_tolerance_bps)`

The underwriter's source-deposit verifier recomputed that preimage from `uwreq.dst_amount`.

Those were the same value until #550 split them. `dst_amount` is now the depot's AMM quote, re-priced at ingestion; `sysio.uwrit.hpp` states it is "**never** the caller's `SwapRequest.target_amount`". The recomputed hash therefore matches only when the caller's target happens to equal the depot's quote **exactly** — and a gap between the two is the *expected* case, which is what `variance_tolerance_bps` exists to bound. Any swap with a non-zero gap fails verification, the UWREQ never leaves `PENDING`, and no underwriter commits.

## Observed

A live ETH→SOL underwritten swap (`target_amount` 98039214, `dst_amount` 97747972) produced **148 consecutive** rejections against a single stable hash pair, until the flow timed out waiting for a CONFIRMED commitment:

```
underwriter: source-deposit verify failed for uwreq 11 — SwapDeposit hash mismatch (id=1):
  on-chain=fd8f16aad2443acf2847c6f49c41feac6f32feadc87682df25835ae328a76695
  recomputed=758ed0f49ec37030baebfdc586a4d57a36f88492c24fcb2758d8f5b4d8ffa2ff
```

Recomputing that preimage both ways confirms the split precisely:

| preimage destination amount | keccak256 | |
|---|---|---|
| `target_amount` = 98039214 | `fd8f16aa…` | the hash `ReserveManager` emitted |
| `dst_amount` = 97747972 | `758ed0f4…` | what the verifier computed |

## The fix

Carry `target_amount` on the plugin's `uw_request`, read it from the `uwreqs` row, and use it in the preimage.

The packing was **duplicated** across the EVM and SVM verifiers, which is how one wrong field became two. Both now call a single `source_deposit_hash` helper (lifted verbatim from #543, so that PR's rebase is a no-op) whose input struct names the field `target_amount` — the depot's mutable quote is no longer reachable from the preimage by accident. The EVM leg also gains the explicit depositor-size check the SVM leg already had; previously it was only implied by a post-hoc buffer-length assertion the helper makes unnecessary.

## Coverage

Two cases pin the preimage to the production vector above — the recomputed hash equals the `SwapDeposit` hash `ReserveManager` actually emitted, and the settlement quote demonstrably does not reproduce it.

Note the tests pin the **helper's** preimage layout and field semantics, not the two call sites' argument selection; the naming (`target_amount` vs `dst_amount` on the input struct) is what guards those.

## Verification

- `test_underwriter_plugin` — 50 cases, `*** No errors detected`
- `plugin_test` — `*** No errors detected`
- `nodeop` links clean

No `contracts/**` changes, so `contracts_unit_test` is not implicated.

[WIRE-316]: https://wire-network.atlassian.net/browse/WIRE-316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugins/underwriter_plugin/include/sysio/underwriter_plugin/source_deposit_constants.hpp
+++ b/plugins/underwriter_plugin/include/sysio/underwriter_plugin/source_deposit_constants.hpp
@@ -12,6 +12,13 @@
 
 namespace sysio::underwriter {
 
+/// Byte length of an EVM address carried in `SwapRequest.actor.address`.
+inline constexpr size_t EVM_DEPOSITOR_SIZE = 20;
+
+/// Byte length of a Solana Ed25519 public key carried in
+/// `SwapRequest.actor.address`.
+inline constexpr size_t SVM_DEPOSITOR_SIZE = 32;
+
 /// ETH: maximum number of recent blocks one source-deposit `eth_getLogs`
 /// lookup may cover.
 ///
--- a/plugins/underwriter_plugin/include/sysio/underwriter_plugin/source_deposit_hash_detail.hpp
+++ b/plugins/underwriter_plugin/include/sysio/underwriter_plugin/source_deposit_hash_detail.hpp
@@ -1,0 +1,62 @@
+#pragma once
+
+#include <climits>
+#include <cstddef>
+#include <cstdint>
+#include <span>
+#include <type_traits>
+#include <vector>
+
+#include <fc/crypto/keccak256.hpp>
+
+namespace sysio::underwriter {
+
+/// Number of unsigned 64-bit fields in the outpost correlation-hash payload.
+inline constexpr size_t SOURCE_DEPOSIT_U64_FIELD_COUNT = 7;
+
+/// Immutable source-outpost fields covered by the `SwapDeposit` correlation
+/// hash. The outposts hash the user's accepted `target_amount`, not the depot's
+/// mutable AMM quote (`uwreq.dst_amount`). Keeping the two names distinct here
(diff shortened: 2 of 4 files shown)


huangminghuang (approved):
Approved. As the PR description notes, these tests pin the helper's preimage rather than the two production call sites. The call-site regression test—using distinct `target_amount` and `dst_amount` values and asserting that the EVM/SVM verifier input selects `target_amount`—should be added in #543, which owns the broader verifier refactor.

PR #558, opened 2026-08-11
Contracts: credit claimable balances instead of pushing transfers on never-throw paths

heifner:
Closes SEC-150.

## The vulnerability

`sysio.token::transfer` calls `require_recipient` on both sides, and `apply_context::exec` dispatches notified receivers in a bare loop with no exception isolation. An assert inside a recipient's `on_notify("sysio.token::transfer")` handler therefore aborts the entire transaction, including every parent inline action. Any account can `setcode` such a handler — `contracts/test_contracts/reject_all.wasm` has done exactly this in-tree for years. A handler that burns CPU instead of asserting is the same vector, so the handler must not run at all rather than be tolerated.

That made every pushed payout on a never-throw path abortable by its own recipient. The contracts pre-validate every reachable `check()` so `sysio.epoch::advance` and the OPP inbound dispatch chain cannot abort, but that discipline stops at the contract's own guards: once value is pushed to an account the protocol does not control, the counterparty decides whether the transaction commits.

Four sites were reachable this way. Three sit in the `sysio.epoch::advance` inline subtree, where an abort halts epoch advancement, emissions accrual and outbound envelope construction chain-wide:

| Site | Recipient | Effect |
|---|---|---|
| `uwrit::drainfwq` -> `reserv::refundwire` | swap-from-WIRE user | permanent chain-wide epoch halt |
| `sysio.system::payepoch` | producers, standbys, batch operators | every pay-epoch aborts |
| `sysio.opreg` WIRE-chain remits | operator being terminated | self-defending deadlock |
| `sysio.reserv::paywire` | swap-to-WIRE recipient | OPP consensus dispatch stalls |

The refund case is the worst of them and the cheapest to mount. The drain's queue-row erase rolls back with the abort, so the offending row survives to re-block every later epoch — an indefinite halt bought for one `min_fromwire_amount` escrow, paid once. The termination case never converges either: the operator blocks its own remit, so the termination never commits and each later advance retries it.

## The fix

All four sites now credit a claimable balance and transfer nothing. The recipient pulls it with an action carrying its own authority, so a hostile handler blocks only its own payout. This generalizes the pattern `sysio.dclaim` already used (credit `pending_claims` on the never-throw path, pay out from a separate user-initiated `claim`) into a shared `sysio.opp.common/claimable.hpp`.

The one deliberate exception is a payout to an account the protocol itself controls: `payepoch`'s T5 category buckets and `fundclaim`'s transfer to `sysio.dclaim` are still pushed. See [UX changes](#ux-changes).

## New actions

The three claims are permissionless and self-authorized: the caller may only claim its own balance.

| Action | Auth | Pays | Credited by | Symbol |
|---|---|---|---|---|
| `sysio.system::claimpay(account_name)` | the claiming account | epoch pay: producer, standby and batch-operator shares | `payepoch` | WIRE |
| `sysio.reserv::claimwire(account)` | the claiming account | swap-to-WIRE payouts and swap-from-WIRE refunds | `paywire`, `refundwire` | WIRE |
| `sysio.opreg::claimremit(account)` | the claiming account | returned CORE_SYM collateral | withdraw flush, deferred lock release, termination payout | SYS |

Each drains the caller's row in full — there is no partial claim — and throws `no epoch pay to claim` / `no claimable WIRE for this account` / `no claimable remit for this account` when nothing is owed. `claimwire` additionally refuses a row past its retention deadline.

A fourth action is maintenance rather than a payout, and is **not** permissionless:

| Action | Auth | Does | Called by |
|---|---|---|---|
| `sysio.reserv::sweepclaims(max_rows)` | `sysio.epoch` or self | erases up to `max_rows` expired `wireclaims` rows and returns their WIRE to the treasury | inlined by `sysio.epoch::advance` every epoch |

## New tables

| Table | Contract | Key | Holds | Expires |
|---|---|---|---|---|
| `payclaims` | `sysio.system` | account | epoch pay owed | stamped, not enforced (WIRE-339) |
| `payclaimtot` | `sysio.system` | singleton | total outstanding epoch pay | n/a |
| `wireclaims` | `sysio.reserv` | account | swap payouts and refunds owed | yes — one year, swept |
| `remitclaims` | `sysio.opreg` | account | collateral owed | stamped, not enforced (WIRE-339) |

`payclaims` / `wireclaims` / `remitclaims` are readable to show a "you have X to claim" balance before the user sends anything.

## UX changes

Anything that previously waited for WIRE to arrive in an account after one of these events must now claim it. Balances no longer move on their own.

- **Bridging into WIRE (swap-to-WIRE) is now two steps.** The swap settles as before and the recipient is credited immediately, but the WIRE stays in `sysio.reserv` custody until the recipient calls `claimwire`. Wallets and bridge UIs need a "claim" step, or a relayer that claims on the user's behalf.
- **Swap-from-WIRE refunds are claimed, not returned.** A reverted or expired swap credits the escrow back rather than transferring it. The user's balance does not change until `claimwire`.
- **Producers, standbys and batch operators claim their epoch pay.** `payepoch` no longer moves WIRE into their accounts. Any monitoring that reads producer balances as a proxy for "was I paid" must read `payclaims` instead, or claim first.
- **The T5 category buckets are still PAID DIRECTLY — nothing changes for `sysio.ops` and `sysio.gov`.** They are pushed rather than credited because no claim path can ever reach them: a claim needs `require_auth(account_name)`, `sysio.roa::is_sysio_account` forces `net_weight`/`cpu_weight` to zero for every `sysio`-prefixed account so neither can pay for a transaction, and unlike `sysio.dclaim` neither carries a contract that could emit the claim inline. Crediting them would convert an earned allocation into a permanently growing `payclaimtot` reservation nobody can move. They are protocol-owned holding accounts with no code, so the notify-handler threat the pull model defends against does not exist for them — the same exception `fundclaim` already makes for `sysio.dclaim`. The constraint this puts on the accounts is written at the push site: any contract governance ever deploys on `sysio.ops` or `sysio.gov` MUST NOT assert or burn CPU in an `on_notify("sysio.token::transfer")` handler.
- **Terminated operators claim their collateral.** Termination debits the operator row as before, but the SYS lands in `remitclaims`. This is independent of `prune` — a pruned operator row does not strand the collateral.
- **Unclaimed `wireclaims` rows expire after one year** and are swept back to the treasury, on an epoch-driven trigger rather than only on settlement traffic. `payclaims` and `remitclaims` record a stamp but nothing expires them today.

Nothing changes for outpost-chain remits, which still flow through `WITHDRAW_REMIT` attestations, or for `sysio.dclaim` staking rewards, which were already claim-based.

## Implementation details worth review

- Credits saturate at the `sysio::asset` magnitude limit rather than the integer limit, so a credited balance is always payable. Saturating at `UINT64_MAX` would merely move the abort from credit time to claim time and strand the balance permanently.
- Claim actions erase the row **before** queueing the transfer, closing the re-entrancy window a notify handler would otherwise use to drain a row twice — the same ordering guard `opreg::deposit` already applies.

## Treasury accounting

Unclaimed pay still sits in the treasury's token balance while being fully owed, so every gate that spends against that balance reserves it: `sysio.system::fundclaim`'s balance cap, `sysio.epoch`'s emissions readiness gate, and `sysio.system::claimnodedis`, which now refuses a node-owner withdrawal that would eat into `payclaimtot.outstanding`. Without this the treasury double-commits and a later claim fails on overdraw, stranding earned pay. The outstanding total is a maintained counter (`payclaimtot`) rather than a table scan, because the epoch gate reads it on every advance.

`claimnodedis` rejects rather than caps: capping would either transfer less than `info.claimable` while `claimed` advanced by the full amount, silently destroying the difference, or thread a partial amount through the vesting accounting. It is user-initiated, so the throw reaches the caller who asked. Every other treasury outflow in `sysio.system` was audited in the same pass — `claimpay` decrements `outstanding` in lockstep, `fundclaim` already subtracted the reserve, and `payepoch`'s category pushes run inside the gate that reserves it.

`sysio.epoch`'s blocklog now reports `sysio_balance` as the *spendable* amount for the same reason — an operator reading `BALANCE_INSUFFICIENT` needs the shortfall against what can really be paid.

## Expiry

`sysio.reserv`'s claimable set is unbounded and caller-influenced, and the rows bill to the sysio RAM pool, so they carry a one-year window. The producer and operator sets are bounded per-epoch and hold earned pay and returned collateral, so they are stamped but not swept. The window is a documented constant rather than config, to keep the swap-settlement surface unchanged; it is a RAM backstop, not an economic lever.

Four pieces enforce the `wireclaims` window, and the sweep itself is one implementation shared by both triggers so they cannot drift on what "expired" means or where the reclaimed WIRE goes:

- **`sweepclaims`, inlined by `sysio.epoch::advance`** with `MAX_CLAIM_SWEEP_PER_EPOCH`. This is what makes the deadline hold when settlement stops. It is placed **before** the emissions gate, deliberately: what it reclaims lands in the very balance the gate measures, so ordering it after the gate lets a `BALANCE_INSUFFICIENT` epoch return without reclaiming the forfeited WIRE that would cover its own shortfall — on every retry, with `sweepclaims` taking only epoch/reserv authority so no ordinary keeper could break the cycle. It is budget-bounded, never throws past its auth check, touches no epoch state, and is guarded on `is_account(sysio.reserv)` (an inline to an absent code account is a hard `action_validate_exception`, which inside `advance` is a chain-wide stall).
- **`credit_wire_claim`'s opportunistic sweep** on the write path — bounded and best-effort; it only fires while credits keep arriving.
- **`claimwire` refuses an expired row.** The sweep is bounded, so a row can outlive its expiry for many epochs while it waits its turn; without this the window would mean "swept eventually" rather than "claimable until", and a balance the policy had already extinguished would still pay out in the meantime.
- **A credit onto an already-forfeited row settles it first** — the old balance goes to the treasury and the row is erased before the fresh claim is created. Otherwise a later `paywire`/`refundwire` would add to a refused balance and refresh its stamp, and a trickle of small credits could keep a system-funded row alive indefinitely.

`payclaims` and `remitclaims` carry `expires_at_sec` plus a `byexpiry` index maintained on every credit, and nothing reads them yet. The schema is free to add before launch and expensive after, and landing it now means a later retention pass inherits real age data instead of starting its clock the day the field is added — refreshing on every credit makes the stamp a "last had activity" signal, so an account still being paid never looks stale. Wiring a sweep there means forfeiting earned pay and returned collateral, which is a policy decision rather than a RAM one, and the in-tree precedents disagree (`sysio.dclaim::pending_claims` forfeits to the capital fund after 180d; `wireclaims` uses a year to the treasury). That decision is tracked in **WIRE-339**.

## Rollout order

These contracts are not independent deployables: `advance` inlines actions into six of them, and the emissions gate reads two more contracts' tables. **The atomic unit is the compatibility-coupled TRIO — `sysio.reserv`, `sysio.epoch`, `sysio.system` — proposed as ONE `sysio.msig` transaction.** The whole release does not fit and must not be deployed uniformly:

- **Size.** The five changed WASMs total 579,408 bytes, past the default `max_transaction_net_usage` of 524,288, and `msig::propose` takes the complete inner transaction before it chunks storage. The trio is 427,064 bytes with ABIs — it fits, but preflight the packed size of the actual proposal. `sysio.opreg` and `sysio.uwrit` are uncoupled and stage in their own transaction, in any order.
- **Deploy action, per account.** Separate system accounts (`sysio.epoch`, `sysio.reserv`, …) go through `sysio.roa::setsyscode` / `setsysabi`, which reconcile gifted RAM out of sysio's pool via `giftram` — a raw `setcode` there leaves the account short against its creation allowance, and this release grows `sysio.reserv` by 6,040 bytes and `sysio.opreg` by 2,447. The root `sysio` account (which holds `sysio.system` and IS the RAM pool) uses the **native** `setcode` / `setabi`, because `giftram` cannot self-target: it would decrement sysio's ROA ledger while `set_resource_limits` and `add_system_resources` cancel on the chain side, silently leaving the ledger below the real quota. So the trio's transaction mixes both action families.

If a staged rollout is forced instead, the order for this release is `sysio.reserv` → `sysio.epoch` → `sysio.system` (`sysio.opreg` and `sysio.uwrit` may land anywhere — they gain no inlined action and no other contract reads their new tables):

- **`sysio.reserv` before `sysio.epoch`** — the new `advance` inlines `sysio.reserv::sweepclaims`, guarded only on the account existing. An old `sysio.reserv` build has the account and not the action, and the CDT dispatcher asserts on an action it does not implement, so every advance would abort — a chain-wide epoch stall.
- **`sysio.epoch` before `sysio.system`** — the new `payepoch` retains WIRE in `payclaims`; the old gate counts that backing as spendable and could double-commit it, leaving credited claims underfunded.

Every intermediate state of that order is safe. The two version rules point in OPPOSITE directions and are stated separately — call edge `callee >= caller` (the contract receiving an inlined action upgrades first), table edge `reader >= writer` (the contract reading the new state upgrades first).

**Downgrades are a data-migration problem before they are an ordering one.** All three claim tables hold value only the new code can pay out (`payclaims`/`claimpay`, `wireclaims`/`claimwire`, `remitclaims`/`claimremit`), so rolling any of those contracts back with rows present strands the funds; and reverting `sysio.epoch` under a live new `sysio.system` lets `payepoch` keep creating claims the old gate does not reserve. The procedure is quiesce the credit writers, drain or migrate all three tables, then roll back `sysio.system` → `sysio.epoch` → `sysio.reserv` with `sysio.opreg` last — and a chain with uncooperative claimants is not safely downgradeable by code order alone.

The full edge list, the per-account deploy actions, the size table, the stale-read cases, and the downgrade procedure are written up in **`docs/contract-upgrade-order.md`**, new in this PR.

## Testing

A new `contracts/test_contracts/block_transfer` contract asserts on every incoming transfer, and three regression tests park it on each attack surface — a producer, a terminating operator, and a swap-from-WIRE refund recipient — asserting the system path completes, the blocker is still credited in full, and the only thing it can break is its own claim.

Retention and reserve coverage added on top of that:

- `expired_wire_claim_is_swept_without_further_traffic` — ages a claim past the window with no further credits, so the opportunistic sweep never fires: the claim is refused, a zero budget is a no-op, and the sweep reclaims the row to the treasury.
- `sweepclaims_leaves_live_rows_alone` — a live row survives the sweep and still claims normally.
- `recredit_after_expiry_does_not_revive_the_forfeited_balance` — a later credit starts a fresh claim; the forfeited balance went to the treasury rather than accumulating with a new window.
- `claimnodedis_reserves_outstanding_epoch_pay` — a treasury one unit short of pay + claim refuses the node-owner withdrawal, spends nothing, and lets it through once funded; the epoch pay is still really payable afterwards.
- `expired_wire_claims_unblock_a_balance_blocked_epoch` — a `BALANCE_INSUFFICIENT` epoch whose shortfall is exactly covered by an expired claim advances on the next attempt with **no** manual `sweepclaims` and **no** top-up. This pins the pre-gate placement of the inline and the queued-inline semantics: the first advance records the block and performs the reclaim, and the next retry sees the larger balance. Verified as a real regression — moving the call back below the gate fails the test.

Existing tests that asserted recipient balances after a payout now claim first and assert the same end state, so they cover the credit and the pull together. Treasury conservation becomes `balance decrease + outstanding claims == distributed`, which collapses to the original assertion once every recipient has claimed.

## Follow-ups required before this is usable end to end

This changes payout semantics, so two repos need companion changes and the swap-to-WIRE e2e flow fails until they land:

- **wire-tools-ts** — `flow-swap-to-wire` polls the recipient's WIRE balance until a deadline, which can no longer become true without a `claimwire`. It now waits on the claimable balance and claims. `WireClient` gains `claimWire` / `claimPay` and their claimable-balance readers.
- **wire-libraries-ts** — `sdk-core/src/types/SysioContractTypes.ts` regenerated for the new actions and tables, or SDK consumers cannot call them.

The other flows were checked and need no change: `flow-swap-from-wire` never exercises a revert, so `refundwire` is not on its path; `flow-reserve-lifecycle`'s WIRE assertions cover the `matchreserve` escrow, which still pushes from the user; and `flow-emissions-soak` covers `sysio.dclaim` staker claims, which were already pull-based.

The Ethereum and Solana outposts are unaffected: the change is depot-side, and outpost remits still flow through `WITHDRAW_REMIT` attestations.

`paywire` is the P1 site rather than one of the P0 halts, and it is the one carrying the user-facing swap change described above. It is the same vulnerability — a hostile swap recipient really can stall consensus dispatch — so it is fixed the same way here.

## Cross-repo companions and merge order

| Order | PR | Purpose |
|---|---|---|
| 1 | **this PR** — Wire-Network/wire-sysio#558 | contracts: credit claimable balances instead of pushing transfers |
| 2 | Wire-Network/wire-libraries-ts#66 | regenerate `SysioContractTypes` for the new actions and tables |
| 3 | Wire-Network/wire-tools-ts#62 | `flow-swap-to-wire` claims the credited payout instead of polling a token balance |
| 4 | Wire-Network/wire-hub-webapp#159 | Hub claimable-WIRE payout and refund-recovery UX |

Merge this PR first — it is self-contained and the others consume its ABI. #66 has been regenerated from this branch's current ABI, so it carries `sweepclaims` and the `expires_at_sec` / `byexpiry` additions alongside the three claim actions. #66 and #62 are independent of each other and can land in either order once this is in: #62 reads the new tables raw rather than through the typed accessor, so it does not wait on an `@wireio/sdk-core` release of #66. Between this merging and #62 landing, `flow-swap-to-wire` fails against master. #159 is stacked on wire-hub-webapp#144 and unblocks once this and #66 are in.


huangminghuang (dismissed):
Approved. The remaining claim-retention risk is explicitly accepted and tracked in WIRE-339.

huangminghuang (approved):
Reviewed current head 369b194e7f. No actionable findings remain; the prior review feedback is addressed, and the latest delta is documentation plus removal of an unused helper with no ABI or WASM change.

heifner:

[Run 31860422870](https://github.com/Wire-Network/wire-platform-build-system/actions/runs/31860422870) · `wire-sysio` @ `a0aceec3c3` · Release

| | ref |
|---|---|
| `wire-sysio` | `fix/sec-150-claimable-payouts` @ `a0aceec3c3` |
| `wire-libraries-ts` | `sec-150-sdk-types` @ `1129fb0` |
| `wire-tools-ts` | `sec-150-claim-flows` @ `86f1f0bc` |

```
✅ batch-operator-slashing (424s)      ✅ swap-non-native-tokens (2220s)
✅ batch-operator-termination (1022s)  ✅ swap-private-reserves (1877s)
✅ emissions-soak (2173s)              ✅ swap-to-wire (675s)
✅ node-owner-nft (555s)               ✅ swap-variance-revert (483s)
✅ operator-collateral-deposit (806s)  ✅ swap-with-underwriting (1080s)
✅ reserve-lifecycle (962s)            ✅ underwriter-slashing (902s)
✅ swap-from-wire (690s)               ✅ yield-distribution (732s)
```

### On the two earlier red runs

Both were in the harness, not in these contracts.

The first was the branch being 9 commits behind `master` — it was missing `e751620a9e` (resumable Solana dispatch driven from the depot) and `d11592395a` (relay hardening against epoch wedges), so every flow wedged at an epoch boundary. A control run with `wire-sysio=master` and the same companion branches reproduced the pass, confirming the branch content wasn't implicated; merging `master` cleared it.

The second was `swap-to-wire` alone, and it was two independent defects in the new claimable read added in wire-tools-ts Wire-Network/wire-tools-ts#62 — `getWireClaimable` querying `sysio.reserv::wireclaims`:

- `c50b6308` — the KV bound was built as a bare name string. `json=true` bounds go through `fc::json::from_string` → `be_key_codec::encode_key`, which needs a JSON **object** keyed by the ABI's `key_names`, so the node couldn't parse it at all.
- `86f1f0bc` — the fix above then set a lower **and** upper bound to the same encoded key. The upper bound is exclusive (`if (has_upper && kv >= ub_sv) break;`), so the range was empty and the read returned `0` for an account that had a balance. Now it sends the lower bound only, takes `limit: 1`, and confirms the returned row's key matches before using it.

Verified locally against `nodeop v1.0.0-a0aceec3c3` before this run: `flow-swap-to-wire` SUCCEEDED with zero failed steps, plus 28/28 `WireClient` unit tests and clean lint.

`contracts_unit_test` on this head: 636 cases, no errors.

PR #559, opened 2026-08-11
WIRE-331: Prevent batch-operator callback use-after-free

huangminghuang:
## Summary

- Retain the `read_write` API, action labels, and owned completion state through the deferred transaction callback.
- Preserve the bounded delivery timeout while making late and duplicate callbacks safe and non-throwing.
- Evaluate direct, immediate-error, deferred-success, and deferred-error push results before completing the waiting relay job.
- Add direct header dependencies, named one-shot test expectations, and a delayed-callback API-lifetime regression test.

## Validation

- Local Release: all 36 `test_batch_operator_plugin` cases passed, including delayed API-lifetime and deferred-error coverage; `unit_test` (1,515), `plugin_test` (285), `contracts_unit_test` (619), and `test_fc` (605) also passed.
- Current-head GitHub checks — pending after this push.
- Remote platform e2e — not required; the effective diff is limited to the `wire-sysio` batch-operator plugin and tests, with no system-contract, Ethereum, or Solana changes.


heifner (commented):
The core change is correct. It closes three dangling references that exist today in `push_action`:

- `std::promise<void> done` was a stack local captured by reference, so `set_value()` ran against a destroyed promise once `wait_for` timed out and the function returned.
- `contract` and `action_name` were captured by reference from the `const std::string&` parameters and dangled on the same path.
- `get_read_write_api()` returned a stack value, but `read_write::push_transaction` stores a raw `this` into the `transaction_async` pipeline and later dereferences `db` and `abi_serializer_max_time` through it.

The shared-state design holds up under the cases I checked. `push_action` keeps its own `completion` shared_ptr for the whole `wait_for`, so a callback destroyed without being invoked cannot make the future ready early through the broken-promise path; the timeout branch still behaves as before. `packed_var.get_object()` is consumed synchronously by `abi_serializer::from_variant` before `on_incoming`, so the const-ref parameter is safe. The `completed.exchange` guard covers the `CATCH_AND_CALL(next)`-after-`next(...)` path that previously threw `std::future_error` out of the relay callback.

A few things to address.

**Non-ASCII character in the rewritten `elog`.** The failure log carries a U+2014 em dash between `failed` and the format placeholder. That text predates this PR, but the line is rewritten by the diff and the repo convention is ASCII only in source. Please fold in `push {}::{} failed: {}` while you are here.

**`complete_push_result` treats the third variant alternative as success.** `next_function_variant<T>` is `std::variant<fc::exception_ptr, T, std::function<t_or_exception<T>()>>`. The helper tests only `get_if<fc::exception_ptr>` and takes every other alternative as success, so a deferred-call result would log `pushed X::Y ok` and release the waiter for a call that was never evaluated. Nothing produces that alternative for `push_transaction` on master today, so this is latent, but the fc-json-stream work turns the third alternative into `chain::deferred_call<T>` and the http_plugin phase-2 path already emits it, so the hazard goes live when that merges. A positive `std::holds_alternative<push_transaction_results>` test with an explicit third branch would be durable against that. Note that `push_recurse` in chain_plugin handles the same variant with `assert(0)`, which compiles out in release, so it is not a good template to copy.

**`BOOST_CHECK_NO_THROW` around a `noexcept` function does not assert anything.** `async_action_completion::complete` is `noexcept`, so if the internal `catch (...)` ever went away the throw would call `std::terminate` and abort the test binary rather than report a failed check. The case named for containing callback exceptions cannot detect the regression it targets; only its `wait_for(0ms) == ready` assertion carries weight, and the same applies to the late-completion case. Dropping `noexcept`, or routing the containment test through a non-`noexcept` wrapper, would make that guard real.

**`get_future()` is unguarded.** Every other member is `noexcept`, but a second call throws `future_already_retrieved`. There is one caller today, though this is now a shared header in `sysio::batch_operator_detail`, so it is worth either documenting the single-call precondition on the declaration or guarding it with the existing atomic.

Only the ASCII fix is blocking from my side; the rest can land as follow-ups if you prefer.

heifner (approved):

PR #560, opened 2026-08-11
fix(ci): serialize macOS gossip peering test

huangminghuang:
## Summary

- Exclude `auto_bp_gossip_peering_test` from the macOS ARM64 sharded NP/LR CTest invocation.
- Run that test in the following workflow step with `ctest -j 1`.

## Root cause

The scheduled macOS job [failed in run 31472683700, job 93719281319](https://github.com/Wire-Network/wire-sysio/actions/runs/31472683700/job/93719281319). It runs the NP/LR suite with three CTest workers, which allowed the 21-node gossip-peering topology test to start concurrently with other integration tests.

## Impact

The test remains part of the scheduled/manual macOS suite, but now has exclusive use of the macOS test host. Other NP/LR tests remain sharded.

## Validation

- Parsed `.github/workflows/macos_arm64_build.yaml` successfully.
- Confirmed `ctest -N -j 1 -R '^auto_bp_gossip_peering_test$'` selects only the intended test.
- Ran `git diff --check`.

### DIFF
--- a/.github/workflows/macos_arm64_build.yaml
+++ b/.github/workflows/macos_arm64_build.yaml
@@ -131,8 +131,18 @@ jobs:
           echo "Running sharded NP/LR tests with ${test_jobs} jobs"
           ctest -j "${test_jobs}" \
             -L "(nonparallelizable_tests|long_running_tests)" \
+            -E "^auto_bp_gossip_peering_test$" \
             --timeout 2700 \
             2>&1 | tee "$RUNNER_TEMP/macos-arm64-np-lr-ctest.log"
+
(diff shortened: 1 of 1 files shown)


heifner (commented):
One request: please add `--output-on-failure` to the sharded NP/LR step too. The parallel step above it sets it, and so does the equivalent step in `linux_amd64_build.yaml`. Without it the failing test's ctest output never reaches the run log, which is why diagnosing this one required pulling the 143 MB test-logs artifact.

heifner (approved):

PR #561, opened 2026-08-12
fix(msgch): rebuild sysio.msgch.wasm from its current source

heifner:
The committed `sysio.msgch.wasm` does not match what its own source compiles to. Building `contracts/sysio.msgch` unchanged produces **155,599** bytes; the committed binary is **188,357**.

## Where it diverged

`6414266929` ("Address final architecture review findings") changed `sysio.msgch.wasm` by **+33,462 bytes while touching no msgch source** — no `src/`, no `include/`, no OPP proto. Its only other files were `sysio.dispatch_tests.cpp` and two `underwriter_plugin` files.

Every prior refresh moved the wasm and its source together in a single commit. The size progression across those is smooth, and this rebuild continues it:

| commit | date | bytes |
|---|---|---|
| `49e9f1cb7d` | Jul 15 | 153,542 |
| `62b91e2ad0` | Jul 16 | 153,583 |
| `51ccc330c8` | Jul 21 | 154,374 |
| `66757dcce3` | Jul 24 | 154,895 |
| `6414266929` | Aug 7 | **188,357**  ← +33,462, no source change |
| this PR | — | **155,599**  ← same source, rebuilt |

The source itself last changed in `a6b5d50ff4` (Aug 6, "prevalidate underwriter UIC signatures") — one commit *before* the wasm-only bump, and without the wasm.

## Why the build environment here is the consistent one

This toolchain reproduces `sysio.epoch`, `sysio.roa`, `sysio.reserv` and `sysio.authex` **byte-for-byte** against master. So the discontinuity sits in the Aug-7 artifact, not in the environment producing this rebuild.

## What is actually wrong with the current artifact

Not that it is untested — CI runs the checked-in wasm, so 188,357 has been exercised and is green. The problem is that **it does not correspond to its source**: CI has been validating a binary that cannot be reproduced from the tree, which means the source a reviewer reads is not the code that ran. This PR restores that correspondence, and this PR's own CI run is what validates the rebuilt binary.

## Scope

- No source change. ABI untouched (last changed `8347d1be4d`, Jul 10), so nothing downstream needs regenerating.
- `contracts_unit_test` against this rebuilt binary: **633/633, no errors.**

Found while syncing `feat/underwriter-challenge` (#549) with master; unrelated to that work, so it is split out here rather than folded into that PR.


huangminghuang (approved):
Approved — rebuilt `sysio.msgch.wasm` is byte-for-byte reproducible from the current source and validates successfully.

PR #563, opened 2026-08-13
contracts: bound chklocks' sweep (epoch-stall risk) + materialize the per-bucket locked total

heifner:
Two changes, both consequences of the same underlying fact: **uwrit locks are held for the full wall-clock challenge window (`collateral_lock_duration_ms`) and are never released by delivery**, so an underwriter's live lock set is (settlement rate × lock duration) — a quantity with no bound in either the sweep that releases it or the scan that reads it. The contract default is 12h; dev clusters shorten it, which changes the magnitude but not the shape. One commit fixes the resulting *sweep* cost, the other the *read* cost. They are separate commits so the rollup can be reverted without losing the stall fix. The rest of the branch is a merge of master and the review follow-ups it required — sharing the bucket-key derivation, making the sweep budget challenge-aware, decrementing the rollup from `sweeplocks`, and the docs those changes invalidated.

---

# 1. `contracts: bound chklocks' per-epoch lock release`

## The defect

`sysio.epoch::advance` inlines four maintenance sweeps. Every one is budgeted against advance's hard, uncatchable transaction CPU deadline — **except `chklocks`**, which collected and released *every* expired lock in one unbounded pass.

The codebase already knew the hazard. `MAX_FWQ_DRAIN_PER_EPOCH`'s own comment states it, and names chklocks as sharing the ceiling:

> That transaction's CPU budget (~150 ms) is a hard, uncatchable deadline, so an oversized queue would abort every advance and **permanently stall epoch progress chain-wide**. […] Conservatively sized to stay well under the transaction CPU ceiling **shared with chklocks** / buildenv / emissions.

chklocks was counted as a consumer of that ceiling but never given a budget.

## Why it's the sharpest case, not the mildest

Lock expiry is **bursty by construction**: every lock is stamped `now + collateral_lock_duration_ms` when its race is won, so a burst of settlements inside one epoch produces a burst of expiries inside one epoch, exactly one lock-duration later. Per expired lock the sweep does an inline `opreg::releaselock` plus an erase. Unbounded, a large enough burst aborts `advance` — and those locks are still expired at the next advance, so it aborts **identically every epoch thereafter**. A permanent chain-wide epoch stall; a wedge, not a transient.

## The fix

`chklocks(max_rows)` walks `byexpire` ascending, mirroring `pruneuwreqs` / `drainfwq`: oldest-expiry-first so the bound is a FIFO drain and no lock starves, `max_rows == 0` is a no-op, and `advance` passes `MAX_LOCK_RELEASE_PER_EPOCH` (32). An oversized burst now costs bounded release *latency* — harmless, since the challenge window has already closed by the time a lock is swept.

---

# 2. `contracts: materialize the per-bucket locked total in uwrit`

## The defect

`available()` = balance − locked − pending, and the `locked` half was derived by **walking every lock row** the underwriter held — `sum_locks_inline`, implemented twice (uwrit and opreg), each documenting the assumption that *"per-underwriter lock counts are O(1)-ish so the scan is cheap"*.

That assumption is false for the reason above. Worse, the scan ran on the worst possible path: **per candidate** inside `try_select_winner` (up to `MAX_UWREQ_CANDIDATES` per uwreq), several uwreqs per dispatch transaction, inside the same consensus-dispatch CPU budget whose overrun stalls the chain.

## The fix

A `locksums` KV table holding the materialized sum per `(underwriter, chain_code, token_code)`, keyed by a `checksum256` over the packed triple — the **same derivation** as `lock_entry::by_underwriter_ck()`, so the key identifying a lock's bucket and the key addressing that bucket's total cannot diverge.

**One writer.** Every path that can change a bucket lives in uwrit. There are THREE: `try_select_winner` ADDS (one lock per required leg, on a win), `chklocks` DECREMENTS (healthy release at expiry), and `sweeplocks` DECREMENTS (erasing a commitment's held locks on an UPHELD underwriter-fault challenge, WIRE-297). A row is erased at zero, so an absent row reads as zero.

`sweeplocks` runs OUTSIDE `chklocks`, which is why it was missed on the first pass at the master merge — this section previously called `chklocks` the sole erase path, and `sweeplocks` erased rows without decrementing. Getting it wrong is permanent and silent: the rollup is authoritative for `available()`, so a bucket left positive after its last row is gone suppresses that collateral forever, because the rows that would decrement it are already erased. Any new erase path inherits the obligation; nothing structural enforces it. Both `sum_locks_inline` rollups become one row read. `opreg::has_active_locks` is unchanged — existence-only, already O(1).

**Overflow direction differs by operation, deliberately:**
- *increment* **saturates** (matching the scan it replaces) — a wrap would understate `reserved` and overstate availability, the one direction admitting an overcommit; saturating overstates `locked`, failing closed.
- *decrement* **clamps at zero** — a wrap would strand a colossal `locked` and zero that underwriter's `available()` permanently, inside `chklocks`, which runs inline in `advance` and must never throw.

## Checking the cache against its source

Rather than keeping a second production derivation that could itself drift, the dispatch-test fixture gained `scan_lock_total()`, recomputing a bucket from the authoritative `locks` rows. The lifecycle test asserts **rollup == scan** after the win, after a partial (budgeted) sweep, and at zero once the bucket empties.

---

## Tests

- Existing `chklocks` cases carry the new signature; `chklocks_zero_budget_is_noop` joins its `pruneuwreqs` twin.
- `chklocks_budget_bounds_the_sweep_and_backlog_drains` drives a real won race to two live locks, ages past expiry **without** an advance, then asserts budget 0 releases nothing, budget 1 releases exactly one, the remainder drains on the next sweep, the bond is left whole (`releaselock` is a no-op for a healthy operator), and the rollup tracks the scan at every stage.
- `chklocks_budget_counts_held_locks_and_still_drains` covers the challenge-aware budget that arrived with the master merge: two HELD locks from a challenged commitment with two ordinary expired locks queued behind them, swept with a budget of 1. The budget is spent inside the held run so nothing is released and the ordinary locks are never reached; the pokes it affords lapse the challenge, and the same bounded sweep then drains the backlog over later ticks. Budget-smaller-than-the-held-run is the discriminator — the release-counting form of the bound fails its first assertion.
- `chkuwchal_uphold_slashes_and_returns_bond` gained the rollup half: once the swept locks are confirmed gone, the rollup must equal an authoritative scan of the lock rows, and be zero. Removing the `sweeplocks` decrement fails it with the scan reading zero against a cache still holding the swept amount.

## Provenance

Found while investigating WIRE-340. **That ticket's root cause is still open** — a hypothesis I posted there (underwriter bond exhaustion) was refuted by direct measurement and I have retracted it. These two defects are independent of it: `chklocks` being the only unbudgeted sweep `advance` inlines, and `available()` scanning the lock table per candidate, are both properties of the contract that hold regardless of what stalled those runs.

**Corrected from an earlier revision of this description**, which claimed WIRE-340's ~50-minute runs "never reached the 12h expiry" so neither path had been exercised at volume. That was wrong: the harness overrides the window to 10 minutes (`ClusterBuildDefaults.ts:55`, `CollateralLockDurationMs = 600_000`), so those runs swept locks roughly five times each. Both paths *have* run — and survived, because the bursts were small. Measured peak on those clusters was ~122 concurrent locks for the underwriter, so a sweep released on the order of a hundred rows and the per-candidate scan walked a similar count. That is comfortably inside the CPU budget, which is exactly why nobody has hit either problem.

Both remain latent rather than theoretical: the quantity that matters is (settlement rate × lock duration), and neither the sweep nor the scan had a bound. At the 12h contract default, or at higher throughput, the same code paths grow without limit — the sweep into an `advance` abort that repeats every epoch, the scan into the consensus-dispatch CPU budget.

## Notes for review

- **ABI:** `chklocks` gains `max_rows: uint32`; uwrit gains the `locksums` table. Regenerated downstream in **Wire-Network/wire-libraries-ts#67** (generated output only, 152 insertions / 2 deletions). `wire-tools-ts` needs no source change — nothing there calls `chklocks` and `locksums` is additive; verified by building and running its gate against the regenerated sdk-core.

  *(Correction to an earlier revision of this description: it claimed the checked-in `SysioContractTypes.ts` was already out of sync with master's uwrit ABI, listing `freelocks`/`holdlocks`/`sweeplocks`. That was wrong — `origin/master`'s generated file matches master's ABI exactly. Those actions come from the `feat/underwriter-challenge` branch, which is what I had checked out when I looked.)*
- **Artifacts:** only contracts whose source changed are refreshed — `sysio.uwrit` + `sysio.epoch` in commit 1, `sysio.uwrit` + `sysio.opreg` in commit 2. `sysio.opreg.abi` is untouched (no ABI surface change, internal reader only).


huangminghuang (commented):
Two inline findings from the lock-rollup review.

huangminghuang (commented):
Re-review after the underwriter-challenge work landed on master found two integration blockers that must be handled while resolving the current conflicts.

huangminghuang (dismissed):
Final re-review of head 0902f94174: no actionable findings. Verified the challenge-aware examined-row budget, both locksums decrement paths, shared bucket-key derivation, regression coverage, generated ABI/artifact scope, and the completed lifecycle documentation follow-ups.

heifner:

[Run 31843613736](https://github.com/Wire-Network/wire-platform-build-system/actions/runs/31843613736) · `wire-sysio` @ `55dd7e9f9e` · Release

Dispatched after merging `origin/master` into the branch — it was 9 commits behind, and those commits carry the resumable-Solana-dispatch work (`e751620a9e`, `d11592395a`) without which every flow wedges at epoch 1. Companion `wire-libraries-ts#67` (`fix/chklocks-bounded-sweep-types`) was pinned for the regenerated `chklocks` / `locksums` types; everything else floated to manifest defaults.

| | ref |
|---|---|
| `wire-sysio` | `fix/chklocks-bounded-per-epoch-sweep` |
| `wire-libraries-ts` | `fix/chklocks-bounded-sweep-types` (#67) |

```
✅ batch-operator-slashing (458s)
✅ batch-operator-termination (965s)
✅ emissions-soak (2211s)
✅ node-owner-nft (560s)
✅ operator-collateral-deposit (793s)
✅ reserve-lifecycle (946s)
✅ swap-from-wire (693s)
✅ swap-non-native-tokens (2265s)
✅ swap-private-reserves (1742s)
✅ swap-to-wire (703s)
✅ swap-variance-revert (496s)
✅ swap-with-underwriting (1094s)
✅ underwriter-slashing (916s)
✅ yield-distribution (733s)
```

`contracts_unit_test` on the merged branch: 636 cases, *** No errors detected.

Relevant to this PR specifically: `underwriter-slashing` (933s) and `batch-operator-termination` (1007s) exercise the collateral-lock lifecycle across many epoch boundaries, which is the surface the bounded `chklocks` sweep touches.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -370,28 +370,23 @@ void opreg::regoperator(name account,
...
-/// rows are filtered on `(chain_code, token_code)` in memory. Per-underwriter
-/// lock counts are O(1)-ish in steady state so the scan is cheap.
+/// This used to scan `sysio.uwrit::locks` through its `byuw` index and filter
+/// `(chain_code, token_code)` in memory, on the stated assumption that
+/// per-underwriter lock counts are "O(1)-ish in steady state so the scan is
+/// cheap". That assumption does not hold: uwrit locks are held for the full
+/// wall-clock challenge window and are never released by delivery, so a
+/// bucket's live lock count is (settlement rate × lock duration). uwrit now
+/// maintains the total at the two sites that can change it; see
+/// `uwrit::lock_sum`.
```
huangminghuang: **[P3] Keep this on the three-path invariant.** This says “two sites,” but the rollup is mutated at three: `try_select_winner` adds, `chklocks` decrements, and `sweeplocks` decrements after an upheld challenge. That third path was the source of the stale-cache bug fixed in this PR, so leaving the old count here undercuts the invariant documented in `lock_sum`. Please update this to say three sites and name them. The adjacent `locks_t` comment in `sysio.uwrit.hpp` also still says `opreg::available()` scans `byunderwriter`; it should identify `locks` as the authority and `locksums` as the O(1) read cache.

heifner: Fixed in bc881ddfff.

You are right on both, and the second one is the more useful catch: the `locks_t` comment was still describing the scan that `locksums` exists to replace, so the file taught the old model at the point a reader is most likely to look it up.

`sum_locks_inline`'s comment now enumerates all three with direction, and states why the count matters at this site specifically — it is the cross-contract reader, it trusts the rollup completely, and a bucket left positive after its last row is gone suppresses that collateral permanently.

`locks_t` now names `locks` as the AUTHORITY and `locksums` as the O(1) read cache `available()` reads, with the reason the old scan does not scale: locks are held for the full wall-clock challenge window and are never released by delivery, so a bucket's live count is (settlement rate × lock duration).

I swept for the claim rather than fixing only the two sites you named, and three more had `available()` reading the lock rows:

- the file header's opening bullet, which had opreg reading "this table via a mirror"
- the `lock_entry` doc, which called the triple the indexing surface `available()` uses and pointed cross-contract at `locks_t` rather than `locksums_t`
- the README's `locks` row ("consulted by `sysio.opreg::available()`") and its integration note — both contradicting the `locksums` row one line below, which already said the read is O(1) instead of a scan

Two adjacent bits of staleness turned up while there. The header bullet described the composite as one of "two secondary indexes", `byuwck` and `byunderwriter` — neither exists; the composite is deliberately not a table-managed index (the `locks_t` comment says so directly), and the uint64 indexes are `byuw`, `byuwreq` and `byexpire`. And `lock_entry` said its rows are erased by `release`, an action this contract does not have — the erase sites are `chklocks` and `sweeplocks`, which is also where the two `sub_locked_total` calls live.

Comments and markdown only. `sysio.uwrit.wasm`, `sysio.opreg.wasm` and both `.abi` files rebuild byte-identical, so this commit carries no artifact.

huangminghuang (dismissed):
Re-reviewed current head bc881ddfff. No actionable findings remain; the documentation follow-up resolves the stale rollup model without changing runtime behavior.

huangminghuang (approved):
Re-reviewed current head 20e9ed60ee after the master merge. No actionable findings remain: the textual PR patch is unchanged from the previously reviewed revision, the merge-base diff passes the delegated rules/checks pass, and current-head CI is green.

PR #564, opened 2026-08-13
Test: stop nodeop_forked_chain_test failing on missed slots

heifner:
Two commits against the preamble of `nodeop_forked_chain_test`, the walk that establishes the production cycle before the network is forked.

## 1. `test: report the correct missed-slot window in nodeop_forked_chain_test`

The missed-slot diagnostic printed a time range that ended nearly a day before it began. The ubsan failure on run 31732928473 reported:

    Slots were missed after the following blocks: 392 (2026-08-14T07:12:44 thru 2026-08-13T07:12:54.5)

`timedelta`'s first positional parameter is `days`, so `timedelta(0.5)` is twelve hours rather than half a second: the start of the window was pushed 12h forward and the end of it 12h back. The same two blocks now report `392 (2026-08-13T19:12:44.500 thru 2026-08-13T19:12:54.000)`, the 20 slots that actually went unproduced.

## 2. `test: derive the production cycle from slot windows, not runs of blocks`

The walk established the cycle by counting how many blocks each producer signed in a row and requiring between 10 and 12. That is a proxy for the schedule rather than the schedule itself, and a loaded machine breaks the proxy. On that same ubsan run one node's block-processing thread stalled for 10.5 seconds: `defproducern` signed 5 of its 12 slots, `defproducero`'s window elapsed with no block at all, and the test failed even though every block that was produced was produced by the right producer at the right time. Block 392 carried no transactions, and the node was stalled between `accepted_block_header` and `accepted_block`, so nothing about the chain was at fault.

Each producer owns one contiguous window of 12 slots. An unproduced slot leaves a hole inside its window; it never moves a boundary and never hands a slot to the next producer. So each block is now attributed to its window and the schedule is checked directly:

- a window belongs to exactly one producer
- a producer appears at most once per cycle
- any handover into the immediately following slot lands at the start of a window, which pins producer runs to a single grid and catches a run of the wrong length

The grid is anchored on an observed block rather than an absolute slot number: a handover into the very next slot can only happen on a boundary, so the first such block identifies offset 0 and the test needs no knowledge of the chain's block timestamp epoch. Unproduced slots and entirely silent windows are reported rather than failing the run, since they describe the machine the test is running on and not the schedule.

This also removes a second spurious failure mode. The old walk terminated on seeing the first producer of the cycle again, so a producer whose entire window was lost shifted the wrap point and tripped the "repeated in slot" check instead.

The trade is deliberate: the old check would notice a producer that merely produced fewer blocks than scheduled, and the new one will not, because that is exactly the environmental symptom. What replaces it is a direct check of the schedule rule itself.

## 3. `test: let the schedule walk wait out a missed-slot stall`

Addressing the review finding. The walk tolerates missed slots in its accounting, but it still has to fetch every block, and `HeadWaiter` aborted that fetch after four seconds once caught up to the head. A stall long enough to skip slots is long enough to end the walk before it can record the gap, so the flake survived in the fetch path:

    BEFORE (leeway=3)   caught up to head    timeout= 4.0s vs 10.5s stall -> ABORTS
    AFTER  (leeway=60)  caught up to head    timeout=61.0s vs 10.5s stall -> waits it out

The leeway is now an argument rather than a fixed three seconds, since it is what bounds how long the chain may go dark before the test calls it dead. This did not fire in the recorded incident only because the walk happened to be running behind the head at that moment; under slightly different timing it would have.

## Verification

`nodeop_forked_chain_test` passes locally end to end, on two separate runs. Each hit four missed-slot gaps of its own, and each left two producers at 10 or 11 blocks against the old lower bound of 10, so the flake reproduced incidentally on an idle machine both times. All 21 producers were found in correct rotation with no silent windows and every handover on a window boundary, which confirms the anchor-derived grid against a real chain.

The block sequence from the failed ubsan run was also extracted from the archived node logs and replayed through the new walk, which accepts it and reports `defproducern:5` with `defproducero` silent. A block moved into another producer's window, a producer scheduled twice, and a producer given a 13 slot run are each still rejected.

### DIFF
--- a/tests/nodeop_forked_chain_test.py
+++ b/tests/nodeop_forked_chain_test.py
@@ -351,7 +351,7 @@ try:
             slotTime=0.5
             slotsDiff=int(timediff.total_seconds() / slotTime)
             if slotsDiff != 1:
-                slotTimeDelta=timedelta(slotTime)
+                slotTimeDelta=timedelta(seconds=slotTime)
                 first=lastTimestamp + slotTimeDelta
                 missed=first.strftime(Utils.TimeFmt)
                 if slotsDiff > 2:
--- a/tests/nodeop_forked_chain_test.py
+++ b/tests/nodeop_forked_chain_test.py
@@ -73,8 +73,13 @@ def censusProductionCycle(getBlock, firstBlockNum, windowPhase, windowSize, cycl
     windowOfSlot=lambda slot: (slot - windowPhase) // windowSize
     cycle=[]
     counts={}
+    # What the censused rotation established, and separately what blocks past it merely suggest.
+    # Kept apart because the two carry different authority: an inferred owner is enough to hold later
+    # blocks to a consistent story, but not to certify that the rotation came round.
     windowToProducer={}
     producerToWindow={}
+    inferredOwner={}
+    inferredPosition={}
     blocksExamined=0
     result={"cycle":cycle, "counts":counts, "silentWindows":[], "wrapProducer":None, "blocksExamined":0, "error":None}
 
@@ -88,7 +93,6 @@ def censusProductionCycle(getBlock, firstBlockNum, windowPhase, windowSize, cycl
     firstWindow=windowOfSlot(slot)
     endWindow=firstWindow + cycleLength   # exclusive, one full rotation
     lastWindow=firstWindow
-    silentPositions=None
 
     while True:
         window=windowOfSlot(slot)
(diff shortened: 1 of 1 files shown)


huangminghuang (commented):
Two remaining test-correctness findings from the re-review.

huangminghuang (commented):
Three remaining test-correctness findings from the re-review of 85f7177.

huangminghuang (commented):
One remaining test-correctness finding from the re-review of 91c6f86.

huangminghuang (commented):
One remaining correctness issue:

huangminghuang (approved):
Re-reviewed the latest head; no actionable findings.

PR #565, opened 2026-08-13
Test: Producer: drive the production timer from a mockable clock

heifner:
## Summary

Block production is driven entirely by the produce timer, and nothing covered the path from arming that timer to a block actually being produced: `block_timing_util`'s tests decide which slot to wake for and stop there, and the integration tests race a real clock. A slot lost between the two emits no log line and no error, it simply produces no block.

This puts the production timer on a clock a test can move, so that path becomes deterministic.

## The seam

`fc::time_point::now()` already consults `fc::mock_time_traits`, so test code can move "now" deterministically. A Boost.Asio timer keeps waiting on the real clock regardless, which is what made production timing untestable.

`fc::system_timer` (new, `libraries/libfc/include/fc/system_timer.hpp`) closes that gap. It is a `boost::asio::basic_waitable_timer` over two small pieces:

```cpp
struct system_clock {                                  // fc's clock, as a Chrono clock
   static time_point now() noexcept { return from_time_point( fc::time_point::now() ); }
};

struct system_timer_wait_traits {                      // how long Asio may sleep before re-reading it
   static constexpr std::chrono::milliseconds mock_poll_interval{1};

   static system_clock::duration to_wait_duration( const system_clock::duration& d ) {
      return mock_time_traits::is_set() ? duration_cast<system_clock::duration>( mock_poll_interval ) : d;
   }
};

using system_timer = boost::asio::basic_waitable_timer<system_clock, system_timer_wait_traits>;
```

`to_wait_duration` is the whole trick. Asio asks how long it may sleep before re-evaluating the clock. Under the real clock the honest answer is the entire remaining delay, so the timer sleeps once and wakes at its deadline exactly as `boost::asio::system_timer` did -- production timing is unchanged. Under mock time that answer would be wrong, because virtual time only moves when a test moves it, so a short poll is returned instead and each wake re-reads the virtual clock; the timer then fires as soon as a test advances past the expiry. `set_now()` sleeps for longer than the poll interval, so a caller that advances the clock can rely on the io_context having had the chance to notice.

Production and test therefore share one timer type: no build variant, no test only option, no runtime polymorphism, only the latch `fc::time_point::now()` already consults.

`system_clock::from_time_point` goes through `fc::time_point::to_system_clock()` rather than scaling the microsecond count directly, because that conversion has to be clamped: this clock's duration is finer than a microsecond on some platforms (nanoseconds on Linux), so scaling `fc::time_point::maximum()` up overflows and wraps to a moment in the past. A deadline meant to be unreachable would become one that has already passed, turning a dormant timer into one that fires at once. `test_system_timer.cpp` covers that saturation at both extremes.

`fc::mock_time_traits::unset()` and `fc::scoped_mock_clock` are added alongside, so a test can release the mock clock rather than freezing it for whatever runs next in the same binary.

## Measured

- Real clock: a 200ms wait fires at 200ms.
- Mock clock: a 30 **second** virtual wait resolves in 2ms of real time.

## The test

`production_follows_the_virtual_clock` drives a round on the virtual clock: real time alone moves nothing, the head keeps advancing for as long as the clock does, and stopping the clock stops production.

What that establishes is **bounded liveness**, not a per-slot guarantee. Blocks ship at their cpu-effort deadline (462.5ms at the default offset) rather than at the 500ms slot boundary, so a node runs progressively further ahead across a round and then waits out a gap of roughly two slots. Requiring a block from every slot would fail on correct behaviour, at a step that moves with startup timing. The case asserts instead that the head never stalls longer than that legitimate gap, twelve times running, and that twelve advances yield at least twelve blocks.

It follows that a regression losing a single deadline is not caught: the head still moves within the allowance, just later. Catching that needs a slot budget across the whole round, and a fixed budget is brittle for the same reason a per-slot assertion is, since accumulated drift depends on where in the round the node started. What the case does catch is a timer that stops re-arming.

Reverting the timer type fails it both ways -- production runs free on real time (`11 != 2`) and advancing the virtual clock produces nothing -- so the test measures the seam rather than passing incidentally.

The clock is engaged through an RAII guard, and the suite was run in first position to confirm it does not depend on running last: all cases pass with `test_trx_full`, which needs the real clock, running immediately after it. Worth noting for future work that this fragility pre-exists -- `test_block_timing_util` already engages the mock clock without releasing it.

## What this does not cover

`production_survives_a_competing_reschedule` gates the app thread so the delayed timer fires and queues its lambda behind a competing `get_integrity_hash()`, which is the interleaving the correlation id recheck is written for. It does not enforce the recheck, and the case says so in its own docstring.

Nothing available to a test observes the stale post. The executor's queue depth cannot grow while the gate holds the app thread, because a `read_write` post only enters the priority queue when that thread runs the wrapper; the skip itself is an `fc_dlog` line, and `logger::add_sink` is private with no log capture used anywhere in the tree.

Removing either recheck also changes nothing measurable. Without `schedule_maybe_produce_block`'s, block counts are identical across repeated runs; without `schedule_delayed_production_loop`'s, the node recovers in the same ten slots. `maybe_produce_block` wraps its work in `LOG_AND_DROP`, aborts and reschedules, and every route into `schedule_production_loop` cancels and re-arms the timer, so a starved timer waits for the next natural rescheduling point instead of stalling. Both rechecks are defence in depth against a transient that heals itself.

The case is therefore a check that the interleaving is survivable, not regression coverage for the recheck. Enforcing it would need a counter or seam in the plugin.

## Scope

This makes the *production timer* virtual, not the whole node; anything else consulting the real clock inside a running plugin still runs on wall time.

`_ro_timer` stays on `boost::asio::system_timer`. It was moved to the mockable timer earlier in this branch together with a read window case, and that case was dropped as flaky -- whether the window opens turns on a race inside the plugin between the main thread draining the read only queue and the timer firing, and `switch_to_read_window` returning to the write window on an empty queue re-arms against a clock only the test moves. With the coverage gone the type change was an untested behaviour change no test in this branch needs, so it is reverted rather than carried. Making the read window deterministic is left to a follow-up, together with the test that would justify the change.

## Notes for review

The libfc side is one new header (`system_timer.hpp`) with its own test, plus `unset()` and `scoped_mock_clock` on the existing `mock_time.hpp`. Nothing existing changes behaviour.

`test_fc` has 4 failures on my machine, all in `test_http_client.cpp` DNS and connect cases. That file has no `mock_time` references and does not include the new header; these are the known WSL2 DNS flakes rather than fallout from this change. Flagging rather than claiming green.


huangminghuang (commented):
Code review findings are posted inline.

huangminghuang (commented):
Follow-up review findings are posted inline.

huangminghuang (commented):
Three additional findings are posted inline. The previous stale-CID coverage finding remains unresolved; I replied in that thread with the requested disposition.

Please also refresh the PR description before approval: it still documents the removed `mockable_time_traits::to_posix_duration` API, claims the production test requires one block per slot although it now permits three slots for head movement, and says `_ro_timer` remains on the real system clock although the diff changes it to `fc::system_timer`.

heifner:

[Run 31843640588](https://github.com/Wire-Network/wire-platform-build-system/actions/runs/31843640588) · `wire-sysio` @ `b5a0c5cd32` · Release

Dispatched after merging `origin/master` into the branch (it was 9 behind). No companion PR needed — this changes `producer_plugin` and `libraries` only, touching no contracts and no ABI.

| | ref |
|---|---|
| `wire-sysio` | `feature/producer-plugin-mockable-timer` |

```
✅ batch-operator-slashing (467s)
✅ batch-operator-termination (976s)
✅ emissions-soak (2188s)
✅ node-owner-nft (554s)
✅ operator-collateral-deposit (809s)
✅ reserve-lifecycle (917s)
✅ swap-from-wire (690s)
✅ swap-non-native-tokens (2098s)
✅ swap-private-reserves (1726s)
✅ swap-to-wire (643s)
✅ swap-variance-revert (482s)
✅ swap-with-underwriting (1067s)
✅ underwriter-slashing (935s)
✅ yield-distribution (733s)
```

Worth noting for a production-timer change: every flow here drives a live multi-producer cluster through repeated epoch boundaries, and the long ones — `emissions-soak` (2172s), `swap-non-native-tokens` (2249s) — sustain block production for 35+ minutes without a missed-slot or scheduling regression surfacing.

plugins/producer_plugin/test/test_mockable_timers.cpp
```diff
@@ -0,0 +1,653 @@
...
+BOOST_AUTO_TEST_CASE(production_follows_the_virtual_clock) {
+   running_node   node;
+   const uint32_t settled = node.blocks_produced();
+
+   // 1. Real time alone must move nothing. This is the property that makes the test deterministic:
+   //    a loaded machine cannot advance the chain behind the test's back.
+   std::this_thread::sleep_for(real_time_patience);
+   BOOST_CHECK_EQUAL(node.blocks_produced(), settled);
+
+   // 2. Every slot the clock advances through must yield a block. Stepping one slot at a time is
```
huangminghuang: **[P2] Match the assertion to bounded liveness**

`advance_until_head_moves()` may advance three virtual slots before requiring a single head increment, so this loop does not make a skipped slot fail as the comment and PR description claim. A timer regression that loses one or two deadlines can be hidden by a later block. Please either assert a valid slot/head invariant, or describe this case consistently as bounded liveness and remove the per-slot/no-skipped-slot claims from the test and PR description.

huangminghuang (commented):
Re-reviewed current head `b5a0c5c`. The `e1e1c62` ownership and teardown fixes are sound; timer arithmetic, cancellation, and real-clock semantics look correct. Required Linux/macOS CI and the 14-flow platform gate are green. I accepted the retained CID case as explicitly scoped survivability coverage. Two P2s remain: the central test claims stronger per-slot coverage than it has, and `_ro_timer` is still an uncovered scope addition. Also update **The seam** in the PR body: it still shows the obsolete `fc::mockable_time_traits::to_posix_duration`; the current implementation is `fc::system_clock` with `system_timer_wait_traits::to_wait_duration`.

plugins/producer_plugin/src/producer_plugin.cpp
```diff
@@ -929,7 +931,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                                                                    // use atomic for simplicity and performance
    fc::time_point                 _ro_read_window_start_time;
    fc::time_point                 _ro_window_deadline;    // only modified on app thread, read-window deadline or write-window deadline
-   boost::asio::system_timer      _ro_timer{_timer_thread.get_executor()}; // only accessible from the main thread
+   // Follows the mock clock under test for the same reason _timer does: the read window is bounded
+   // by wall clock, so its cycling is only testable deterministically if a test owns the clock.
+   fc::system_timer               _ro_timer{_timer_thread.get_executor()}; // only accessible from the main thread
```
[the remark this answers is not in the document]
huangminghuang: Thanks for correcting the Scope text. This remains actionable: `_ro_timer` now reacts to the global mock clock, but no test covers the read-only task draining, futures, and window re-arming on that clock. Real-clock equivalence does not validate the newly enabled path. Since this change is not needed by any test in this PR, please retain `boost::asio::system_timer` here or add deterministic read-window coverage, and leave groundwork for a follow-up together with its test.

plugins/producer_plugin/test/test_mockable_timers.cpp
heifner: Agreed, and I took the second option: the per-slot claim was wrong, not just loosely worded.

`advance_until_head_moves` allows three slots deliberately. Blocks ship at their cpu-effort deadline -- 462.5ms at the default offset -- rather than at the 500ms slot boundary, so the node runs progressively further ahead across a round and then waits out a gap of roughly two slots. Asserting a block per slot would therefore fail on correct behaviour, at a step that moves with startup timing, which is why the allowance is there.

The docstring and the loop comment now describe bounded liveness, and they state what the case does not establish as plainly as what it does: a single lost deadline goes undetected, because the head still moves inside the allowance. Catching that needs a slot budget across the whole round, and a fixed budget is brittle for exactly the same reason a per-slot assertion is -- accumulated drift depends on where in the round the node started. What the case does catch is a timer that stops re-arming.

One thing I could tighten honestly: the final check was `blocks_produced() > settled`, which is weaker than the loop already requires. It is now `>= settled + slots_to_step`, since each of the twelve steps required a head move and this node is the only producer. Ran the case twelve times, no flakes.

The per-slot and no-skipped-slot claims are out of the PR description too. Fixed in 20cdcaaff2.

plugins/producer_plugin/src/producer_plugin.cpp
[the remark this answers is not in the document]
heifner: Reverting it, as offered.

You are right that real-clock equivalence does not validate the path the mock clock newly enables, and I confirmed nothing in the tree exercises the read-only window at all -- so it was an uncovered behaviour change that no test in this branch needs.

`_ro_timer` is back on `boost::asio::system_timer`. The comment describing mock-clock behaviour it no longer has is gone with it, and the explicit `<boost/asio/system_timer.hpp>` include is restored: the branch had dropped it when the last direct use went away, and it was compiling through the `boost/asio.hpp` umbrella, which is not something to rely on once there is a direct use again. Scope in the description updated to match.

Making the read window deterministic stays a follow-up, together with the test that would justify the type change. Fixed in 4099fc2b50.

huangminghuang (commented):
Re-reviewed current head `20cdcaaf`. The two prior P2s are addressed: `_ro_timer` is fully back on `boost::asio::system_timer`, the test and PR description now accurately describe bounded liveness, and the obsolete seam description is corrected. The timer implementation remains production-correct. One new P2 remains in the strengthened counter assertion, so I did not approve this head. Current-head CI is still running with no failures so far.

plugins/producer_plugin/test/test_mockable_timers.cpp
```diff
@@ -0,0 +1,671 @@
...
+   for (uint32_t step = 1; step <= slots_to_step; ++step) {
+      BOOST_REQUIRE_MESSAGE(node.advance_until_head_moves(),
+                            "production stopped at step " << step << " of " << slots_to_step
+                                                          << ": the head did not move across three slots of "
+                                                             "virtual time");
+   }
+
+   // Each of those steps required the head to move at least once, and this node is the only
+   // producer, so every advance is a block it produced.
+   BOOST_CHECK_GE(node.blocks_produced(), settled + slots_to_step);
```
huangminghuang: **[P2] Publish the block count before the head**

`advance_until_head_moves()` returns as soon as `record_head()` makes the new head observable, but the accepted-block callback increments `_blocks_produced` only afterward. On the final iteration this assertion can therefore read eleven increments after observing twelve head moves; the `held` sample below can likewise precede that late increment and make the frozen-clock equality fail. Please increment the counter before publishing the head, update both as one synchronized snapshot, or use `head_block_num()` consistently for these checks.

heifner: Confirmed and fixed in 6bd94dd0b4. I reproduced it rather than agreeing on inspection: restoring the two-step publish with a 5ms gap between the head and the count fails exactly the two checks you named -- `[11 < 12]` on the count assertion and `[12 != 11]` on the frozen-clock equality.

I took the second option. The count now lives in `head_state` and is updated under the same lock as the head, so `blocks_produced()` reads it from the same snapshot. Incrementing before publishing would have closed these two windows by ordering, but that pairing appears at a dozen call sites in this file, and the next one added would have to get the order right again -- sharing the snapshot removes the window instead of the symptom.

The assertion that exposed this was mine, added last round to tighten `> settled`. It was measuring something real and racing to do it. Worth recording that fifteen clean repeat runs did not catch it; the window is only microseconds wide unless deliberately widened.

huangminghuang (approved):
Re-reviewed current head `6bd94dd0`. The counter/head publication race is resolved by keeping the accepted count in the same mutex-protected snapshot as the head. The earlier scope, lifetime, test-validity, saturation, and timer-semantics findings are all addressed; the complete PR diff and description are clean. macOS and ASan checks pass on this SHA, with the remaining required checks still running and continuing to gate merge.

PR #569, opened 2026-08-18
contracts(msgch): derive consensus thresholds from the live batch-op group + re-drive evalcons past the boundary (WIRE-346, WIRE-322, WIRE-354)

heifner:
Fixes two independent ways an OPP epoch could stall **permanently** in `sysio.msgch`. Both are CertiK "Wire Network - Sysio Audit 1" findings, verified still present on `origin/master` before anything was changed.

**No ABI change.** No new action, table, or signature. Verified by rebuilding *every* contract and diffing against the committed artifacts: the only file that moved is `sysio.msgch.wasm` (151,948 → 153,620). `sysio.msgch.abi` is byte-identical and no other contract's `.wasm`/`.abi` changed — so the `SysioContractTypes` → wire-libraries-ts → wire-tools-ts regen chain does **not** fire.

| Ticket | Finding | Severity | |
|---|---|---|---|
| WIRE-346 | WNS-15 | **Major** | both defects below |
| WIRE-322 | WNS-08 | Medium | duplicate of (a), from the earlier 2026-07-31 assessment |
| WIRE-354 | WNS-21 | Informational | duplicate of (b) |

CertiK filed (a) twice across two assessments and rated (b) Informational in one report and Major in the other. All three tickets are one change.

## (a) Quorum came from the configured size, not the group that can deliver

`evalcons` derived both thresholds from `epochcfg.operators_per_epoch`. The set that can actually deliver is the resident group **intersected with current opreg ACTIVE status**:

- `advance()` fills a new tail only *up to* the configured size, and its own comment calls a short group *"a degraded but non-fatal state: batch operators were terminated faster than replacements could activate"*.
- `deliver()` refuses a member slashed or terminated mid-window (*"delivering operator is not ACTIVE in sysio.opreg"*) even though it remains in the `batch_op_groups` snapshot.

With the configured 7 and S eligible, unanimity was unreachable for any `S < 7` and the majority path for `S <= 3`. Consensus became arithmetically impossible while every remaining operator behaved correctly — no attacker required.

`active_batch_op_group_size()` now counts exactly the set `deliver()` admits, failing closed on a missing operator row the same way `deliver()` does.

Two details worth review attention:

- The majority test is `count * 2 > group_size`, **not** `count > group_size / 2` — a short tail can be **even**, where integer division would accept an exact half as a majority. `maybe_open_dispute`'s own majority check moved to the identical form so the two cannot disagree about what counts as resolved.
- `group_size == 0` returns early. Without it, `agreeing >= 0` would make the first delivery a trivial "majority".

This restores the model in `.claude/rules/opp-consensus.md`, which requires both thresholds to derive from `group_size` and explicitly forbids a stored scalar. The formula *shapes* were already correct; the **source** was not — and that rule predicts this exact failure ("breaks every flow where one group member can't deliver").

## (b) A pre-boundary majority was never re-examined

`evalcons` runs **only** as an inline of `deliver()`, and `deliver()` rejects a second delivery from the same operator. So a majority that formed before the epoch boundary, with no further delivery after it, had nothing on chain able to re-examine it once the time gate opened.

`maybe_open_dispute` is not an escape either — it requires 3+ distinct checksums **and** no majority, so a stalled majority is specifically excluded from the dispute path.

`chkcons` now collects every outpost lacking a current-epoch consensus row (previously it `break`'d on the first) and, past the boundary, inline-sends `evalcons` for each. It is already permissionless and cranked every tick by `batch_operator_plugin`, so no new action and no new caller were needed. The inline is sent as `msgch` itself, so `evalcons`'s `require_auth(get_self())` is unchanged — **no authority is widened**. Inlines run after the action returns, so the refreshed row is observed by the next tick.

## Tests

Two regression cases in `sysio_msgch_chain_tests`:

- `short_group_after_slash_still_reaches_consensus` — a port of the PoC CertiK filed on WIRE-322
- `pre_boundary_majority_finalized_by_chkcons_crank`

**Both were confirmed red before being accepted.** Against the unmodified contract with the tests in place, each fails with `critical check !opc.is_null() has failed` — no consensus row is ever written and the epoch never advances, i.e. the stall itself, reproduced. Green with the fix.

`contracts_unit_test -- --sys-vm`: **647 cases, `*** No errors detected`** (645 baseline + 2 new).

## Deliberately not included

- **WIRE-348** (WNS-18, unacked envelopes pruned) — same file, but a *retention-model* change. `buildenv`'s one-deep invariant is load-bearing (it reconstructs `previous_envelope_hash` from "the single surviving row"), and "retain until acknowledged" has no acknowledgement signal today: `outpcons` records inbound consensus, not outbound receipt. Needs a design decision first.
- **WIRE-304** (no chain-deactivation path) — closely related, and it touches this same `chkcons` gate, but its remedy is a new `deactivchain` action on `sysio.chains`, i.e. an ABI change and the full four-repo regen. It also carries an unresolved auto-drop-vs-governance decision. Keeping it separate so this Major is not held behind it.


huangminghuang (commented):
Two blocking consensus issues found in the live-group quorum change.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1407,22 +1448,32 @@ void msgch::evalcons(uint64_t chain_code, uint32_t epoch_index) {
...
 
    // Consensus check
    bool consensus_reached = false;
    size_t consensus_group = 0;
 
    for (size_t g = 0; g < seen_checksums.size(); ++g) {
       // Option A: ALL operators delivered identical data
-      if (checksum_counts[g] == operators_per_group &&
-          total_deliveries == operators_per_group) {
+      if (checksum_counts[g] == group_size &&
```
huangminghuang: `active_batch_op_group_size()` removes a slashed operator from `group_size`, yet the tally above still counts that operator's pre-slash row. With A/B/C: A delivers X, A is slashed, then B delivers X. `group_size == 2`, `checksum_counts[X] == total_deliveries == 2`, and Option A accepts immediately even though C—one of the two current ACTIVE members—never delivered. Filter the numerator and total to the same live set as the denominator (or use one consistent epoch snapshot), and add a slash-after-delivery regression.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1159,9 +1198,11 @@ void maybe_open_dispute(name self, uint64_t chain_code, uint32_t epoch_index,
    for (auto c : checksum_counts) {
       if (c > max_count) max_count = c;
    }
-   if (max_count > operators_per_group / 2) {  // a majority exists -> no vote needed
+   // Same strict-majority form as evalcons Option B -- these two MUST agree, or a split that
+   // evalcons treats as un-resolved could be declined a dispute here (or vice versa).
+   if (max_count * 2 > group_size) {  // a majority exists -> no vote needed
```
huangminghuang: This now supports even shortened groups, but a two-member 1–1 split has neither a majority nor a recovery path. After C is slashed from A/B/C, A and B can deliver different envelopes; `maybe_open_dispute` returns for fewer than three candidates and `sysio.chalg::opendispute` rejects fewer than three as well. Every later `chkcons` re-drive repeats the unresolved state, with no eligible operator left to change it. Add a two-candidate dispute escape (including `sysio.chalg`) or preserve an odd-quorum invariant.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: Confirmed and fixed in 95d7881050 — this was a real defect, and your A/B/C walkthrough is exactly right.

I reproduced it before fixing it: with the previous commit's contract and the new regression test in place, `consensus_reached` comes back **true** after only A(slashed) + B, with C never heard from. Test fails at the `!consensus_reached` assertion, precisely as you described.

`active_batch_op_group_size()` is now `eligible_batch_operators()` and returns the set rather than a count. It is resolved once, before the tally; the tally skips deliveries from operators outside it, and the threshold is its size. Numerator and denominator therefore come from one snapshot and cannot diverge — the "one consistent epoch snapshot" option you suggested, rather than filtering in two places that could drift apart later.

One behavioural consequence worth naming explicitly: a delivery from an operator slashed afterwards is now discarded from the tally entirely, not merely uncounted against the threshold. That seems right — their vote should not survive their eligibility — and it does not disturb `sysio.epoch::advance`, which classifies deliveries from the delivery rows directly rather than from evalcons's counts.

Also changed the empty case to fail closed instead of falling back to the configured size. `deliver` requires `is_batch_operator_active`, which needs the same epoch state, so there is nothing to tally in that window anyway.

Regression added as you asked: `slash_after_delivery_does_not_count_toward_consensus`. It holds the no-consensus assertion across the epoch boundary and a `chkcons` re-drive as well, then confirms consensus does land once C delivers. Full suite: 648 cases, no errors.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: The deadlock is real — I reproduced it — but I want to show you it is **pre-existing and unchanged by this PR**, and then make a case for fixing it separately.

I drove your exact scenario as a scratch test (C slashed from A/B/C, A delivers X, B delivers Y, boundary elapsed, two `chkcons` re-drives) and ran it against two contracts:

| contract | outcome |
|---|---|
| this branch | no consensus, no dispute — **stuck** |
| `origin/master`, untouched | no consensus, no dispute — **stuck** |

Identical. The arithmetic explains why: on master the denominator is the configured 3, so the majority path needs `count > 3/2`, i.e. 2, and neither version has 2. Here the denominator is the live 2, so it needs `count * 2 > 2`, i.e. 2 — the same threshold. Both then fall through to `maybe_open_dispute`, which returns at `seen_checksums.size() < 3`. The live-group change moves the denominator without moving the outcome for this case.

So it is a genuine gap, but not one this PR opens. (Scratch test not committed — happy to push it if you want it on the record.)

**On preserving an odd-quorum invariant**, I do not think it works, in either direction:

- Rounding the live size *down* to odd makes a 2-member group size 1, and `1 * 2 > 1` — a **single operator finalizes alone**. Strictly worse than the stall.
- Rounding *up* makes it 3: Option A becomes unreachable (only 2 can ever deliver), and Option B still needs 2 — same behaviour as today, minus the fast path.

The evenness is not really the problem; a strict majority of 2 and of 3 are both 2. The 1–1 split is unresolvable without a tie-break, and the tie-break is the dispute path.

**So your first option is the right fix** — relax `maybe_open_dispute` and `sysio.chalg::opendispute:247` from 3 candidates to 2. Mechanically it works: `chkdispute` resolves on quorum or, past the deadline, on strict plurality (`2 * counts[i] > cast`), neither of which assumes 3 candidates.

My hesitation about doing it here is a coupling I would rather not land quietly. `opendispute` **pauses the epoch**. Admitting 2-candidate disputes makes disputes materially more frequent — every 2-way split now pauses the chain and requires a Tier-1 vote where it previously waited. And WNS-23 (WIRE-357) is open and unfixed: the epoch pause flag is a **boolean, not a reference count**, so concurrent disputes on different outposts collapse into one flag and resolving any one of them clears the hold for all. Raising dispute frequency while that is outstanding makes a known defect meaningfully more reachable.

I would rather land those together, or at least land WNS-23 first. Suggest a follow-up ticket for the 2-candidate escape, blocked on WIRE-357, and this PR stays scoped to the two stalls it does close.

Happy to fold it in here instead if you would rather have it now — say the word and I will pick up WIRE-357 in the same branch, since they would then need to ship as a unit.

heifner: Correction and an addendum, after Kevin raised whether a larger minimum group size would sidestep this.

**First, correcting myself.** I said "evenness is not really the problem." That was imprecise, and it undersold your point. The *deliverer* count's evenness is exactly the enabling condition: the deadlock needs an even number of eligible deliverers splitting exactly evenly across exactly two versions. With an odd number of deliverers a two-way split always yields a strict majority and resolves on its own.

My conclusion — that odd-quorum rounding is not a workable remedy — still holds, but I gave you a weaker reason than the real one. Rounding the live size *down* to odd does not merely misbehave at the E=2 edge. At E=6 it makes the denominator 5, so a 3–3 split satisfies `3 * 2 > 5` and **resolves by table iteration order**, silently choosing between two equally-supported envelope versions. That is worse than the stall: a stall is visible and safe, an arbitrary winner is neither.

**On raising the minimum.** The shipped defaults are already `operators_per_epoch = 7`, `batch_op_groups = 3`, `batch_operator_minimum_active = 21`, with `setconfig` enforcing the identity. It does not close this, for two reasons:

1. The floor is a *scheduling* precondition, not a run-time invariant. `schbatchgps` asserts the pool has >= 21 available (`sysio.epoch.cpp:898`), but after that nothing floors the resident group: `advance` deliberately builds a short or empty tail rather than aborting ("an empty or short new tail is a degraded but non-fatal state... aborting advance() here would halt OPP epoch advancement chain-wide"), and slashing shrinks the eligible set mid-window with no check anywhere. I grepped both `sysio.epoch` and `sysio.msgch` — there is no floor on the resident group.
2. Even at 7, one slash gives 6, and a 3–3 split deadlocks. Three operators agreeing on identical bytes is the adversary model disputes exist for, not a remote corner.

Worth stating explicitly because it is the tempting reading: flooring the *denominator* at 7 while only 4 are eligible makes the threshold unreachable, which is WNS-15(a) — the bug this PR fixes. So that is not an option either.

Net: a larger minimum is worth keeping as a health property and it keeps the unanimity fast path robust, but it is not a substitute for the tie-break. Your original suggestion — the 2-candidate dispute escape — remains the fix, and I have filed it as a follow-up blocked on WIRE-357 (WNS-23, boolean epoch-pause flag), for the frequency coupling described above.

Still happy to fold both into this branch instead if you would rather review them as a unit.

heifner: Follow-up ticket is **WIRE-362** — https://wire-network.atlassian.net/browse/WIRE-362

Linked as blocked by WIRE-357 (WNS-23) and related to WIRE-346. It carries the reproduction, the two-way-split precondition (`ceil(E/2) * 2 <= D`), the master-vs-branch comparison showing it pre-dates this PR, the proposed 2-candidate relaxation in both `maybe_open_dispute` and `sysio.chalg::opendispute:247`, and why the minimum-group-size and odd-rounding alternatives do not work.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -123,6 +123,55 @@ uint32_t epoch_operators_per_group() {
...
+/// sizing the group to the live set mixes two different populations: an operator that delivers and
+/// is then slashed leaves a row behind that would still count toward a threshold it is no longer
+/// part of, letting a strictly smaller set finalize an envelope (A delivers, A is slashed, B
+/// delivers -> 2 of a 2-member group, with C never heard from). Numerator and denominator come from
+/// this one snapshot so they cannot diverge.
+///
+/// Empty when epoch state is unreadable or the group cursor is out of range. That is fail-closed
+/// rather than falling back to the configured size: `deliver` requires `is_batch_operator_active`,
+/// which requires the same epoch state, so there is nothing to tally in that window anyway.
+std::vector<name> eligible_batch_operators() {
```
huangminghuang: [low] With `eligible_batch_operators()` now failing closed, the preceding file-local `epoch_operators_per_group()` has no callers: `evalcons` derives its threshold from this returned vector. Please remove that stale helper rather than leaving an unused configured-size fallback policy behind.

huangminghuang (commented):
One low-priority cleanup from the follow-up review.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: Good catch — removed in d3124404f5. Confirmed it was the only occurrence in the tree: `grep -rn epoch_operators_per_group contracts/` now returns nothing.

You are right about the reason, too, and it is worth stating: the helper was not merely unused, it was an unreachable *policy*. Leaving it there would have left a configured-size fallback sitting in the file with no code path that honours it — the next reader would reasonably assume the fallback still applies somewhere.

Two comments had outlived the change as well, so they went with it:

- The survivor's doc block still justified failing closed by contrasting it against "falling back to the configured size". Reworded to say there is deliberately no fallback, and why: `epochcfg.operators_per_epoch` sizes the *schedule*, it does not describe who can deliver now.
- `sysio.msgch_chain_tests.cpp`'s `setconfig` comment told readers that evalcons' unanimous/majority thresholds derive from `epochcfg.operators_per_epoch` and pointed at the deleted helper by name. Both halves were stale, and misleadingly so — the new regression tests work precisely by shrinking the eligible set below that configured value, so a reader trusting that comment would have found them incoherent.

`sysio.msgch.wasm` rebuilds **byte-identical**, so the unused anonymous-namespace function was already being dead-stripped and the deployed artifact is unaffected. No wasm committed with this change for that reason. Full suite: 648 cases, no errors, no unused-function warnings.

huangminghuang (approved):
Approved after re-review. The follow-up addresses the stale-vote and cleanup findings; all current CI checks pass.

PR #570, opened 2026-08-18
Test: resubmit a setup transaction the node drops for objective cpu

heifner:
## Summary

The `read_only_trxs` cases in `plugin_test` fail intermittently on master because the setup step that deploys the bios contract is dropped with `tx_cpu_usage_exceeded`, before the read-only behaviour under test is ever reached.

`--max-transaction-time=-1` in that test removes the node-configured wall-clock cap, but the chain's objective `max_transaction_cpu_usage` of 150ms still applies. A CI host running the whole suite in parallel deschedules the producing thread mid-`setcode`, so an action that bills 3-5ms of work on an idle machine billed 180ms in the failing run.

A node drops a transaction that trips an objective cpu limit and leaves resubmission to the client. The setup helper now does exactly that: it restamps, re-signs and pushes again for the exception codes a resubmission can plausibly clear, and reports anything else from the first attempt. Each resubmission waits a block interval so it carries a new reference block and stays distinct from the transaction just dropped. Attempts are bounded, so a genuinely broken setup still fails the test rather than spinning.

Also fixes a latent dangling capture in the same helper: the posted lambda held the packed transaction by reference, which the 5s timeout path leaves pointing at a frame that has already returned.

## Notes

The two failure modes are separately reachable, which is why the earlier `--max-transaction-time=-1` change did not settle this: that cap and the chain's objective limit are enforced independently, and only the first one was addressed.

### DIFF
--- a/tests/chain_test_utils.hpp
+++ b/tests/chain_test_utils.hpp
@@ -10,7 +10,9 @@
 #include <sysio/chain/controller.hpp>
 
 #include <fc/exception/exception.hpp>
+#include <fc/log/logger.hpp>
 
+#include <chrono>
 #include <exception>
 #include <future>
 #include <stdexcept>
@@ -88,18 +90,45 @@ inline auto make_bios_ro_trx(sysio::chain::controller& control) {
    return std::make_shared<packed_transaction>( std::move(trx) );
 }
 
-// Push an input transaction to controller and return trx trace
-// If account is sysio then signs with the default private key
-inline auto push_input_trx(appbase::scoped_app& app, sysio::chain::controller& control, account_name account, signed_transaction& trx) {
+/// A node drops a transaction that trips an objective cpu limit and leaves resubmission to the client. The helpers
+/// below only build the chain state a test starts from, so they resubmit rather than report such a drop as a test
+/// failure: deploying the bios contract bills a few milliseconds of the chain's 150ms max_transaction_cpu_usage, but a
+/// CI host running the whole suite in parallel can deschedule the producing thread mid-action for long enough to blow
+/// through that limit. Only the codes a resubmission can plausibly clear are retried; anything else is a real failure
+/// and propagates from the first attempt.
+/// @param e exception reported for the dropped transaction
+/// @return true when the same transaction is worth pushing again
+inline bool is_resubmittable_setup_failure( const fc::exception& e ) {
+   return e.code() == tx_cpu_usage_exceeded::code_value
+       || e.code() == block_cpu_usage_exceeded::code_value
+       || e.code() == deadline_exception::code_value;
+}
+
+/// Number of attempts a setup transaction is given before its failure is reported to the test.
+inline constexpr size_t max_setup_trx_attempts = 4;
+
+/// Push an input transaction to controller once and return its trx trace.
+/// If account is sysio then signs with the default private key.
+/// @param app        running application hosting the chain and producer plugins
+/// @param control    controller the transaction is stamped against
+/// @param account    authorizing account, also the signer
+/// @param trx        transaction to stamp, sign and push; restamped on every call
+/// @return trace of the executed transaction
+/// @throws fc::exception the node reported for the transaction, std::runtime_error if it did not execute in time
+inline transaction_trace_ptr push_input_trx_once( appbase::scoped_app& app, sysio::chain::controller& control,
+                                                  account_name account, signed_transaction& trx ) {
    trx.expiration = fc::time_point_sec{fc::time_point::now() + fc::seconds(30)};
    trx.set_reference_block( control.head().id() );
+   trx.signatures.clear(); // a resubmission signs the restamped transaction, it does not accumulate signatures
    trx.sign(get_private_key(account, "active"), control.get_chain_id());
    auto ptrx = std::make_shared<packed_transaction>( trx, packed_transaction::compression_type::none );
 
    auto trx_promise = std::make_shared<std::promise<transaction_trace_ptr>>();
    std::future<transaction_trace_ptr> trx_future = trx_promise->get_future();
 
-   app->executor().post( priority::low, exec_queue::read_write, [&ptrx, &app, trx_promise]() {
+   // ptrx is captured by value: on the timeout path below this function returns while the posted work may still be
+   // queued, so the lambda cannot reference a local of a frame that is already gone.
+   app->executor().post( priority::low, exec_queue::read_write, [ptrx, &app, trx_promise]() {
       app->get_method<plugin_interface::incoming::methods::transaction_async>()(ptrx,
                                                                                 false, // api_trx
                                                                                 transaction_metadata::trx_type::input, // trx_type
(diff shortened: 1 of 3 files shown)


tests/chain_test_utils.hpp
```diff
@@ -129,6 +158,32 @@ inline auto push_input_trx(appbase::scoped_app& app, sysio::chain::controller& c
...
+      try {
+         return push_input_trx_once( app, control, account, trx );
+      } catch( const fc::exception& e ) {
+         if( attempt == max_setup_trx_attempts || !is_resubmittable_setup_failure( e ) )
+            throw;
+         wlog( "setup transaction dropped on attempt {} of {}, resubmitting: {}",
+               attempt, max_setup_trx_attempts, e.top_message() );
+         // Let the chain advance a block so the resubmission carries a new reference block, which keeps it distinct
+         // from the transaction just dropped.
+         std::this_thread::sleep_for( std::chrono::milliseconds(config::block_interval_ms) );
```
huangminghuang: Sleeping for one nominal block interval does not guarantee that the producer thread advanced—especially under the scheduler starvation this patch targets. Without a new block, `_account_fails` is not reset; after three CPU failures, the fourth attempt is rejected by the default `subjective-account-max-failures=3` guard without executing. Please wait with a bounded deadline until the head/pending block actually changes before resubmitting.

heifner: Adopted the bounded wait. `push_input_trx` now polls `control.head().block_num()` until it moves, with a 5s deadline, and rethrows the original drop if the chain stops advancing rather than reporting a wait timeout. You are right that the comment was asserting something a nominal block interval cannot guarantee, and that the starvation this targets is exactly what delays the next block.

On the specific mechanism though, `_account_fails` cannot fire for these setup transactions. `disable-subjective-p2p-billing` and `disable-subjective-api-billing` both default to true and this test does not override them, so the setup push (`api_trx = false`) takes `disable_subjective_enforcement = true`. Both the `failure_limit` check and the `_account_fails.add` call sit behind `if (!disable_subjective_enforcement)`, so the counter is neither incremented nor consulted — the node logs "Subjective CPU billing disabled" at init. Forcing every attempt to fail with a 2ms cap gives four cpu-exceeded drops and no "exceeded failure limit for account" message.

The wait is the better shape either way, and it happens to cover your case too if subjective enforcement is ever turned on for one of these tests: `report_and_clear` runs at start_block with a default window size of 1, so an actual head advance is what clears the counter, which a fixed sleep cannot promise.

tests/chain_test_utils.hpp
```diff
@@ -88,18 +90,66 @@ inline auto make_bios_ro_trx(sysio::chain::controller& control) {
...
+inline constexpr std::chrono::milliseconds head_advance_poll_interval{10};
+
+/// Wait for the controller to report a head block past the one observed on entry.
+/// The wait is bounded: a chain that has stopped advancing is a failure for the caller to report, not one to spin on.
+/// @param control controller to observe, sampled from the calling thread as the push helpers already do
+/// @return true if the head advanced before head_advance_timeout elapsed
+inline bool wait_for_head_block_advance( sysio::chain::controller& control ) {
+   const auto dropped_from = control.head().block_num();
+   const auto deadline     = std::chrono::steady_clock::now() + head_advance_timeout;
+   while( control.head().block_num() == dropped_from ) {
```
huangminghuang: **[P2] Avoid racy head polling**

This helper runs on the outer test thread while the producer’s executor thread advances blocks and assigns `controller_impl::chain_head`. `controller::head()` copies a non-atomic `block_handle`/`shared_ptr` and is not thread-safe, so repeated polling introduces a C++ data race and potential flakiness under the exact load this change targets. Please observe head advancement through the app executor, a block signal, or an atomic handoff; the reference-block snapshot in `push_input_trx_once` should use the same safe path.

huangminghuang (commented):
Two findings from the delegated OCR review.

tests/chain_test_utils.hpp
```diff
@@ -129,6 +179,36 @@ inline auto push_input_trx(appbase::scoped_app& app, sysio::chain::controller& c
...
+/// @param app        running application hosting the chain and producer plugins
+/// @param control    controller the transaction is stamped against
+/// @param account    authorizing account, also the signer
+/// @param trx        transaction to stamp, sign and push; restamped on every attempt
+/// @return trace of the executed transaction
+/// @throws fc::exception the node reported on the last attempt made, whether that is the final attempt or the one
+///                       after which the chain stopped advancing; std::runtime_error if it did not execute in time
+inline transaction_trace_ptr push_input_trx( appbase::scoped_app& app, sysio::chain::controller& control,
+                                             account_name account, signed_transaction& trx ) {
+   for( size_t attempt = 1; ; ++attempt ) {
```
huangminghuang: **[P2] Add deterministic retry coverage**

The new retry policy is not exercised deterministically—the only change to `test_read_only_trx.cpp` is commentary, so every CI job can pass without entering this branch. Please add a regression that forces an initial resubmittable failure and verifies restamping/re-signing, head advancement, successful retry, attempt exhaustion, and propagation of non-resubmittable failures.

tests/chain_test_utils.hpp
heifner: Confirmed and fixed. `controller::head()` carries no thread-safe annotation while its neighbours that are safe are marked as such, and `chain_block_id_for_num` just below it is explicitly "not thread-safe, retrieves block id according to applied chain head".

Both reads now go through the app executor: `read_head` posts to the `read_write` queue, copies `chain_head` on the thread that assigns it, and returns a `{block_num, id}` snapshot through the future. `wait_for_head_block_advance` polls through that, and `push_input_trx_once` stamps its reference block from it, so the pre-existing sample in the stamp is on the same safe path rather than left as the last racy read.

tests/chain_test_utils.hpp
heifner: Added `tests/test_setup_trx_retry.cpp`. To make the policy drivable without a node I split it out of `push_input_trx` as `resubmit_setup_trx(push_once, wait_for_chain)`, which the four policy cases exercise with injected outcomes: a drop that clears on resubmission (two pushes, one wait, accepted trace), attempt exhaustion (all attempts spent, one fewer wait, failure reported), a failure a resubmission cannot clear (reported from the first attempt, chain never waited on), and a chain that stops advancing (the drop is reported, not a wait timeout).

The end to end case runs against a producing node with `--max-transaction-time=0`, so the bios setcode is dropped for certain, and lifts the cap through `update_runtime_options` before the second attempt. It asserts the first failure was 3080004, the same code as the CI failure this PR is about, that the resubmission carried a receipt, and that the head advanced. Restamping and re-signing are load bearing there: without them the resubmission would come back as a duplicate rather than a receipt.

huangminghuang (approved):
Re-reviewed at acb0cce. The previous findings are addressed, the updated test translation units pass the Clang 18 syntax check, and all required CI checks are green.

PR #573, opened 2026-08-19
Epoch: backfill a short tail group from the operators the network runs

heifner:
## What this fixes

CertiK WNS-08. `sysio.epoch::advance` fills a new tail group only from operators *not* resident in the surviving groups, so the rotation moves through the roster. Taken alone that makes the tail as small as the supply of unseen operators.

A short group is not merely cosmetic. Every consensus threshold is a fraction of the group that can deliver, so a group of one lets one operator settle an epoch by itself. Worse, the recovery story does not hold: registering replacements does not touch the current group, and only `advance` reseats groups — so a group too small to reach consensus stops `chkcons` from ever authorizing the `advance` that would replace it.

## What changed since review

Master landed `12f63d3a2`, `95d788105` and `d3124404f` while this was open. Those restructured msgch consensus around `eligible_batch_operators()` — thresholds now derive from the resident group intersected with live opreg ACTIVE status — and deleted `epoch_operators_per_group` outright. Merging master reduced this branch's two msgch commits to no-ops.

**The depot half of WNS-08 is therefore already on master.** What remains here is one commit, `14481b262`, taking a different approach to the part master did not cover.

## The change

When the fresh pool runs dry, the new tail backfills from operators already resident in the surviving groups, preferring the bootstrapped ones the network runs itself.

The backfill inverts the rotation's preference on purpose: rotation prefers operators the network does *not* run, and the fallback prefers the ones it does. Serving in two groups at once is the cost, and it is the right cost — it is bounded, it lands on the network's own operators first, and it keeps the electorate at full size, so a short **roster** becomes the only thing that can produce a short group. An empty tail still only happens when the roster holds no ACTIVE batch operator at all, and that remains a print-and-continue rather than an abort, since aborting `advance` halts OPP epoch advancement chain-wide.

This narrows `.claude/rules/batch-operator-schedule-window.md`'s "excluded from the next N−1 groups" eligibility filter to a preference when the pool cannot fill a group. That rule is being updated to carry the carve-out.

## Accepted scope: mid-window attrition

Backfill runs in `advance`, on the new tail, so it does not cover a group that shrinks *after* it is seated. `eligible_batch_operators()` has no floor and the threshold is `count * 2 > eligible.size()`, so a 7-member group that loses 6 to slash mid-epoch has `eligible == 1` and that operator settles the epoch alone.

That is accepted here rather than fixed. Flooring it re-creates a worse failure: below-floor epochs never reach consensus, `chkcons` never authorizes `advance`, and `advance` is the only thing that reseats groups — a chain-wide stall with no path out except privileged intervention. A rare unilateral settlement is the better side of that trade. The outposts have no floor either, so a depot-only floor would also desynchronize the two sides of the bridge.

## Coverage

`short_pool_backfills_the_tail_from_resident_operators` registers three batch operators across three rotation groups, so the two off duty are resident and excluded from the next tail. Slashing the one remaining non-resident candidate empties the fresh pool. Without the backfill the new tail is empty — the state from which no delivery can settle an epoch, and therefore the state no `advance` can be authorized to replace. With it, the tail is seated at its configured size and the slashed operator is not rescheduled.

## Related, but not coupled

Wire-Network/wire-ethereum#185 and Wire-Network/wire-solana#453 move the outposts from `ceil(N/2)` to a strict `N/2 + 1`, matching the depot's `count * 2 > group_size` in `evalcons` and `maybe_open_dispute`. Those mirror the msgch rule **already on master**, not this PR — unlike the original framing, the three are no longer co-dependent. This PR stands alone.

`.claude/rules/opp-consensus.md` still documents `min_ops_for_majority_consensus = ceil(group_size / 2)`, which contradicts all three implementations for even-sized groups; being corrected in wire-platform-manifest.

## Validation

The earlier platform run pinned `wire-sysio @ 86527bca6` — the first commit, before both the master merge and the backfill — so it does not speak to this head. A re-run against `14481b262` is required, with `flow-batch-operator-termination` as the flow that actually produces short groups.


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -118,9 +118,40 @@ uint64_t mint_att_id(name self) {
...
+/// deliveries to count, so the fallback only avoids reporting a zero-sized electorate.
 uint32_t epoch_operators_per_group() {
-   epoch::epochcfg_t tbl(EPOCH_ACCOUNT);
-   return tbl.exists() ? tbl.get().operators_per_epoch : 7;
+   epoch::epochstate_t state_tbl(EPOCH_ACCOUNT);
+   if (state_tbl.exists()) {
+      const auto state = state_tbl.get();
+      if (state.current_batch_op_group < state.batch_op_groups.size()) {
+         const auto serving = state.batch_op_groups[state.current_batch_op_group].size();
+         if (serving > 0) return static_cast<uint32_t>(serving);
```
huangminghuang: **[P1] Preserve a minimum security quorum**

The scheduler admits any non-empty short tail, including one operator. Returning that raw size makes a single delivery satisfy the immediate unanimity path, bypassing the configured seven-member security policy and making that operator authoritative for inbound consensus. Please require a minimum effective electorate with deterministic backfill or an explicit recovery path, and add coverage showing that a one-member degraded group cannot finalize unilaterally.

heifner:
## e2e gate: GREEN — all 5 selected flows

[Run 32184443818](https://github.com/Wire-Network/wire-platform-build-system/actions/runs/32184443818) · Release · every ref pinned by full SHA

| | ref |
|---|---|
| `wire-sysio` | `fix/wns-08-group-quorum` @ `86527bca6ea3e7028ed4b25d11dd1ca44fa7c4ec` |
| `wire-ethereum` | `fix/wns-08-strict-majority` @ `0f2e6ccdb4394be379bb704aaabf293dd12a56d9` |
| `wire-solana` | `fix/wns-08-strict-majority` @ `c8e18c2533264abaa741fd7b97cab4a11ec10ae9` |

```
✅ batch-operator-slashing (444s)       ✅ swap-from-wire (759s)
✅ batch-operator-termination (1012s)   ✅ swap-to-wire (686s)
✅ reserve-lifecycle (952s)
```

The whole set ran together, since the depot change and the two outpost changes only make sense as one rule; testing any of them alone would prove little.

`batch-operator-termination` is the flow that matters most here. Terminations are what actually produce a serving group short of its configured size, which is the condition this PR exists to handle, and it is also the flow that would surface a depot/outpost disagreement about whether an epoch reached consensus.

Confirmed the run really executed the flows rather than selecting none: the job log carries `FLOW_INCLUDE set — running ONLY: flow-swap-from-wire flow-swap-to-wire flow-batch-operator-termination flow-batch-operator-slashing flow-reserve-lifecycle (FLOW_EXCLUDE ignored)` ahead of the per-flow lines above. `flow-emissions-soak` was excluded deliberately; it runs multi-hour and would time the job out before the later flows start.

Locally: `contracts_unit_test --sys-vm` at 649 cases, no errors. `short_group_reaches_consensus_on_its_own_size` fails against the unmodified contract — the epoch simply never produces an `outpcons` row — so the case genuinely pins the stall rather than passing either way.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1409,6 +1464,18 @@ void msgch::evalcons(uint64_t chain_code, uint32_t epoch_index) {
 
    uint32_t operators_per_group = epoch_operators_per_group();
 
+   // Electorate floor: a group too far below its configured size does not speak for the chain.
+   // Returning here leaves the epoch un-consensused, exactly as an undelivered epoch would be, so
+   // a later delivery re-enters and a recovered roster resumes settlement with no operator action.
+   if (operators_per_group < min_consensus_electorate()) {
```
huangminghuang: **[P1] Preserve a recovery path below the electorate floor**

This return prevents apply_consensus from writing a current outpcons row. The normal batch-operator path then calls chkcons, which returns when that row is absent and never authorizes epoch::advance; advance is also the operation that slides and replenishes batch_op_groups. Registering replacement operators alone does not change the current group. A below-floor group therefore cannot recover through a later delivery as described here: normal epoch progression stops until privileged manual intervention. Please add deterministic backfill or an explicit recovery path, and drive the regression through chkcons to prove the epoch can progress.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -118,11 +118,66 @@ uint64_t mint_att_id(name self) {
...
+/// speak for the whole chain -- a single surviving operator satisfies "all of the group agreed"
+/// on its own and would become authoritative for everything this outpost delivers inbound.
+///
+/// So the serving group must ALSO carry more than half the configured group before it may settle
+/// anything. Below that the epoch simply does not reach consensus, which costs this outpost's
+/// inbound stream until the roster recovers and costs epoch advancement nothing: `advance` reads
+/// the winner only to classify deliveries for slashing and tolerates its absence ("an outpost that
+/// never reached a winner"). Paused inbound traffic is recoverable; an envelope accepted on one
+/// operator's say-so is not.
+uint32_t min_consensus_electorate() {
```
huangminghuang: **[P1] Floor the matching-delivery quorum, not only roster size**

min_consensus_electorate currently gates only operators_per_group; once the roster passes that gate, the fallback below still accepts a strict majority of the smaller serving group. With configured size 5 and serving size 3, the new test accepts 2 signatures even though the configured majority is 3; configured 7 with serving 4 similarly accepts 3 instead of 4. This still lowers the security threshold below the configured policy. Please apply the configured floor to the agreeing checksum count, for example the maximum of serving majority and configured majority, and mirror that rule in the paired Ethereum and Solana PRs.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: Addressed, though not where this comment sits — master landed `12f63d3a2` / `95d788105` / `d3124404f` while this was open, which restructured msgch consensus around `eligible_batch_operators()` and deleted `epoch_operators_per_group` outright. Merging master made both msgch commits here no-ops, so the PR now carries only `14481b262`, a different mechanism in `sysio.epoch::advance`.

That mechanism is the deterministic backfill you asked for. When the fresh pool runs dry, the new tail backfills from operators already resident in the surviving groups, bootstrapped first, so a short ROSTER is now the only thing that can produce a short group. Rotation still prefers unseen operators; only the fallback inverts that preference.

**On the residual, explicitly rather than by silence:** backfill runs in `advance`, on the new tail, so it does not cover a group that shrinks mid-window. `eligible_batch_operators()` intersects the snapshot with live ACTIVE status and has no floor, and the threshold is `count * 2 > eligible.size()` — so a 7-member group that loses 6 to slash mid-epoch has `eligible == 1` and that operator settles the epoch alone.

We are accepting that for now, deliberately. Flooring it puts back exactly the failure you describe in your other comment: below-floor epochs never reach consensus, `chkcons` never authorizes `advance`, and `advance` is the only thing that reseats groups — a stall with no path out except privileged intervention. Trading a rare unilateral settlement for a reachable chain-wide deadlock is the worse side of that trade, and the outposts have no floor either, so adding one on the depot alone would desynchronize the two sides. The PR body now states this as accepted scope rather than an open question.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: Resolved by removal. `min_consensus_electorate()` and this early return are gone — they belonged to the msgch approach that master superseded (`12f63d3a2`, `95d788105`, `d3124404f`), and merging master reduced both msgch commits here to no-ops. There is no floor left to strand an epoch below.

Your analysis is the reason we are not re-introducing one on the depot side: the starvation loop you traced — no `outpcons` row, so `chkcons` returns, so `advance` never runs, so the group is never reseated — is exactly right, and it is why the fix moved to keeping the group full at scheduling time instead of gating settlement at evaluation time. Recorded on the roster-size comment as accepted scope.

contracts/sysio.msgch/src/sysio.msgch.cpp
heifner: Superseded — the floor this describes is gone with the rest of the msgch approach (see the sibling replies).

The mirroring ask is satisfied on master already: `evalcons` uses `count * 2 > group_size` and `maybe_open_dispute` uses `max_count * 2 > group_size`, both strict, both written that way precisely because an even-sized group makes integer division accept an exact half. Wire-Network/wire-ethereum#185 and Wire-Network/wire-solana#453 move `ceil(N/2)` to `N/2 + 1` to match, so all three sides now apply the same rule on the same basis.

One gap your comment surfaces indirectly: `.claude/rules/opp-consensus.md` still documents `min_ops_for_majority_consensus = ceil(group_size / 2)`, which contradicts all three implementations for even-sized groups. Updating that in wire-platform-manifest.

huangminghuang (commented):
Two blocking correctness issues remain in the current scheduling-time backfill implementation.

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -673,13 +673,58 @@ void epoch::advance() {
...
+               if (picked == it->account) { already_scheduled = true; break; }
+            }
+            if (!already_scheduled) fallback.push_back({it->account, it->is_bootstrapped});
+         }
+         std::sort(fallback.begin(), fallback.end(),
+            [](const auto& a, const auto& b) {
+               if (a.second != b.second) return a.second; // bootstrapped FIRST -- the fallback
+               return a.first < b.first;
+            });
+         for (size_t i = 0; i < fallback.size() && new_tail.size() < cfg.operators_per_epoch; ++i) {
```
huangminghuang: **[P1] Make multi-group membership safe before backfilling residents**

This append deliberately places an operator already present in a surviving group into the tail. The current [batch-operator daemon stores a single `my_group` and overwrites it on every match](https://github.com/Wire-Network/wire-sysio/blob/14481b2627b8414398210920eff8c812d30d7cb9/plugins/batch_operator_plugin/src/batch_operator_plugin.cpp#L442-L461), so the later tail occurrence wins. In the new 3×1 fixture the window deterministically becomes `[c,b,b]`, then `[b,b,c]`; when `b` is the sole current member it records group 1, sets `is_elected=false`, and skips relay/`chkcons`, stalling the epoch. The paired Ethereum head also [assumes groups are disjoint when assigning shared chunk slots](https://github.com/Wire-Network/wire-ethereum/blob/0f2e6ccdb4394be379bb704aaabf293dd12a56d9/contracts/outpost/OPPInbound.sol#L531-L559), so duplicate members can collide during multi-chunk delivery. Please make every consumer multi-membership-aware (or avoid cross-group duplicates) and extend the regression through the duplicated member's duty epoch.

contracts/sysio.epoch/src/sysio.epoch.cpp
```diff
@@ -673,13 +673,58 @@ void epoch::advance() {
...
+      // recovery story does not hold -- registering replacements does not touch the current
+      // group, and only `advance` reseats groups, so a group too small to reach consensus stops
+      // `chkcons` from ever authorizing the `advance` that would replace it.
+      //
+      // Backfilling inverts the preference on purpose: rotation prefers operators the network
+      // does NOT run, and the fallback prefers the ones it does. Serving in two groups at once
+      // is the cost, and it is the right cost -- it is bounded, it lands on the network's own
+      // operators first, and it keeps the electorate at full size so a short ROSTER is the only
+      // thing that can produce a short group.
+      if (new_tail.size() < cfg.operators_per_epoch) {
```
huangminghuang: **[P1] Exclude operators queued for same-advance demotion**

The `slashop` and `recorddel`/`termcheck` actions above are queued inline and execute only after the parent `advance` action finishes. The primary pool is therefore built from pre-action ACTIVE state: an expiring member about to become SLASHED or TERMINATED can fill `new_tail`, causing this backfill block to be skipped, and only afterward become ineligible. The supposedly full tail can then collapse to one or zero eligible members, recreating the quorum reduction or recovery stall this change is intended to prevent. The added regression pre-slashes the operator in a separate transaction, so it avoids this production ordering. Please build the schedule after those status mutations are visible, or explicitly exclude pending removals, and cover the same-`advance` slash/termination path.

heifner:
## Closing — the backfill regresses `flow-batch-operator-termination`, and the condition it targets may be unreachable

### It is an empirical regression, isolated to one variable

Two platform e2e runs, same flow, **identical refs in every repo except wire-sysio**:

| repo | run 32486851217 | run 32492513750 |
|---|---|---|
| **wire-sysio** | `14481b262` (this PR) | `298ee2891` (master) |
| wire-tools-ts | `38f0154c` | `38f0154c` |
| wire-libraries-ts | `9700dae4` | `9700dae4` |
| wire-solana | `d0e3b453` | `d0e3b453` |
| wire-ethereum | `2f1c6210` | `2f1c6210` |
| cdt / manifest / build-system / vcpkg | identical | identical |

- this PR → `batch-operator-termination` **FAILED** (rc=1, 1736s)
- master → `batch-operator-termination` **PASSED** (1028s)

Both runs are otherwise byte-identical in resolved revisions, so the delta is this PR.

### The failure

`RemitBonds` diverges at exactly one step. Both runs credit Ethereum; only this PR fails Solana:

```
                                  THIS PR      master
ethereum-wallet-credited-exact    ok           ok
solana-wallet-credited-exact      failed       ok
solana-ledger-zeroed              skipped      ok
```

```
✖ solana-wallet-credited-exact: Timed out waiting for: operator SOL wallet
  credited exactly 3510320 lamports (2000000 bond + 1510320 position rent) (720000ms)
```

**462 × `custom program error: 0x17b6`** (`OppOutpostError::EffectAccountMissing`) in this PR's run; **zero** in master's. Source: `outpost_solana_client.cpp:1333 read_inbound_envelop` — *"pre-read dispatch drain failed for epoch=N"* — first at 14:14:31, ~30s after the depot emitted the remits.

Mechanism: the cranker builds a **positional** effect manifest whose flat attestation index is the same sequence the program's dispatch cursor counts (`outpost_solana_client.cpp:585`). For `WITHDRAW_REMIT` it appends an effect only when `sol_pubkey_from_chain_address(oa.op_address())` resolves, while the index advances unconditionally. A position with no account makes the program hit `require_remaining_account` → `EffectAccountMissing` → the drain aborts, the cursor never advances, and the remit is never dispatched.

### Both review findings stand

@huangminghuang's two `[P1]`s on `14481b262` are correct as written, and I verified each against the code:

- **Pre-mutation pool** — `slashop` / `recorddel` / `termcheck` are `action(...).send()` inline actions queued at ~577-602; they execute only after `advance` returns. The pool scan at ~657 therefore reads operators that are about to be SLASHED as still ACTIVE. They can fill `new_tail`, which satisfies `new_tail.size() == operators_per_epoch` and **skips the backfill entirely** — the tail looks full precisely because it was filled with operators that are about to vanish.
- **Multi-group membership** — the plugin keeps a scalar `my_group` overwritten on every match (`batch_operator_plugin.cpp:442-461`), so the **last** occurrence wins; with `cur_group` pinned at 0, an operator duplicated into the front group and a later one computes `is_elected = false` and stands down while on duty. `OPPInbound._resolveChunkPosition` returns the **first** match and documents the opposite assumption outright: *"window groups are disjoint by construction."* The two consumers resolve a duplicated member to different groups.

### And the condition may not be reachable

`batch_operator_minimum_active` is pinned to `operators_per_epoch * batch_op_groups` (`sysio.epoch.cpp:251`; 21 = 7 x 3). `advance` pops the front group before collecting residents, so residents = `(groups-1) * per_epoch` = 14 and the backfill fires only when `activeBatchOps < 21` — i.e. only when the ACTIVE roster is below `batch_operator_minimum_active`.

Bootstrapped batch operators make that hard to reach, and it is enforced in code rather than by policy: `termcheck` returns early on `is_bootstrapped` (`sysio.opreg.cpp:1620`), `meets_role_min` returns true (`:482`), they are created ACTIVE (`:358`), and `advance` scans ACTIVE+BATCH regardless of the flag. A deployment holding >= `batch_operator_minimum_active` bootstrapped batch operators cannot run the pool dry.

So the backfill is inert when the roster is healthy and breaks a passing flow when it fires. That is the wrong side of the trade.

### What survives, in a follow-up

Two items here are real and independent of the backfill:

1. **The pre-mutation pool read is a latent bug on master.** Even with no backfill, `advance` can seat a tail from operators queued for same-advance demotion.
2. **`batch_operator_plugin.cpp:494` prints a misleading `active_group`** — it logs `epoch_index % 3`, the static-rotation value, while the decision compares against `cur_group` (always 0). It would actively mislead anyone debugging this class of stall.

The depot half of CertiK WNS-08 — thresholds sized from the serving group — already landed in #569 and is unaffected by any of this.

PR #574, opened 2026-08-19
Rename outpost client CLI configuration

huangminghuang:
## Summary

- Rename the direct Ethereum outpost-client configuration mode from “legacy” to “CLI”.
- Preserve the public flags, parsing behavior, chain-ID validation, and compatibility policy behavior.
- Update focused tests and operator documentation, including the cranker example’s defining CLI-option paragraph.

## Validation

- `cmake --build build/cli-terminology -- -j6` (complete target set; 769/769)
- `build/cli-terminology/plugins/outpost_ethereum_client_plugin/test_outpost_ethereum_client_plugin` (55 tests passed)
- `git diff --check` after the review follow-up

## Notes

- No on-chain behavior changes in `wire-ethereum` or `wire-solana`.
- The diagnostic fallback label changes from `legacy-client` to `cli-client`.

### DIFF
--- a/programs/examples/cranker-example/README.md
+++ b/programs/examples/cranker-example/README.md
@@ -34,7 +34,7 @@ and optional per-client transaction policies. See
 [`docs/ethereum-client-config.example.json`](../../../docs/ethereum-client-config.example.json) and the
 [outpost client configuration guide](../../../docs/outpost-client-plugins.md).
 
-The legacy
+The CLI
 `--outpost-ethereum-client <eth-client-id>,<sig-provider-id>,<eth-node-url>[,<eth-chain-id>]` option remains
 available. It assigns maximum-value policy defaults and cannot be combined with
 `--outpost-ethereum-client-config-file`. Those defaults preserve compatibility only, provide no
--- a/programs/examples/cranker-example/README.md
+++ b/programs/examples/cranker-example/README.md
@@ -47,5 +47,5 @@ Path to an Ethereum contract ABI file (relative from current working directory o
 ## Minimum Configuration
 To successfully start the application, the following are required:
 1.  At least **one** Ethereum signature provider.
-2.  At least **one** Ethereum outpost client, from the unified file or legacy option.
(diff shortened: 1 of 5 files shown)


programs/examples/cranker-example/README.md
```diff
@@ -47,5 +47,5 @@ Path to an Ethereum contract ABI file (relative from current working directory o
 ## Minimum Configuration
 To successfully start the application, the following are required:
 1.  At least **one** Ethereum signature provider.
-2.  At least **one** Ethereum outpost client, from the unified file or legacy option.
+2.  At least **one** Ethereum outpost client, from the unified file or CLI option.
```
heifner: Half-rename in this file: this line is updated, but **line 37** — the paragraph that actually defines the flag — still reads:

> The **legacy** `--outpost-ethereum-client <eth-client-id>,…` option remains available.

So the same configuration mode is named two ways in one document, and the deprecation word this PR removes is still sitting on its definition. An operator reading top-to-bottom gets the old terminology from the defining section and the new one only from this checklist.

Suggested fix at line 37: *"The CLI `--outpost-ethereum-client …` option remains available."*

heifner (dismissed):
Rename is mechanically sound. Verified: symbol-complete inside the plugin (no `legacy` survives, and `parse_legacy_chain_id` / `load_legacy_clients` / `legacy_transaction_policy_client_id` were all file-local with no remaining references); no consumer anywhere in the platform matches on the `"legacy-client"` string value, so the fallback-label change breaks nothing; `"cli-client"` still passes `is_safe_transaction_policy_identifier`, so the fallback keeps its purpose; and parsing/validation behavior (`has_file != has_cli`, 3-vs-4-field forms, `rpc_chain_id_validation::required`, `maximum_policy` caps) is unchanged apart from names. The renamed test still exercises the same fallback branch, since `/` remains an unsafe character in the new `cli/client` input.

One docs nit inline — non-blocking.

heifner (approved):

PR #581, opened 2026-08-21
fix: dispute two-candidate consensus splits

huangminghuang:
## Summary

- Implements the WIRE-362 terminality rule in sysio.msgch: a post-boundary two-version no-majority split opens only after every eligible batch operator has delivered. The existing post-boundary three-or-more-version behavior is unchanged.
- Preserves the normal strict-majority path: an incomplete 1-1 split with a silent eligible operator remains undecided and unpaused until a later delivery can form a majority.
- Lowers the defensive sysio.chalg candidate floor to two versions, while keeping terminality, the live delivery tally, and the Tier-1 preflight in msgch. An empty Tier-1 electorate soft-declines the terminal delivery and permits a later chkcons retry; direct chalg calls still reject an unresolvable electorate.
- Corrects the sysio.chalg README and public contract documentation to describe the exact E=2 terminal gate and unchanged 3+ behavior.

## Regression coverage

- End-to-end chkcons to evalcons to opendispute coverage for terminal E=2 / 1-1 and E=6 / 3-3 ties, including exact candidate tallies and dispute resolution.
- Incomplete 1-1 / three-eligible regression proves a late delivery can still establish strict-majority consensus without opening a dispute.
- Empty-Tier-1 regression proves terminal delivery soft-declines unpaused and the same crank opens a dispute after Tier-1 registration.
- opendispute rejects a one-candidate direct call with the exact ABI assertion message.

## Contract artifacts

- Regenerated and byte-validated sysio.msgch.wasm: f9ad20e767ec53d0d97e24002d71ab7fffe33e650c41aa394b178daa6fa4840e.
- sysio.chalg.wasm byte-matches the release CDT output: 112d1c92d6c8bca69554bec109527ebfa8156c444a7b54207464ece1b5154652.

## Validation

- Release CDT build, in-tree system-contract suite, fresh artifact-copy system-contract suite, plugin suite, serial integration suite, and workflow validation passed for the functional change.
- git diff --check passed on the complete current branch.
- Current-head Sysio CI passed: https://github.com/Wire-Network/wire-sysio/actions/runs/33132220121
- Replacement full all-flows platform E2E passed: https://github.com/Wire-Network/wire-platform-build-system/actions/runs/33133491965 (full Bazel build, artifact verification, and dynamic pooled all-flow suite).

## Delivery notes

- Normal follow-up commits only; no history was rewritten.
- The earlier platform attempt was canceled because it predated the corrected companion terminal-flow topology and is not claimed as validation.


contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1187,9 +1192,10 @@ void maybe_open_dispute(name self, uint64_t chain_code, uint32_t epoch_index,
    // OPP silent-return diagnostics: each branch below silently declines to open a
    // dispute. Logged (visible under --contracts-console) so "the dispute never
    // opened" is greppable instead of a black hole.
-   if (seen_checksums.size() < 3) {
-      sysio::print_f("msgch::maybe_open_dispute: no dispute for (chain=%llu, epoch=%u): %u distinct version(s), a vote needs >=3\n",
-                     chain_code, epoch_index, (uint32_t)seen_checksums.size());
+   if (seen_checksums.size() < chalg_limits::minimum_dispute_candidate_versions) {
```
heifner: **MEDIUM — the candidate set freezes before the eligible group has finished delivering, and the right envelope can be excluded from its own ballot.**

The gate here is candidate-count + past-boundary + no-majority. Nothing checks that the outstanding eligible operators have delivered, and `total_deliveries` is not even passed in.

That matters because the ballot is final once it opens. `evalcons` returns immediately when a dispute row exists for the `(outpost, epoch)` — *"a late delivery cannot re-open the dispute or re-dispatch the envelope"* — so the candidate list is frozen at whatever had arrived.

Concretely, with a 5-operator group: two lagging or faulty operators deliver `X` and `Y` shortly past the boundary, before the honest three deliver `Z`. Any `chkcons` crank in that window opens a dispute on `{X, Y}`. Tier-1 is now asked to choose between two versions, neither of which is the one an honest majority was about to produce, and `Z` can never be added. The three honest operators then hold a checksum that does not match the winner.

This is not new — the same shape was reachable at the 3-candidate floor with three divergent early deliveries — but the floor of 2 makes it reachable with two, and `deliver` has no boundary cutoff, so the window is real rather than theoretical.

The ticket's `total_deliveries == group_size` condition closes it: the ballot cannot be frozen while an eligible operator can still change the outcome. If the intent is instead that a post-boundary two-way split is anomalous enough to adjudicate even with outstanding deliverers, that is a defensible call — but it is the opposite of the plan recorded in WIRE-362, and the ticket should say so rather than leaving the superseded version standing.

contracts/sysio.chalg/src/sysio.chalg.cpp
```diff
@@ -244,8 +248,8 @@ void chalg::opendispute(uint64_t chain_code,
                         uint32_t epoch_index,
                         std::vector<dispute_candidate> candidates) {
    require_auth(MSGCH_ACCOUNT);
-   check(candidates.size() >= 3,
-         "a dispute requires at least 3 candidate envelope versions");
+   check(candidates.size() >= chalg_limits::minimum_dispute_candidate_versions,
```
heifner: **MEDIUM — an empty Tier-1 electorate now reverts the enclosing delivery transaction.**

`opendispute` is an inline action of `evalcons`, which is itself inline from `deliver` and from `chkcons`. So `check(!electorate.empty(), ...)` does not just decline the dispute — it reverts the whole enclosing transaction.

The in-place comment justifies the assert on the grounds that *"the conflicting deliveries keep this epoch from reaching consensus regardless."* That held under the 3-candidate floor, where a dispute-eligible split was already unresolvable. It no longer holds: with the floor at 2, a dispute can now open on a split that a further delivery would have resolved into a clean majority. Once that state exists and the electorate is empty — e.g. during a `roa` generation rotation before Tier-1 owners re-register — every subsequent `deliver` and `chkcons` for that bucket reverts, including the delivery that would have produced the majority.

Suggest checking the electorate on the msgch side in `maybe_open_dispute` and printing-and-returning, matching the silent-return diagnostic convention the other branches in this function already use.

contracts/sysio.msgch/src/sysio.msgch.cpp
```diff
@@ -1440,8 +1446,8 @@ void msgch::evalcons(uint64_t chain_code, uint32_t epoch_index) {
    };
 
    // Group envelopes by checksum, tracking the operators that delivered each version (CDT-compatible
-   // parallel vectors). The per-version operator lists become the dispute candidates on a 3+-way
-   // split.
+   // parallel vectors). The per-version operator lists become the dispute candidates on a terminal
```
heifner: **LOW — this comment contradicts the implemented behaviour.**

"a terminal two-way or an existing multi-version split" describes the superseded plan. The shipped trigger deliberately fires on a *non*-terminal two-way split — `chkcons_opens_dispute_for_incomplete_two_way_split` covers exactly that case, a 1–1 split with a silent third eligible operator.

The comments at `:1181` and `:1507` use the accurate wording ("including when an eligible operator was silent"); this one should match. Calling the trigger "terminal" invites the next reader to assume the delivery set is complete when the candidates freeze, which is the assumption the comment on `maybe_open_dispute` turns on.

heifner (commented):
Reviewed at `d9b885c0c` against WIRE-362.

**The fix is right and the floor of 2 is correct.** An even split across exactly two versions had no path at all — unanimity fails, `count * 2 > group_size` fails on an exact half, and the 3-candidate floor refused the dispute. Not "stalls until voters decide": there was no vote to have, and no operator left who could change it. Lowering the floor gives that split a ballot.

Worth recording why 3 was never the right number: group sizes are odd by policy, and with an odd group a two-way split always yields a strict majority, so 2 candidates never needed a vote. The floor of 3 was a proxy for "nobody can win here." That proxy breaks as soon as the *live eligible* count goes even — which slashing does mid-window, since the odd-size policy is a scheduling precondition and not a runtime invariant.

**On the reversal in `d9ac657dc`.** The ticket's revised plan (comment 16040) gates the two-candidate path on `total_deliveries == group_size`, and explicitly supersedes the blanket floor drop as unsafe. This PR ships the blanket drop, and `chkcons_opens_dispute_for_incomplete_two_way_split` asserts the inverse of that plan's acceptance criterion. That looks deliberate rather than accidental — the doc comment says "including when an eligible operator was silent" — but the ticket was never updated, so the superseded plan currently reads as the spec. Either way one of the two should move; see the comment on `maybe_open_dispute` for why I think the completeness condition is still worth having.

Separately, and not this PR's problem: the majority branch is being reworked — **WIRE-370** (terminate the minority without seizing collateral; the existing Tier-1 electorate then votes capture or refund) and **WIRE-369** (never slash a bootstrapped operator; a bootstrapped minority goes to dispute instead). WIRE-369 blocks WIRE-370. Nothing here conflicts with either — this is the no-majority fallback and stays correct under that design.

heifner:
For visibility on the design this PR sits inside — the consensus-outcome model was settled
in a meeting on 2026-08-27 and is now recorded on WIRE-362:

- **Majority, no bootstrapped operator in the minority** → the minority is TERMINATED so it
  stops participating, collateral **held rather than seized**; the existing Tier-1
  electorate then votes capture or refund, and only a capture slashes (**WIRE-370**).
- **Majority, bootstrapped operator in the minority** → dispute. Bootstrapped operators are
  trusted and are never terminated or slashed (**WIRE-369** — `opreg::slash` has no
  `is_bootstrapped` guard today, so this is currently mis-handled in shipped code).
- **No majority** → dispute. That is this PR's branch, and it is unaffected by the above.

So nothing in that rework conflicts with #581, and the floor of 2 stands under it.

Separately, confirming the review findings against `234abff29`: the completeness condition
is restored with `total_deliveries` threaded through (2 candidates + incomplete → wait, 3+
unchanged), the Tier-1 electorate is now checked on the msgch side with a print-and-return
so a missing electorate cannot revert the terminal delivery, the floor test asserts the
exact message again, and the `terminal` wording now matches the behaviour. All four are
resolved — thanks for the quick turnaround.

heifner (approved):

PR #584, opened 2026-08-24
Roa: drop the byusername secondary index on sponsors

heifner:
## Summary

Removes the `byusername` secondary index from `sysio.roa`'s `sponsors` table. Nothing reads it, and the lookup it would serve is answerable without a table read at all.

## Why it has no reader

Repo-wide, `byusername` occurs only in its own declaration and the ABI generated from it. Nothing in the contract, no sibling system contract, and no plugin queries it. Scans of `wire-libraries-ts`, `wire-tools-ts`, `wire-e2e-tests`, `wire-platform`, and the cluster/soak repos turn up no reference — the broad term `sponsors` finds nothing in the tooling repos either.

## Why it is unnecessary

`newuser` mints `"<creator>.<generated>"`, and the generated suffix draws from a charmap of `1`-`5` and `a`-`z`, which contains no dot. `name::prefix()` cuts at the rightmost dot, so the creator is recoverable from any sponsored username by string arithmetic — no table read, and correct even when the creator's own name contains dots.

The only lookup the contract performs against the table is the point read behind the `(creator, nonce)` replay guard, which the primary key already serves. The index's sole unique capability was username → nonce within a scope the caller already had to know, and the nonce is the sponsor's own idempotency key.

## Effect

Halves the per-sponsorship RAM cost, ~296 → ~152 bytes. That matters more than the raw number: `sponsors` and `sponsorcount` are the only rows in `sysio.roa` billed to a node owner rather than to the contract, so this is the sole RAM cost a tier-1 owner carries for onboarding users.

## Scope

The ABI change is confined to dropping the `secondary_indexes` block from the `sponsors` table — 8 lines out, 1 in, nothing else.

Generated TypeScript types need no update. They encode ABI types rather than indexes, and `SysioRoaSponsorType` is byte-identical before and after:

```ts
export interface SysioRoaSponsorType {
  nonce: string
  username: string
}
```

Pre-launch, so there are no stored rows to migrate.

## Verification

`unit_test`, `plugin_test`, `contracts_unit_test`, and the `sysio_roa_tests` suite were run against the rebuilt contract. The deep-mind and snapshot reference data needed no regeneration — `sysio.roa` is not part of those reference chains.

## Risk

`get_table_rows` by secondary index is public API surface. Any explorer or third-party tool outside these repos that queries `sponsors` by `byusername` would break. Nothing visible from here does.


huangminghuang (approved):
Reviewed the source change, generated ABI, and committed WASM artifact. No actionable findings; the primary-key access pattern remains intact and the removed secondary index has no relevant repository consumers.

PR #585, opened 2026-08-24
Roa: provision a personal policy only for tier-1 node owners

heifner:
## Summary

`regnodeowner` now provisions a personal self-issued policy only for tier-1 node owners. Tiers 2 and 3 are registered with their `nodeowners` budget, a `reslimit` row, and the 10% `sysio` RAM grant, but no allocation of their own.

## Why tiers 2 and 3 do not need one

Managing policies costs a node owner nothing. `addpolicy`, `expandpolicy`, `extendpolicy`, and `reducepolicy` are actions on `sysio.roa`, so `action::payer()` resolves to `sysio.roa`, which carries unlimited CPU and NET. Every row those actions write is emplaced with `get_self()` as the RAM payer — 21 of the 23 row writes in the contract bill to `sysio.roa`.

Confirmed directly: a registered node owner reduced to `ram = usage, net = 0, cpu = 0` issued a 1/1/1 SYS policy successfully, and its CPU used stayed at `0`.

Membership — the `nodeowners` row — is what confers the ability to issue, not any allocation the owner holds. A tier-2 or tier-3 owner can self-issue on demand against the recorded tier budget.

## Why tier 1 keeps it

Tier 1 is the only tier that can call `newuser`, and `newuser`'s `sponsors` and `sponsorcount` rows are the only writes in the contract billed to a node owner instead of to `sysio.roa`. A tier-1 owner therefore needs RAM headroom before its first sponsorship.

## Why it was disproportionate

The personal weights are a flat constant (80 units RAM, 500 each NET/CPU) while tier budgets are a fraction of supply. Tier 1 draws 4% of supply against tier 3's 0.003%, so the identical deduction is roughly three orders of magnitude more significant for a tier-3 owner. Against the launch configuration it consumed 4.77% of a tier-3 allocation versus 0.0036% of a tier-1's, and returning it raises a tier-3 owner's issuable budget by 5.6%.

## Behaviour changes

- Tiers 2 and 3 hold no personal policy, zero NET and zero CPU, and only the `newaccount_ram` their account was created with.
- Their `nodeowners` accounting excludes the personal weights: `allocated_bw` is zero and `allocated_sys` equals the 10% grant, leaving the remainder issuable.
- The `sysio` RAM grant is unchanged for every tier. Its guard was split from the personal policy's so it is no longer gated behind a policy tiers 2 and 3 never create.
- `increase_reslimit` is still called with zero weights, so the `reslimit` row is created for every tier. That keeps it a complete registry of node owners and preserves the row-equals-quota invariant; the following `set_resource_limits` is a no-op against what `native::newaccount` set.

## Tests

Adds `regnodeowner_personal_policy_is_tier1_only`, covering: tier 1 retains its personal policy; tiers 2 and 3 have none; every tier still carries the 10% `sysio` grant; tier 2/3 hold zero bandwidth and exclude the personal weights from their accounting; and a tier-3 owner can still self-issue via `addpolicy`, with the resulting quota landing on chain.

Two existing SEC-087 reconcile tests registered at tier 2 and asserted the personal weights stacking on top of a planted policy. Their expectations are updated — the planted policy now survives the reconcile unchanged. What they exist to prove, that a pre-existing `reslimit` row is reconciled rather than aborting the claim, is unaffected.

`unit_test`, `plugin_test`, `contracts_unit_test`, and the `sysio_roa_tests` suite were run against the rebuilt contract.

## Note

Pre-launch, so no already-registered owners need migrating. This touches `sysio.roa` alongside the `byusername` index removal; whichever lands second will need the wasm rebuilt rather than a side picked.


huangminghuang (dismissed):
Reviewed the contract changes, tier accounting, reslimit reconciliation, authorization path, tests, and committed WASM. No actionable findings.

contracts/sysio.roa/sysio.roa.cpp
```diff
@@ -860,18 +877,23 @@ namespace sysio {
         allocated_sys += sysio_allocation;
         allocated_ram += sysio_allocation; // Also RAM allocation since it's for sysio policy
 
-        // Minimal default net/cpu: 0.0500 SYS each for NET and CPU (0.1000 SYS total)
-        asset net_cpu_weight(500, state.total_sys.symbol); // 0.0500 SYS
+        // Minimal default net/cpu for a tier-1 owner: 0.0500 SYS each. Zero for tiers 2 and 3.
+        // Adding a zero asset below is a no-op, so the nodeowners totals stay correct for every
+        // tier without branching the accounting.
+        asset net_cpu_weight(provision_personal ? 500 : 0, state.total_sys.symbol);
```
huangminghuang: [P2] Preserve a CPU bootstrap path for tier-2/3 owners

Setting CPU weight for the owner to zero makes `addpolicy` unreachable on a supported producer configuration where subjective API/P2P billing is enabled and `--subjective-account-cpu-allowed-us=0`. Objective billing uses `sysio.roa`, but `transaction_context::verify_init_subjective_billing()` separately checks the first authorizer; with zero CPU the owner is rejected before `addpolicy` can grant CPU. Please retain enough bootstrap CPU for one policy action, or provide another path whose admission does not depend on pre-existing owner CPU.

heifner: The mechanism is real and correctly identified -- `verify_init_subjective_billing()` does check first-authorizers that are not payers, and a zero-CPU owner reaches it with `available = allowance + 0 - 0`.

First, a correction to my own evidence. The "a zero-CPU owner can still issue" result quoted in the commit message did not actually cover your scenario: that test used the tester's default `billed_cpu_time_us = 2000`, which sets `explicit_billed_cpu_time` and makes `verify_init_subjective_billing()` return on its first line. It never reached the subjective path. Re-run pushing with `billed_cpu_time_us = 0` so the check genuinely engages:

```
tier-2 owner limits: ram=1144 net=0 cpu=0
subjective allowance = 300000us
target after addpolicy: net=100 cpu=100
issuer still: net=0 cpu=0
```

At the default allowance the owner is admitted and issues while holding nothing.

On `--subjective-account-cpu-allowed-us=0` specifically: it is not a configuration this chain can run under, independent of this change. Every Wire user account holds zero objective CPU by design (`native::newaccount` sets `0, 0, 0`), and an ordinary user is a non-payer first authorizer, so the same check rejects them on the same arithmetic. An ordinary user calling a fully provisioned contract:

```
allowance=300000 -> ordinary user OK
allowance=0      -> Subjectively terminated trx ... Authorized account alice exceeded
                    subjective CPU limit 0us by 0us with an objective cpu limit of 0us.
allowance restored -> ordinary user OK again
```

That is the gasless user path, not a node-owner edge case. Under `=0` nobody transacts at all, so retaining bootstrap CPU for tiers 2 and 3 would not yield a working chain -- it would only change which account fails first. At any allowance above 0 the bootstrap is not needed. Leaving this as is.

One coverage note that fell out of the investigation: no test using the standard tester helpers exercises subjective billing, since they all pass a non-zero billed CPU. Reaching that path requires pushing with `billed_cpu_time_us = 0`.

heifner: Added coverage for this in `100acf66a2`.

`zero_cpu_owner_issues_under_subjective_billing` registers a tier-2 owner, asserts it holds zero NET and CPU, and issues a policy with subjective billing enabled -- pushed through a new `push_subjective_action` helper that passes `billed_cpu_time_us = 0` so `verify_init_subjective_billing()` actually runs. It then drives the allowance to zero and asserts the identical push is rejected with `Subjectively terminated trx`.

That negative half is the point of the test. Without it the positive assertions would pass just as happily if `explicit_billed_cpu_time` had short-circuited the check -- which is exactly how my original evidence went wrong, so the test proves the gate is reached rather than assuming it.

One clarification on my previous reply: I said no test exercises subjective billing. That was accurate about the standard tester helpers but understated what exists -- `subjective_billing_integration_test` covers the chain-level case well, including an account created with `net_weight` and `cpu_weight` of zero, pushed with `explicit_billed_cpu_time` false, driven until admission is refused. So the gap was specific to the roa suite, not chain-wide, and I have not duplicated the chain-level coverage.

ROA suite is at 63 cases.

huangminghuang (approved):
Re-reviewed current head 670d406a50 after the subjective-billing coverage and master merge. The prior P2 is not actionable under Wire resource semantics. Contract logic, regression coverage, rebuilt WASM/ABI, and all required CI checks are clean.

PR #586, opened 2026-08-25
Test: cover sysio.code native auth actions and session-key limits

heifner:
Covers a path that had none — a contract inline-sending the native auth actions (`updateauth`, `deleteauth`, `unlinkauth`) authorized by a `<contract>@sysio.code` seat — and pins the real constraints of the session-key pattern that path enables.

## The gap

`sysio.code` is well covered already: `create_account(include_code=true)` bakes it into every test account, and the roa, opreg, dispute, dispatch and msig suites drive cross-account inline actions through it. What none of them exercise is `sysio.code` against the *native* auth actions, which take the `special_case` branch of `authorization_manager::check_authorization` into `check_updateauth_authorization` / `check_deleteauth_authorization` — where the signer must satisfy the permission being modified — rather than the linked-permission lookup every existing test hits.

## Topology

Review of the first revision found the obvious arrangement unsafe. With the session key and the `sysio.code` seat in the **same** permission, the key can rewrite that authority and drop the seat — a permission satisfies itself — locking the contract out while the `linkauth` survives untouched. `combined_permission_lets_key_evict_reaper` demonstrates it: the same contract reaps an untouched permission, then is refused on an identical one the moment its key rewrites the authority.

The suite therefore uses a code-only parent over a key-only child:

| permission | parent | authority |
|---|---|---|
| `alice@sessgate` | `active` | `[{sessmgr, sysio.code}]`, no keys |
| `alice@session` | `alice@sessgate` | `[session key]`, no code seat |

The key may still rewrite its own permission, but there is nothing there to take, and it cannot reach the parent *in place* because a child never satisfies its parent. The reaper reaches every descendant, since `satisfies` walks up the target's ancestry — so a child created to block `deleteauth` defers cleanup rather than preventing it.

## Reaping is not guaranteed

The parent split stops in-place eviction, not escape. Top-level authorization is decided for every action up front against **pre-state**, so a transaction signed by the session key alone can `unlinkauth` itself, `deleteauth` itself, `updateauth` the same permission name under a *different* parent, and `linkauth` it again. Precheck sees the original child throughout, so every action is authorized; by execution the delete has run, so `updateauth` takes its create branch and accepts the new parent. The `"Changing parent authority is not currently supported"` guard never fires, because it sits inside `if (permission)`.

The permission ends up outside the gate's subtree and the contract can never clean it up. `session_key_escapes_gate_by_recreation` covers it.

What the escape does **not** change is authority: the recreated permission still only satisfies what it is linked to. So a real TTL has to be enforced by the consuming contract at use time, and the parent split should be read as hygiene and RAM recovery as best-effort — not as an enforceable expiry boundary.

## Coverage

`unittests/auth_tests.cpp` calls `check_authorization` directly with `provided_permissions` shaped as `apply_context::execute_inline` builds it: the seat authorizing `deleteauth`, `updateauth` and `unlinkauth` on descendants; absent or foreign `sysio.code` failing; the child unable to reach or delete the parent; and the narrower-than-it-sounds property that an unlinked child cannot authorize *linked* actions while still acting on itself.

`contracts/tests/sysio.inline_auth_tests.cpp` drives the same paths end to end through the existing `sendinline` forwarder, with an unrelated account signing the outer transaction to show cleanup is permissionless.

## Deployment constraints

`sysio.roa_tests.cpp` pins what standing this up actually requires — the authorization tests run without RAM limits and would otherwise imply it works on any account:

- **It does not fit in the creation gift.** ~770 bytes against 140 free. Even the code-only parent alone exceeds it.
- **The contract cannot pay.** `updateauth`, `linkauth` and `deleteauth` all bill the permission's owner; none takes a payer parameter, and `sysio.payer` only authorizes charging another account rather than redirecting these.
- **A ROA policy is required**, and only a node owner may issue or reclaim one.
- **Grant and spend can share a transaction** — RAM is verified once at finalize, not per action — so a grant cannot be intercepted.
- **Two top-level `updateauth` actions cannot create a permission and its child together**, since the child's parent lookup runs against pre-state. That is a property of the top-level path only: an inline action is authorized during execution, so grant, gate, inline-created child and link all fit in one transaction (`session_onboarding_single_transaction_via_inline`).
- **`sysio.payer` cannot be attached to the native auth actions.** `validate_referenced_accounts` requires the payer to also hold a real authorization on the action, and adding one exceeds the single authorization those actions accept. Both barriers are pinned, alongside a funded contract inline-creating a permission and leaving the charge on the user.

No new test contract and no CMakeLists change — `sendinline` already exists with its wasm committed, and the `contracts/tests` glob is `CONFIGURE_DEPENDS`.

## Note

The figures above shift with #587, which corrects RAM billing for permissions and keys. Assertions are written as existence and exception checks rather than literal byte counts, so they hold either way.


contracts/tests/sysio.inline_auth_tests.cpp
```diff
@@ -0,0 +1,174 @@
...
+   chain.set_code( account, system_contracts::testing::test_contracts::sendinline_wasm() );
+   chain.set_abi( account, system_contracts::testing::test_contracts::sendinline_abi().data() );
+}
+
+/// Create `alice@session` as a child of active: a session key plus a `sysio.code` seat for
+/// @p grantee.
+void grant_session_permission( validating_tester& chain, account_name grantee ) {
+   const authority session_auth( 1,
+                                 {{chain.get_public_key( alice, "session" ), 1}},
+                                 {{{grantee, config::sysio_code_name}, 1}} );
```
huangminghuang: **[P1] The session key can remove its own reaper.** This 1-of-{session key, `sessmgr@sysio.code`} topology lets the session key self-authorize `updateauth` and replace the entire authority, removing `sessmgr` while the existing action link remains usable. I reproduced this locally: the replacement key continued authorizing `reqauth`, while `sessmgr@sysio.code` could no longer authorize cleanup. The key can also create descendants that prevent deletion and the claimed RAM refund. Please use a code-only parent permission with a key-only session child, add an adversarial regression proving the child cannot eject the parent, and describe deletion/RAM recovery as best-effort rather than an adversarial guarantee.

unittests/auth_tests.cpp
```diff
@@ -812,4 +812,108 @@ BOOST_AUTO_TEST_CASE( authority_threshold_not_lowered ) { try {
...
+
+   // The seat lives on a child of active. It must not let sessmgr rewrite active itself -- that
+   // would promote a scoped grant into full control of the account.
+   const action escalate{ {{"alice"_n, "session"_n}},
+                          updateauth{ .account    = "alice"_n,
+                                      .permission = config::active_name,
+                                      .parent     = config::owner_name,
+                                      .auth       = authority( chain.get_public_key( "alice"_n, "attacker" ) ) } };
+   BOOST_CHECK_THROW( authmgr.check_authorization( {escalate}, {}, {{"sessmgr"_n, config::sysio_code_name}} ),
+                      irrelevant_auth_exception );
```
huangminghuang: **[P2] The positive `sysio.code`-authorized `updateauth` path is not covered.** This case targets `active` and is rejected by the hierarchy check before `checker.satisfied` evaluates `provided_permissions`. A regression in the allowed `updateauth` path would therefore remain green despite the PR claiming coverage for `updateauth`, `deleteauth`, and `unlinkauth`. Please add direct and end-to-end success cases, ideally for the code-only parent updating its session child.

huangminghuang (commented):
I found one security-sensitive flaw in the proposed session/reaper topology, one missing positive coverage path, and one overbroad test claim. Details are inline.

unittests/auth_tests.cpp
```diff
@@ -812,4 +812,108 @@ BOOST_AUTO_TEST_CASE( authority_threshold_not_lowered ) { try {
...
+   validating_tester chain;
+   chain.create_accounts( {"alice"_n, "sessmgr"_n} );
+   chain.produce_block();
+   setup_session_permission( chain, "sessmgr"_n );
+
+   const auto session_priv_key = chain.get_private_key( "alice"_n, "session" );
+
+   // With no link the minimum permission defaults to active, and a child of active does not
+   // satisfy active -- so an unlinked session key cannot sign anything, whoever holds it.
+   BOOST_CHECK_THROW( chain.push_reqauth( "alice"_n, {permission_level{"alice"_n, "session"_n}}, {session_priv_key} ),
```
huangminghuang: **[P3] An unlinked child permission is not fully inert.** This assertion proves only that it cannot authorize an ordinary `reqauth` action whose minimum permission defaults to `active`; native special cases still let the child key update or delete itself and create descendants. Please rename/reword the test and PR description to state the narrower property.

contracts/tests/sysio.inline_auth_tests.cpp
heifner: Confirmed, and this was the right catch — the pattern as written was unsafe. `check_updateauth_authorization` takes the *existing* permission as `min_permission`, and `satisfies` returns true for `id == other.id`, so the session key authorizes rewriting its own authority and drops the seat while the `linkauth` survives as a separate object. Your descendants point holds too: `remove_permission` asserts "Cannot remove a permission which has children."

Restructured to a code-only parent over a key-only child, as suggested — `alice@sessgate` holds `[{sessmgr, sysio.code}]` with no keys, `alice@session` holds the key with no seat. The key can still rewrite its own permission but there is nothing there to take, and it cannot reach the parent. `session_key_cannot_eject_reaper` rotates the child with the session key, proves it can neither rewrite nor delete the gate, and shows the reap still succeeds.

I also added `combined_permission_lets_key_evict_reaper`, which executes the exploit against the old topology: the contract reaps an untouched combined permission, then is refused on an identical one the moment its key rewrites the authority. It seemed worth keeping the failure mode executable so the reason for the split does not decay into a comment.

On deletion and RAM: reworded to deferred rather than guaranteed. `descendant_defers_but_does_not_block_reap` covers it — a child does block `deleteauth`, but the gate satisfies grandchildren, so the reaper unwinds depth-first.

unittests/auth_tests.cpp
heifner: Correct — that case targets `active` and dies in the hierarchy check before `checker.satisfied` ever consults `provided_permissions`, so it proves nothing about the allowed path. Added `code_parent_satisfies_child_updateauth` at the `check_authorization` level and `code_parent_can_update_child` end to end, both the gate rotating its child's key, with the end-to-end one asserting the rotated key actually lands in the stored authority.

unittests/auth_tests.cpp
heifner: Agreed, the name overclaimed. Renamed to `unlinked_child_cannot_authorize_linked_actions`, and it now also asserts what the child *can* still do with no link — rotate itself via the native special case — so the narrowness is demonstrated rather than just described. PR description updated to match.

heifner:
All three findings were valid and are addressed in the amended commit; replies are inline on each thread.

One PR-wide note, since review prompted a scope addition. The authorization suites run without RAM limits, so they proved the mechanism works while saying nothing about whether anyone can actually deploy it. `sysio.roa_tests.cpp` now pins that: the topology does not fit in the account creation gift, the contract cannot pay for it (the native auth actions all bill the permission owner), and a node-owner-issued ROA policy is required.

That last test also turned up something I had assumed the other way. A grant and its spend *can* share a transaction, because RAM is verified once at finalize rather than per action — but a permission and its child *cannot*, because authorization is checked for every action before any of them executes, so the child's parent lookup runs against a parent that does not yet exist. I found it by writing the test and getting `permission_query_exception` where I had predicted `ram_usage_exceeded`. Onboarding is therefore two transactions, with the grant atomic against the first.

Figures cited in the description shift with #587, which corrects RAM billing for permissions and keys; the assertions are existence and exception checks rather than literal byte counts, so they hold either way.

contracts/tests/sysio.inline_auth_tests.cpp
```diff
@@ -0,0 +1,264 @@
...
+   BOOST_REQUIRE_EXCEPTION(
+      set_authority( alice, code_parent, authority( get_public_key( alice, "attacker" ) ), config::active_name,
+                     { permission_level{alice, key_child} }, { get_private_key( alice, "rotated" ) } ),
+      irrelevant_auth_exception,
+      fc_exception_message_starts_with( "updateauth action declares irrelevant authority" ) );
+
+   // Reaper still wins.
+   reap( sessmgr, key_child );
+   produce_block();
+   BOOST_REQUIRE( find_perm( key_child ) == nullptr );
```
huangminghuang: **[P1] The session key can still eject the gate atomically.** Transaction authorization is checked against pre-state for every top-level action. A transaction signed by the existing session key can therefore contain `unlinkauth(session)` → `deleteauth(session)` → `updateauth(session, parent=active)` → `linkauth(session)`, all declaring `alice@session`. Precheck sees the old child/link and accepts every action; execution then removes the link and child, recreates the same permission name as a sibling of `sessgate`, and relinks it. The gate no longer satisfies the recreated permission, so the session survives expiry and the reaper is locked out. Please add this batched regression and avoid presenting the parent split as an enforceable TTL boundary unless the production TOCTOU is fixed or the workflow prevents top-level session-signed native actions.

contracts/tests/sysio.roa_tests.cpp
```diff
@@ -858,6 +860,91 @@ class sysio_roa_full_tester : public sysio_roa_tester {
...
 };
 
+// Session-key onboarding: what can and cannot share a transaction.
+//
+// RAM is verified once at transaction finalize rather than per action, so a ROA grant and the
+// spending of that grant may ride in the same transaction -- nothing can land in between and
+// consume it. Authorization, by contrast, is checked for every action BEFORE any of them execute
+// (controller checks the whole action list, then calls trx_context.exec()), so a permission and
+// its child cannot be created together: the child's parent lookup runs while the parent does not
+// yet exist. Onboarding therefore takes two transactions, with the grant atomic against the first.
```
huangminghuang: **[P1] The two-transaction onboarding conclusion is false for the contract-managed path.** In one transaction, `addpolicy` can grant RAM, top-level `updateauth` can create `sessgate`, and a subsequent call to `sessmgr` can inline-create `session` under `{user, sessgate}`. Inline authorization happens during execution after the gate exists and runs depth-first, so a following top-level `linkauth` can see the child. Add this positive one-transaction regression and narrow/remove the categorical claim; the current split unnecessarily exposes the unused grant between transactions.

contracts/tests/sysio.roa_tests.cpp
```diff
@@ -858,6 +860,91 @@ class sysio_roa_full_tester : public sysio_roa_tester {
...
+      return push_transaction( trx );
+   };
+
+   const auto& authmgr = control->get_authorization_manager();
+
+   // Without a grant, even the code-only parent alone exceeds the creation gift.
+   {
+      signed_transaction trx;
+      trx.actions.emplace_back( as_user, make_gate() );
+      BOOST_CHECK_THROW( sign_and_push( trx ), ram_usage_exceeded );
```
huangminghuang: **[P2] The advertised contract-payer constraint is not pinned.** This only proves that a top-level user-paid `updateauth` exceeds the creation gift. `sessmgr` is not deployed or invoked and `sysio.payer` is never included, so a regression that redirected native-auth RAM to the contract would leave this test green. Add a funded inline sender case proving the user still incurs the permission/link RAM while the contract balance is unchanged, and explicitly cover the native action's payer authorization rejection, or narrow the PR description.

contracts/tests/sysio.inline_auth_tests.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+using mvo = fc::mutable_variant_object;
+
+/// End-to-end coverage for a contract driving the *native* auth actions through `sysio.code`.
+///
+/// The account holder grants `<contract>@sysio.code` a seat in a child permission; that contract
+/// can then inline-send `unlinkauth` / `deleteauth` naming `{holder, child}` as the authorization,
+/// because a permission satisfies itself and the seat supplies the weight. This is the mechanism
+/// behind contract-managed session keys, where a permissionless reaper cleans up a permission
+/// whose lease has run out.
```
huangminghuang: **[P3] This header still describes the rejected topology.** The revised suite puts the `sysio.code` seat in a parent and authorizes descendant cleanup through hierarchy; it no longer puts the seat in the child or relies on that permission satisfying itself. Update this overview so it does not re-document the unsafe arrangement that the tests below now avoid.

huangminghuang (commented):
Re-review findings so far: two high-severity workflow/security issues remain, plus one unpinned deployment claim and one stale header comment. Details are inline.

contracts/tests/sysio.inline_auth_tests.cpp
heifner: Confirmed, and this one is not a test bug — it defeats the design I proposed. `session_key_escapes_gate_by_recreation` executes your batch: unlink -> delete -> recreate under `active` -> relink, signed by the session key alone. Precheck authorizes all four against pre-state where the old child exists; by execution the delete has run, so `apply_sysio_updateauth` finds `nullptr`, takes its create branch, and accepts the new parent. The re-parenting guard never fires because it lives inside `if (permission)`. The gate ends up a sibling and cleanup is lost permanently.

I have stopped presenting the split as a TTL boundary. The file header, the topology comment, the commit message and the PR description now all say reaping is best-effort against a cooperative holder.

One qualification worth recording: the escape costs the attacker nothing but also buys them nothing. The recreated permission still only satisfies what it is linked to, so authority is unchanged — what is lost is the contract's ability to clean up, not containment. That is why the real TTL has to be the consuming contract checking expiry at use time, which no amount of permission surgery can bypass.

contracts/tests/sysio.roa_tests.cpp
heifner: You're right and my conclusion was over-general. `session_onboarding_single_transaction_via_inline` now proves the positive case: `addpolicy`, a top-level `updateauth` for the gate, an inline `updateauth` from the contract creating the child under `{user, sessgate}`, and a trailing top-level `linkauth` — all in one transaction. The link's precheck resolves the currently linked permission (none) and only `apply_sysio_linkauth` needs the requirement to exist, by which point the inline has created it. The categorical claim is narrowed to the top-level path in both the test comment and the description.

contracts/tests/sysio.roa_tests.cpp
heifner: Correct — the old test only showed the user could not afford it, which a payer-redirect regression would have sailed past. `contract_cannot_pay_for_permission_ram` now deploys and funds the forwarder, has it inline-create a permission through its seat, and asserts the charge lands on the user with the contract's usage unchanged.

It also pins two independent barriers against `sysio.payer`, which turned out to be more interesting than expected: `validate_referenced_accounts` requires the payer to also hold a real authorization on the same action, and supplying one pushes past the single authorization the native auth actions accept. No arrangement satisfies both, so the mechanism is unreachable for these actions rather than merely unused.

contracts/tests/sysio.inline_auth_tests.cpp
heifner: Fixed. The file header was still describing the seat-in-the-child arrangement while the namespace comment twelve lines below described the parent split — two contradictory topologies in one file. It now documents the parent split and leads with the reaping caveat.

huangminghuang (dismissed):
Re-reviewed the amended head: previous findings are addressed and focused regression coverage passes locally.

contracts/tests/sysio.inline_auth_tests.cpp
```diff
@@ -0,0 +1,327 @@
...
+   trx.actions.emplace_back( as_child, linkauth{ alice, config::system_account_name, "reqauth"_n, key_child } );
+   set_transaction_headers( trx );
+   trx.sign( get_private_key( alice, "session" ), control->get_chain_id() );
+   push_transaction( trx );
+   produce_block();
+
+   // Same name, same key, now a sibling of the gate rather than its child.
+   const auto* escaped = find_perm( key_child );
+   BOOST_REQUIRE( escaped != nullptr );
+   BOOST_REQUIRE( escaped->parent != find_perm( code_parent )->id );
```
huangminghuang: **[P2] Prove the recreated session still has the linked capability.**

The transaction being accepted only proves all four top-level actions were authorized against the pre-state child. The postconditions show that `session` was recreated outside `sessgate` and that the reaper is locked out, but they never exercise the new permission or verify the `reqauth` link. If the final `linkauth` became a no-op or pointed at the wrong requirement, this test would still pass even though the claimed usable escape no longer exists. Please invoke `push_reqauth` with the session key after recreation (or at minimum assert `lookup_linked_permission(...) == key_child`); the behavioral call pins both the relink and unchanged key.

contracts/tests/sysio.roa_tests.cpp
```diff
@@ -874,6 +876,246 @@ class sysio_roa_full_tester : public sysio_roa_tester {
...
+                   ( "action_name", "updateauth" )
+                   ( "auths", std::vector<permission_level>{{user, "sessgate"_n}} )
+                   ( "payload", fc::raw::pack( updateauth{ .account = user, .permission = "session"_n,
+                                                           .parent  = "sessgate"_n, .auth = session_auth } ) ) );
+   produce_block();
+
+   BOOST_TEST( rm.get_account_ram_usage( user ) > user_before );      // the owner pays
+   BOOST_TEST( rm.get_account_ram_usage( sessmgr ) == mgr_before );   // the contract does not
+
+   // Nor can a payer role be attached. Two independent barriers make sysio.payer unreachable for
```
huangminghuang: **[P2] Cover payer rejection for each native auth action.**

These vectors exercise only `updateauth`, but the comment and PR description generalize the conclusion to all native auth actions. `check_deleteauth_authorization`, `check_linkauth_authorization`, and `check_unlinkauth_authorization` each contain their own independent `auths.size() == 1` guard, so a regression in any one would leave this test green while making `sysio.payer` attachable on that path. Please parameterize the payer-only and paired-payer cases across all four actions (with valid pre-state for delete/link/unlink), or narrow the claimed constraint to `updateauth`.

huangminghuang (commented):
Two test-coverage findings.

contracts/tests/sysio.inline_auth_tests.cpp
heifner: Right — the postconditions showed the permission was recreated outside `sessgate` and that the reaper was locked out, but nothing exercised the result. A final `linkauth` that had become a no-op, or that named the wrong requirement, would have left this green while the usable escape it describes did not exist.

Added the behavioral call you suggested: after recreation the test finishes with `push_reqauth` under `{alice, session}`, signed with the original session key. That pins the relink and the surviving key together rather than asserting either on its own.

contracts/tests/sysio.roa_tests.cpp
heifner: Correct, and I had generalized from a single case. Each of the four carries its own `auths.size() == 1` guard with its own message — `updateauth`, `deleteauth`, `link`, `unlink` — so a regression in any of the other three would have left this test passing while making `sysio.payer` attachable on that path.

Both barriers are now exercised across all four actions, with pre-state (a permission to delete, one to link, an existing link to unlink) so each rejection is about the payer rather than about missing state. The per-action assert messages double as the discriminator: if a case failed to reach its own guard, the match would fail rather than silently pass.

Building that pre-state exposed a defect in my own setup worth recording, since it is the same shape as the finding: `link_authority()` takes no keys and signs as `get_private_key(user, "active")`, but `create_newuser` seeds the account with its creator's key — so the pre-state link had been failing silently. It is now an explicitly signed transaction.

contracts/tests/sysio.roa_tests.cpp
```diff
@@ -874,6 +876,283 @@ class sysio_roa_full_tester : public sysio_roa_tester {
...
+   trx.sign( signing_key, control->get_chain_id() );
+   push_transaction( trx );
+   produce_block();
+
+   const auto& authmgr = control->get_authorization_manager();
+   const auto* gate  = authmgr.find_permission( {user, "sessgate"_n} );
+   const auto* child = authmgr.find_permission( {user, "session"_n} );
+   BOOST_REQUIRE( gate != nullptr );
+   BOOST_REQUIRE( child != nullptr );
+   BOOST_REQUIRE( child->parent == gate->id );
```
huangminghuang: **[P2] Assert the onboarding link, not only permission creation.**

Both onboarding cases execute a final `linkauth`, but their postconditions only show that the gate and child permissions exist (and here, that the child has the expected parent). If that final link becomes a no-op, is omitted, or names the wrong requirement, both tests remain green even though `user@session` cannot authorize `sysio::reqauth`, so they do not pin the complete onboarding result described by the PR. Please exercise `push_reqauth` with `user@session` and the session key after each flow, or directly assert the persisted permission-link state.

heifner: Right, and this is the same gap you caught in the escape test one round earlier — I fixed the instance you pointed at and did not sweep for the pattern. Both onboarding cases had it: they asserted the gate and child existed, and in one case that the hierarchy was correct, but nothing exercised the `linkauth` that makes the session usable. A no-op or wrong-requirement link would have left them green.

Added `check_session_is_live`, called at the end of both flows. It builds the `reqauth` action directly and runs it through `check_authorization` with the session key — `push_reqauth` is not usable here because `sysio` carries the system contract in this fixture rather than bios, so `reqauth` is absent from its ABI (the payload is just the account name).

That single call covers both halves of what you asked for. With no link, or one naming a different requirement, the minimum permission for `sysio::reqauth` falls back to `active`, which a child of the gate does not satisfy — so a broken link throws `irrelevant_auth_exception` rather than passing silently. I had also asserted the link explicitly via `lookup_linked_permission`, but it is private and turned out redundant.

Verified by pointing the `linkauth` at `sessgate` instead of `session`: both cases fail at exactly that assertion.

contracts/tests/sysio.roa_tests.cpp
```diff
@@ -872,8 +874,319 @@ class sysio_roa_full_tester : public sysio_roa_tester {
...
+      // requirement, the minimum permission for sysio::reqauth falls back to active, which a child
+      // of the gate does not satisfy -- so the call below would throw irrelevant_auth_exception.
+      action reqauth_act;
+      reqauth_act.account       = config::system_account_name;
+      reqauth_act.name          = "reqauth"_n;
+      reqauth_act.authorization = { permission_level{user, "session"_n} };
+      reqauth_act.data          = fc::raw::pack( user );
+
+      BOOST_CHECK_NO_THROW( authmgr.check_authorization(
+         { reqauth_act }, { get_public_key( issuer, "session" ) } ) );
```
huangminghuang: **[P2] Verify that `reqauth` resolves exactly to `session`.**

This authorization-only check still succeeds if the final link points `sysio::reqauth` to `sysio.any`. In that case `lookup_minimum_permission` returns empty, `check_authorization` skips the minimum-permission relevance check, and the supplied session key still satisfies the declared `user@session`; the helper therefore blesses a wrong and overbroad requirement despite the comment claiming every wrong requirement fails. Please assert that `authmgr.lookup_minimum_permission(user, config::system_account_name, "reqauth"_n)` is present and equals `"session"_n` before checking the key.

heifner: Correct, and this is a case where I asserted the conclusion one message after getting it wrong. You are right about the mechanism: `lookup_minimum_permission` returns an empty optional for `sysio.any`, `check_authorization` guards the relevance check with `if (min_permission_name)`, and the key still satisfies the declared `user@session` — so a link pointing `reqauth` at `sysio.any` passed, blessing a requirement that is both wrong and overbroad.

My falsification had used `sessgate` as the "wrong requirement", which does fail, and I generalized from that single value to all wrong values. `sysio.any` is precisely the one that behaves differently.

Worse, I had this assertion and removed it. The first version resolved the link explicitly, hit a compile error, and I concluded it was redundant — but the error was that `lookup_linked_permission` is private. `lookup_minimum_permission`, which you named, is public. I reached for the wrong accessor and drew the wrong lesson from the failure.

The helper now requires `lookup_minimum_permission(user, sysio, reqauth)` to be present and equal `session` before checking the key, and the comment now says why satisfiability alone is insufficient. Verified by pointing both onboarding links at `sysio.any`: each case fails on `minimum.has_value()`, which is exactly the shape the previous check waved through.

huangminghuang (approved):
Re-reviewed the latest follow-up and the full effective merge. The explicit minimum-permission assertion closes the `sysio.any` false-positive while retaining the session-key liveness check. No remaining findings.

PR #587, opened 2026-08-25
Fix: correct RAM billing for permissions, keys and account init

heifner:
Audits every `billable_size` specialization in the tree against the object it actually bills, and fixes the three that were wrong. Pre-launch, so these are safe to change now; after launch each would need a protocol feature.

## What was wrong

**`permission_object` billed for a `permission_usage_object` that no longer exists, in two separate terms.** It charged five indices where `permission_index` declares four — the fifth was the usage object's index — and its fixed-field term was 64, being upstream's 48 bytes (including an 8-byte `usage_id`) plus the 16-byte usage row. `9dcb06b84b` removed the `usage_id` and the usage row along with `get_permission_last_used`; only the billing outlived them. The correct fixed term is **40**, exactly what `permission_object` holds outside its authority member (`id`, `parent`, `owner`, `name`, `last_updated`). Over-billed 48 bytes per permission.

**`key_weight` billed the serialized key, not the resident one.** The charge was a fixed `8` plus `fc::raw::pack_size(key)`, which mixes two models. K1, R1, EM and Ed25519 store their bytes *inside* the 56-byte `shared_key_weight` element, so `pack_size` double-counted them while the element itself was billed at 16. WebAuthn and BLS do keep their payload on the heap, so there the dynamic term is right but the element was under-charged by 40. The net was under actual memory for every reachable key type:

| key type | billed | resident | delta |
|---|---|---|---|
| K1 / R1 / EM | 50 | 56 | −6 |
| Ed25519 | 49 | 56 | −7 |
| WebAuthn | 18 + N | 64 + N | −46 |

It now charges the element plus the bytes held outside it. For the heap-backed alternatives that means chainbase's actual allocation — `shared_cow_string::_alloc` requests `sizeof(impl) + size`, where `impl` prepends an 8-byte `reference_count`/`size` header — so the resident cost of a WebAuthn key is `64 + N`, not `56 + N`. Billing the payload alone would leave heap-backed keys at exactly their resident footprint while inline keys carried the alignment surplus; counting the header puts every key type at resident + 8, matching what the `accounts` vector already did.

That header size is a fixed protocol constant rather than a `sizeof` read out of chainbase, for the same reason the element charge is: a layout change there must become a deliberate billing decision rather than a silent one.

**`initialize_account` skipped that rounding.** It returned `billable_size<T>::value` rather than `billable_size_v<T>` — the only call site in the tree that does, and upstream Antelope uses `_v` here. `resource_object` is unaffected at 160, but `resource_pending_object` is 104 and rounds to 112, so every account under-paid 8.

Audited and left alone as correct: `account_object`, `account_metadata_object`, `permission_link_object`, `kv_object`, `kv_index_object`, `resource_object`, `resource_pending_object`, `permission_level_weight`, `shared_authority`.

## Effect

Account creation goes from **1004 to 944** bytes:

| component | before | after |
|---|---|---|
| `account_object` | 96 | 96 |
| 2 × `permission_object` | 544 | 448 |
| owner authority (1 key) | 50 | 64 |
| active authority (1 key) | 50 | 64 |
| `resource_object` + `resource_pending_object` | 264 | 272 |
| **total** | **1004** | **944** |

## A bug this surfaces

The creation gift `newaccount_ram` is 1144. Under the old billing that left a fresh account with fixed-size keys 140 bytes against a `permission_link_object` costing 144 — **such an account could not perform a single `linkauth`**, so it could never scope a permission to a contract. Headroom is now 200, enough for one link with 56 to spare. This applies to the fixed-size key types (K1, R1, EM, Ed25519); a WebAuthn authority never leaves a link's worth of the gift free, before or after this change — see below.

`ram_tests/gifted_account_affords_linkauth` pins this. Reverting the three fixes makes it fail with the shortfall printed literally:

```
check ram_bytes - usage_before >= link_cost has failed [1144 - 1004 < 144]
```

It also pins both constants independently of the code under test: 944 for a fixed-size-key account and 144 for a `permission_link_object`, the latter with a `static_assert`. `new_account_ram_tests` already asserts 944 for the default K1 account, so asserting it here is what extends that pin to the R1, EM and Ed25519 accounts the test creates.

`newaccount_ram` is unchanged. Shrinking it to 1040, the next multiple of `bytes_per_unit`, would drop free space to 96 — still short of the 144 a link costs, so 1144 is worth keeping.

## WebAuthn

Billing a WebAuthn key honestly costs more than under-billing it, so an account whose owner and active are both WebAuthn now fits a shorter RPID than before — measured at `1030 + 2L` against `1010 + 2L`, narrowing the creatable band from 67 characters to 57. Longer RPIDs are not rejected by key validation (`post_init` only requires a non-empty RPID); they simply exceed the creation gift and need a ROA policy, like any other account that outgrows it.

`webauthn_account_rpid_bounded_by_gift` pins this. It measures the boundary *and* pins it against values derived independently from the protocol constants: 1032 bytes at RPID length 1, and a 57-character maximum landing exactly on the 1144-byte gift. Measuring alone was not enough, and the original claim that deriving made a billing change visible was backwards — an additive error in any term simply moved the derived boundary, leaving every assertion green. Raising the `shared_string` header from 8 to 16 would have shifted the maximum from 57 to 49 and still landed on 1144.

It also asserts the failure lands on `ram_usage_exceeded` rather than key validation, and that a WebAuthn account has less than a link's worth of gift left at *any* RPID length — which is why `gifted_account_affords_linkauth` is scoped to the fixed-size key types rather than claiming a universal invariant.

The `permission_object` correction pulls the other way and recovers part of the narrowing — the band was 49 before that term was fixed, and 65 before the string header was counted. Raising `newaccount_ram` would close the remaining ten characters, at the cost of enlarging every account's gift for the worst-case key shape. Deferred: longer RPIDs work today with a policy, and the gift can be raised later.

## A diagnostic fix

`check_deleteauth_authorization` reported the wrong action name. Its assert text was copied from `check_updateauth_authorization` and read `updateauth action declares irrelevant authority`, which misidentifies the failing action — especially in a transaction carrying both. Corrected to `deleteauth`.

`auth_tests/deleteauth_error_names_deleteauth` reaches it. No existing case did: the `delete_auth` cases cover a missing permission, a linked permission, and success, none of which produce a same-account authority that fails to satisfy the permission being deleted. The test creates two sibling permissions under `active` — neither an ancestor of the other, so neither satisfies the other — and deletes one while declaring the other.

## Reference data

Deep-mind and snapshot reference data are regenerated against the final billing, so the committed fixtures match what this branch produces rather than any intermediate state. `chain_id` and `head_block_num` are unchanged, so this is not a regenesis. The `shared_string` header change needed no regeneration of its own, since it moves only heap-backed keys and these fixtures use K1.

**The `blocks.log` diff is larger than these changes explain.** The 39,288-byte shrink decomposes exactly into the three contracts this fixture actually deploys — `sysio.bios` −13,105, `sysio.roa` −23,083, `snapshot_test` −3,100 — none of which changed in this PR. Their committed wasms are byte-identical to the last reference-data regeneration in `44965ec34d`, and the build reproduces them exactly, so the previous `blocks.log` had been carrying larger builds of those three than the tracked sources produce. Deleting it and regenerating per the documented procedure brings the fixture back in step.

Nothing caught the drift because `blocks.log` and `snap_v1` were generated together and are self-consistent; nothing asserts the log matches current contracts.

**Regenerated again after integrating `master`.** Review caught that the same self-consistency hides a second drift across a merge. This branch was 154 commits behind, and `master` had rebuilt `sysio.roa` via #584 and #585 — a conflict-free merge selects `bc23439825ad36e2`, while fixtures generated before it embed the merge-base's `569dfaf8cf0e97da`. `libraries/testing` embeds its own copy of `sysio.roa` from the build directory via `INCBIN`, so the fixtures and the embedded contract move together and agree with each other while disagreeing with the tracked contract; the compatibility tests compare only the first pair.

The reference data is now generated from the merged tree. The old code hash no longer appears in `snap_v1.json.gz` and the new one appears in its place, the embedded and tracked contracts match, and `head_block_id` is updated to `0000001dc59bcae1…`. `chain_id` and `head_block_num` are unchanged, so this remains a non-regenesis. The deep-mind log re-derives byte-identically, since these fixtures deploy no contract this merge altered.

## Review notes

- The `key_weight` change is a billing **policy** shift (packed → resident), unlike the other two which are arithmetic miscounts. Worth separate scrutiny.
- Its element charge is a fixed protocol constant, deliberately not `sizeof(shared_key_weight)`: that type embeds a `std::variant` whose layout is a toolchain property, and deriving a consensus billing value from it would let a layout change cross a 16-byte bucket and make heterogeneous builds debit different RAM. The `static_assert` verifies the resident type still fits.
- `newaccount_ram = 1144` is duplicated in `contracts/sysio.system/include/sysio.system/native.hpp` and `libraries/testing/include/sysio/testing/tester.hpp`, joined only by a "should match" comment with no `static_assert`. Untouched here.


libraries/chain/include/sysio/chain/authority.hpp
```diff
@@ -189,8 +204,11 @@ namespace config {
 
    template<>
    struct billable_size<key_weight> {
-      // no sizeof guard: key billed dynamically via shared_authority::get_billable_size()
-      static const uint64_t value = 8; ///< over value of weight for safety, dynamically sizing key
+      /// Footprint of one shared_key_weight element in shared_authority::keys. Bytes the key holds
+      /// outside that element are added per key by shared_authority::get_billable_size().
+      static const uint64_t value = sizeof(shared_key_weight);
```
huangminghuang: **High — pin this consensus billing constant.**

`billable_size_v<key_weight>` participates in `newaccount` and `updateauth` RAM debits, but this value now comes from `sizeof(shared_key_weight)`. That type contains `std::variant`, so its layout is an ABI and standard-library property rather than a protocol constant. A layout change crossing a 16-byte billing bucket would make heterogeneous builds charge different RAM and could change transaction acceptance. The current `static_assert` is tautological because `value == sizeof(...)`.

Please use a fixed protocol value (56 on the current ABI) and assert that the resident type still fits it.

libraries/chain/include/sysio/chain/authority.hpp
```diff
@@ -283,7 +301,7 @@ struct shared_authority {
       size_t keys_size = 0;
       for (const auto& k: keys) {
          keys_size += config::billable_size_v<key_weight>;
-         keys_size += fc::raw::pack_size(k.key);  ///< serialized size of the key
+         keys_size += heap_size(k.key);  ///< bytes held outside the element (WebAuthn, BLS)
```
huangminghuang: **Medium — preserve or document WebAuthn account-creation limits.**

For a WebAuthn RPID length `L < 128`, the stored payload is `35 + L` bytes. With this new heap charge, an account using WebAuthn for both owner and active costs `1046 + 2L` bytes, so fixed `newaccount_ram = 1144` rejects `L >= 50`. Before this patch it cost `1008 + 2L`, so valid lengths 50–68 fitted. `public_key::post_init()` only requires a non-empty RPID.

Also, even one WebAuthn authority leaves less than the 144 bytes needed for `linkauth`, so the new K1-only regression test does not establish its stated invariant for every supported authority. Please add WebAuthn boundary coverage and either size the gift for a documented supported shape or enforce an explicit size limit.

huangminghuang:
**Reference-data provenance correction:** the PR description attributes the 39,288-byte `blocks.log` shrink to `sysio.msgch.wasm`, but that contract is not present in this snapshot fixture. The log-size delta decomposes exactly into `sysio.bios` −13,105 bytes, `sysio.roa` −23,083 bytes, and `snapshot_test` −3,100 bytes. Their snapshot hashes match the tracked WASMs, so the committed fixtures are internally consistent; only the description attribution needs correction.

libraries/chain/include/sysio/chain/authority.hpp
heifner: You're right, and this was a determinism bug I introduced. Pinned to a fixed protocol constant of 56, with the assert now able to fire — `static_assert(sizeof(shared_key_weight) <= value)` fails loudly if the resident type outgrows the constant, rather than being tautological against itself. Numerically identical on the current ABI, so no reference-data churn.

The comment records why it isn't `sizeof`: `shared_key_weight` embeds a `std::variant` whose layout is a toolchain and standard-library property, and deriving the charge from it would let a layout change cross a 16-byte billing bucket and make heterogeneous builds debit different RAM for the same authority. This also puts it back in line with the other specializations in the file, which all use a literal plus a fitting assert — a convention I had broken without noticing.

libraries/chain/include/sysio/chain/authority.hpp
heifner: Confirmed by measurement, and your arithmetic was right to within 2 bytes. I measured `1046 + 2L` after the patch against `1010 + 2L` before, so the creatable RPID band narrows from L <= 67 to L <= 49. That is a regression, not merely a limit.

Your second point holds too, and I had overclaimed: at *every* RPID length a WebAuthn account has less than the 144 bytes a link needs, both before and after this patch — so `gifted_account_affords_linkauth` never established its invariant for WebAuthn. It is now scoped explicitly to the fixed-size key types, pointing at the new coverage.

Added `webauthn_account_rpid_bounded_by_gift`. It derives the boundary from a measured shortest account rather than hardcoding it, so a future billing change moves it visibly; asserts the longest supported length fits and one more fails on `ram_usage_exceeded` rather than key validation; and pins the sub-144-byte remainder that scopes the other test.

On sizing the gift versus enforcing a limit — taking the documented-shape option. The under-billing this PR corrects was largest precisely for WebAuthn (about -39 bytes), so the narrowing is the accounting becoming honest rather than becoming wrong, and raising `newaccount_ram` would enlarge every account's gift to accommodate the worst-case key shape. Longer RPIDs work with a ROA policy, like any other account that outgrows the gift, and the gift can be raised later if WebAuthn use becomes real.

huangminghuang:
Re-reviewed current head `cfeb679c44`. The fixed `56` protocol constant and size guard resolve the ABI-dependent consensus-billing issue, and the new WebAuthn coverage correctly pins the gift boundary and RAM failure mode. I found no remaining code defect.

One P3 documentation/provenance issue remains: the **Reference data** section still attributes about 33 KB of the `blocks.log` shrink to `sysio.msgch.wasm`, but this fixture contains no `sysio.msgch` account or code. The full 39,288-byte reduction decomposes exactly as `sysio.bios` −13,105 bytes, `sysio.roa` −23,083 bytes, and `snapshot_test` −3,100 bytes. The snapshot code hashes match those three tracked WASMs, so the artifacts are internally consistent; please correct the narrative and avoid asserting a `sysio.msgch` cause.

heifner:
You're right, and thank you for actually decomposing it. `sysio.msgch` is not in this fixture at all — `snapshot_tests.cpp` deploys `snapshot_test`, plus bios and roa via the setup policy, and no msgch account or code appears anywhere in it. I checked that after reading your comment rather than before writing the claim, which is the mistake: I saw a roughly 33KB msgch rebuild land in the right timeframe, saw a roughly 39KB log shrink, and asserted causation from the matching magnitude without verifying the contract was even present.

Your decomposition sums exactly, and I confirmed the other half of it: the committed wasms for all three are byte-identical to `44965ec34d`, and the current build reproduces them exactly. So the previous `blocks.log` had been carrying larger builds of those three than the tracked sources produce, and deleting and regenerating per the documented procedure simply brings the fixture back in step with what is committed.

The Reference data section now states the decomposition and drops the msgch attribution entirely.

Separately, a third commit on this branch fixes something noticed while writing the session-key tests: `check_deleteauth_authorization` raised its irrelevant-authority assert with the text "updateauth action declares irrelevant authority", a copy of the sibling check, so a failed `deleteauth` reported the wrong action. It is its own commit since it is unrelated to the billing work.

huangminghuang (commented):
I found two material issues in the current head; details are inline. Also, the PR description does not mention commit `dd7e8fff` changing the observable `deleteauth` diagnostic or its new regression test. Please update the description to capture that behavior, per the repository review-follow-up instructions.

unittests/ram_tests.cpp
```diff
@@ -78,6 +78,129 @@ BOOST_FIXTURE_TEST_CASE(new_account_ram_tests, sysio_system::sysio_system_tester
...
+    };
+
+    // Measure the shortest supported shape, then derive the longest that still fits. Each extra
+    // RPID character costs one byte in each of the two authorities.
+    create_webauthn_account( "wakeyshort"_n, 1 );
+    produce_block();
+    const int64_t at_len_one = resource_manager.get_account_ram_usage( "wakeyshort"_n );
+    BOOST_TEST( at_len_one <= (int64_t)newaccount_ram );
+
+    const int64_t max_rpid_len = 1 + ( (int64_t)newaccount_ram - at_len_one ) / 2;
```
huangminghuang: **[P2] Pin an independent WebAuthn billing value**

`at_len_one` comes from the implementation under test and then determines `max_rpid_len`; an additive error in the new heap/header charge only shifts this derived boundary, so the fit/fail assertions all remain green. The direct `heap_size` test also computes its expectation with the production `shared_string_header_billable_size`, so it does not independently pin that protocol number. The fixed terms in this PR imply `at_len_one == 1032`, `max_rpid_len == 57`, and max-length usage `== 1144`. Please assert at least one independently derived exact value so an accidental billing change fails instead of silently moving the boundary.

huangminghuang (commented):
I am not approving this head because the integrated reference data is stale.

**[P2] Regenerate reference data from current `master`**

Head `c8bf4d8` has merge base `dc8d2cf` and is 144 commits behind `master` `796dd34`. A conflict-free merge selects the current `contracts/sysio.roa/sysio.roa.wasm` (SHA-256 `bc234398…`), while the committed JSON snapshots still embed `569dfaf8…` from the old contract. The successful Linux run does not disprove this: its synthetic merge already contained the same mismatch, and the compatibility tests only establish that `blocks.log` and the snapshots are self-consistent; they do not compare embedded code with the tracked contracts.

The repository regeneration guidance requires reference data to be refreshed whenever a production contract is recompiled. Please update from `master`, regenerate the snapshot/deep-mind fixtures from that integrated tree, update the snapshot-info expectation and PR size attribution, and rerun CI.

Also, the PR body still omits the third commit’s observable `deleteauth` diagnostic change and regression test; please include them when refreshing the description.

unittests/ram_tests.cpp
heifner: Fixed in `e65abfbb29`. You are right, and the criticism is sharper than "not independently pinned": the derived form is insensitive to exactly the error class it was meant to catch.

I worked the numbers from the constants rather than from the code. Outside its keys `newaccount` bills 96 for the `account_object` (32 + 2 indices x 32), 448 for two `permission_object`s (48 `shared_authority` + 40 + 4 indices x 32, aligned to 224 each), and 272 from `initialize_account` (`resource_object` 160 + `resource_pending_object` 112) — 816. Each authority adds a 64-byte `key_weight` element (56 aligned up) plus the heap allocation: an 8-byte `shared_cow_string::impl` header, two `uint32_t`, over 33 point bytes + 1 presence byte + 1 varint length + the RPID. So length 1 is 816 + 2 x (64 + 8 + 36) = 1032 and length 57 is exactly 1144. Your three values reproduce.

Confirming the failure mode: if `shared_string_header_billable_size` were 16 rather than 8, `at_len_one` becomes 1048, the derived `max_rpid_len` becomes 49, and usage at 49 is again exactly 1144 — every assertion stays green. The boundary absorbs any additive error.

Changes:

- `webauthn_account_rpid_bounded_by_gift` pins `at_len_one == 1032` and `max_rpid_len == 57` against constants written out from the derivation, and the max-length account now asserts `== newaccount_ram` rather than `<=`.
- `key_heap_size_covers_every_alternative` no longer expresses either expectation through `shared_string_header_billable_size`. It uses 8 and 96 directly, asserts the BLS serialization is 96, and adds a `static_assert` pinning the constant itself so it cannot cancel out.
- `gifted_account_affords_linkauth` had the same weakness twice over: `link_cost` came from `billable_size_v<permission_link_object>`, so a change there moved both sides of `usage_after - usage_before == link_cost`. It now `static_assert`s that at 144 and pins the fixed-size-key account at 944. `new_account_ram_tests` already asserts 944 for the default K1 account, so pinning here is what extends it to the R1, EM and Ed25519 accounts compared against it.
- The comment claiming the derived boundary made a billing change visible said the opposite of the truth, and is corrected.

huangminghuang (commented):
Re-reviewed current head `e65abfbb29`. The independent WebAuthn billing oracle is fixed, its pinned values are correct, and I found no additional source defect. I still cannot approve because:

**[P2] Regenerate reference data from the integrated tree**

This head is 147 commits behind current `master` (`4148432`). A conflict-free merge selects `contracts/sysio.roa/sysio.roa.wasm` with SHA-256 `bc234398…`, while the committed snapshot retains the old `sysio.roa` code hash `569dfaf8…` from the merge base. Green CI does not catch this because the compatibility tests establish fixture self-consistency but do not compare embedded code with the tracked contract. Please update from `master`, regenerate the snapshot/deep-mind reference data from that integrated tree, update the snapshot-info expectation and size attribution, and rerun CI.

**[P3] Refresh the PR description**

The body still omits the user-visible `deleteauth` diagnostic change and its regression test, as well as the `e65abfbb` follow-up that independently pins the WebAuthn/account/link billing values. Please update it to describe the complete current PR.

heifner:
Both addressed. Merge is `8af5085e01`, regenerated data is `6496feae58`.

## [P2] Reference data regenerated from the integrated tree

Confirmed, including your hashes. The branch was 154 commits behind by the time I merged. The merge is conflict-free and selects `bc23439825ad36e2` for `contracts/sysio.roa/sysio.roa.wasm`, while the committed fixtures embedded the merge base's `569dfaf8cf0e97da` — twice in `snap_v1.json.gz`, with the new hash appearing nowhere. `sysio.roa` changed on master via #584 and #585.

Your account of why CI stays green is right, and worth stating precisely because it is structural rather than an oversight: `libraries/testing` embeds its *own* copy of `sysio.bios` and `sysio.roa` from `${CMAKE_BINARY_DIR}` via `INCBIN`, and the compatibility tests establish agreement between the fixtures and that embedded copy. Both move together, so they stay self-consistent while diverging from the tracked contract. My build directory held exactly that stale pair until the reconfigure restaged it.

Regenerated per the documented procedure — deep-mind first, then snapshots with the stale copies deleted from both the source tree and the build directory, then the `head_block_id` expectation. Afterwards:

- `569dfaf8cf0e97da` appears **0** times in `snap_v1.json.gz`; `bc23439825ad36e2…` appears in its place
- embedded and tracked `sysio.roa` are both `bc23439825ad36e2`
- `head_block_id` is now `0000001dc59bcae1448e5cdd5e637fe8f0833300970eb4c630bba7b2606d08d2`
- `chain_id` and `head_block_num` are unchanged, so this remains a non-regenesis
- the deep-mind log re-derives byte-identically and needed no update, since these fixtures deploy no contract the merge altered

Verified against the integrated tree: `deep_mind_tests`, `snapshot_part2_tests` (8 cases), `ram_tests` (11), `auth_tests` (21), `savanna_misc_tests/verify_block_compatibitity`, and `sysio_util_snapshot_info_test` run through ctest.

## [P3] Description refreshed

Added a section for the `deleteauth` diagnostic and `auth_tests/deleteauth_error_names_deleteauth`, and recorded the `e65abfbb` pins: 944 for a fixed-size-key account, 144 for a `permission_link_object`, and 1032 / 57 for the WebAuthn boundary.

One correction beyond what you listed. The WebAuthn section claimed the test "derives the boundary from a measured shortest account rather than hardcoding it, so a future billing change moves it visibly." That is backwards, and it is the reasoning your earlier finding refuted — deriving is precisely what makes an additive error invisible. Rewritten to say the boundary is measured *and* pinned, keeping the 8-to-16 header example that shows why measuring alone left every assertion green.

I also added a subsection to "Reference data" recording this merge-drift mode, since the note already there covers only the earlier within-branch drift.

huangminghuang (approved):
Re-reviewed current head `6496feae58`. The prior WebAuthn-oracle, integrated reference-data, and PR-description findings are resolved. Source changes and regenerated snapshot artifacts are consistent with current `master`; no actionable findings remain. Approving while the current-head CI matrix continues.

PR #588, opened 2026-08-25
logging: add es_sink — batched Elasticsearch/OpenSearch bulk log forwarding

jglanz:
## What

A new `es_sink` logging sink type that forwards nodeop logs to an Elasticsearch/OpenSearch endpoint via the bulk API, plus an open token-template `layout` option on the `json` formatter that owns the document shape.

### `es_sink` (`fc::es_sink_mt`)
- The logging thread only formats and appends to an in-memory batch — never network I/O. A single `worker_task_queue` worker delivers batches with **size** (`batch_size`), **byte** (`max_batch_bytes`), and **interval** (`flush_interval_ms`) flush triggers.
- Per-document size cap (`max_doc_bytes`, oversized docs dropped + counted), bounded delivery queue (drop-newest, never blocks logging), bounded retry with capped exponential backoff — **429/5xx/connect-class only**, never after a 2xx (partial-success replay would duplicate documents) and never on other 4xx.
- Optional HTTP basic auth (`username`/`password`), positive-signal bulk-response accounting (a non-bulk 2xx is counted failed, never indexed), and a bounded graceful shutdown drain built on gap-free enqueued/completed counters (queue `size()` alone has an idle-looking window between pop and delivery).
- Sink diagnostics are counters + one rate-limited stderr summary — **never fc loggers** (the sink may be attached to `default`; an fc log call from sink internals would recurse into `sink_it_`).
- SIGHUP reload safe: destruction (which runs on the app thread during `handle_sighup()` fan-out) is bounded by `shutdown_flush_timeout_ms` (default 500 ms) + cancellation.

### `json` formatter `layout` templates (`fc::log::json_layout`)
- `layout` is an open token-template string: literal JSON authored freely, with `${token[:modifier]}` placeholders (`timestamp[:iso8601|epoch_millis]`, `level[:upper|lower]`, `message`, `logger`, `file`, `line`, `func`, `thread`, `extra_flat`, `extra_object`; `$${` escapes a literal `${`).
- Empty/absent `layout` renders the historical fc JSONL shape **byte-identically** (pinned by tests). `fc::log::es_default_layout` ships the ES document shape (`@timestamp` epoch millis, uppercase `level`, `message`, `category`, `sourceLocation`, `data.thread`, `extra_fields` flattened top-level) and is the es_sink's default formatter.
- An es_sink's final formatter (default or explicit override) is sample-render-validated at configure time: a template that doesn't render one-line JSON fails `configure_logging` loudly instead of silently 400-ing every bulk request. Custom JSON-valid shapes are accepted.

### Supporting changes
- Shared JSON escaping extracted to `fc::log::detail` (`json_escape.hpp`), consumed by the formatter, the layout engine, and the sink's action-line builder.
- New multi-request `fc-test/capture_http_server.hpp` fixture (records every request, scripted status/body/delay responses).
- `programs/nodeop/logging_elasticsearch.json` — `logging_default.json` plus the `es` sink (placeholder `url`/`index`) attached to every `stderr_color` logger.

## Testing
- **`es_sink_tests`** (25 cases): flush triggers, oversized-doc drop, basic auth, queue-full non-blocking drop, retry semantics pinned (`max_retries` = additional attempts), 429-retries/4xx-terminal, `errors:true` partial accounting, non-bulk 2xx, shutdown tail drain (incl. slow-response in-flight case), unreachable endpoint, 8-thread concurrency, two coexisting sinks, and every `configure_logging` acceptance/rejection path. 5/5 consecutive green runs.
- **`json_formatter_tests`** +13 layout cases incl. byte-identity regression for the default shape.
- Full `test_fc`, `unit_test -- --sys-vm`, and `plugin_test` all green; full `all` build clean.
- **Live cluster verification**: 6-node `wire-cluster-tool` cluster with this config shipped 2,600+ schema-conformant documents to a live OpenSearch index (correct `@timestamp`/`level`/`category`/`sourceLocation`/identity fields, exactly one principal bucket per node, zero sink warnings), stayed thread-flat across a double-SIGHUP reload with an unchanged doc rate, and tore down cleanly.

### DIFF
--- a/libraries/libfc/include/fc/log/logger_config.hpp
+++ b/libraries/libfc/include/fc/log/logger_config.hpp
@@ -31,6 +31,13 @@ namespace fc {
          // (string/number/bool/null) coerced to string at load time. Nested
          // objects or arrays throw fc::bad_cast_exception on configure_logging.
          fc::variant_object extra_fields;
+         // layout is a fc::log::json_layout token template controlling the document
+         // shape (see fc/log/json_layout.hpp). Empty/absent renders the historical
+         // fc JSONL shape (fc::log::default_layout); fc::log::es_default_layout is
+         // the shipped Elasticsearch/OpenSearch document template. A malformed
+         // template (unknown token/modifier, unterminated placeholder) fails
+         // configure_logging.
+         std::string layout;
       };
    } // namespace format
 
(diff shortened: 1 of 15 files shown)


libraries/libfc/src/log/es_sink.cpp
```diff
@@ -0,0 +1,404 @@
...
+      if (!retryable || attempt >= _cfg.max_retries) {
+         _failed_batches.fetch_add(1, std::memory_order_relaxed);
+         return;
+      }
+      // Capped exponential backoff, interruptible by shutdown's cancel request. The
+      // predicate deliberately keys on _cancel_requested (set at drain-timeout), NOT
+      // _shutting_down (set at teardown start) -- otherwise retries would spin at zero
+      // backoff for the whole graceful-drain window.
+      const auto backoff =
+         std::min<std::chrono::milliseconds>(std::chrono::milliseconds(_cfg.retry_backoff_ms) * (1u << attempt),
```
heifner: `1u << attempt` is UB once `attempt` reaches the width of `unsigned`, which `max_retries >= 33` allows. `std::min(..., es_max_backoff)` caps the result, not the shift, so it doesn't prevent it.

`1u << std::min(attempt, 16u)` fixes it with no behavior change: backoff is already capped at 2000 ms, so the exponent stops mattering past about four attempts.

libraries/libfc/include/fc/log/logger_config.hpp
```diff
@@ -86,6 +93,54 @@ namespace fc {
...
+      /// ADDITIONAL delivery attempts after the first (total attempts = max_retries + 1).
+      inline constexpr uint32_t default_es_max_retries = 2;
+      /// Initial retry backoff; doubles per attempt, capped inside the sink.
+      inline constexpr uint32_t default_es_retry_backoff_ms = 250;
+      inline constexpr uint32_t default_es_connect_timeout_ms = 5000;
+      inline constexpr uint32_t default_es_request_timeout_ms = 10000;
+      /// App-thread stall budget: the sink destructor runs on the app/SIGHUP thread
+      /// during plugin handle_sighup() fan-out -- this bounds that stall, NOT a
+      /// background wait.
+      inline constexpr uint32_t default_es_shutdown_flush_timeout_ms = 500;
```
heifner: This lands on the block-production thread, not just "the app thread".

SIGHUP posts at `priority::medium` through `post_cb`, which reaches `priority_queue_executor`'s two-argument `post` and routes to `exec_queue::read_write`, the same single-threaded queue block production uses at `priority::high`. `execute_highest` runs handlers to completion, so priority orders queued handlers but doesn't interrupt a running one: N ms here delays the next production handler by up to N ms. `exec_window_` stays `write` throughout, so the read-only pool is held off too.

The destructor runs inside that handler, since the sink's refcount reaches zero as the `handle_sighup()` fan-out re-points the statics. Budget is this 500 ms drain plus the `_queue->stop()` join with the worker mid-`perform()` (~50 ms cancel-poll granularity), so roughly 550 ms against `config::block_interval_ms` of 500.

It only costs anything when a delivery is in flight and slow, but `connect_timeout_ms` is 5 s with `max_retries` 2, so against a dead endpoint the drain always times out, and logrotate SIGHUPs on every rotation. Suggest ~100 ms, or detaching the drain so destruction returns immediately.

heifner (changes requested):
Two inline. The backoff shift is UB once `max_retries` reaches 33, and `shutdown_flush_timeout_ms` defaults to 500 ms on the same single-threaded queue as block production at a 500 ms block interval.

heifner (approved):

PR #589, opened 2026-08-25
fix: re-evaluate operator eligibility on collateral exit (WIRE-351)

huangminghuang:
## Summary

Fixes WIRE-351 by re-evaluating operator eligibility as soon as collateral withdrawal reservations are created or canceled. Producers and batch operators therefore stop receiving active work while their available collateral is below the configured minimum, rather than remaining active until the withdrawal wait period is flushed.

## Scope

This PR is limited to:

- contracts/sysio.opreg/src/sysio.opreg.cpp
- contracts/tests/sysio.opreg_tests.cpp
- regenerated contracts/sysio.opreg/sysio.opreg.wasm

## Changes

- re-evaluate eligibility after successful withdraw and withdrawinle reservations
- re-evaluate eligibility after cancelwtdw removes a reservation
- erase matured withdrawal reservations before the flush-time re-evaluation, preventing the amount from being subtracted twice
- preserve SLASHED and TERMINATED terminal states against stale eligibility callbacks
- cover direct and inbound withdrawal, cancellation, rejection, flush ordering, and terminal-state regressions
- synchronize the tracked sysio.opreg WASM with the tested source build

## Validation

Validated at e9d3647e56385cc87b6a16d2ee2159b3e8d76fdc:

- hygiene, canonical CDT SDK freshness, and required submodules passed
- Release configure and the repository-required default full build passed
- all 662 system-contract test cases passed with --sys-vm
- the artifact-copy CI configuration and default full build passed
- the artifact-copy contracts_unit_test CTest passed
- tracked and source-built sysio.opreg.wasm are byte-identical (sha256 acef9d0dd1a81eb52be0811c9346711725888c6d6e4fa6098fc43e8feff03d40)
- current-head GitHub CI is pending after this follow-up push

## Scope boundaries

No Solana client, version-label, protocol, schema, or external-repository changes remain in this PR. The separate Solana and cluster-tool proposals were closed as outside WIRE-351 and are tracked by SOL-406, SOL-407, and SOL-408.

## Migration and risk

There is no table, storage, action-interface, or ABI migration. Deployment must use the tracked regenerated sysio.opreg.wasm together with the source change.


huangminghuang (commented):
Re-reviewed the complete current diff at 020f04ad160abcc6bd91c6f06f77520e447c0bf3 against WIRE-351 and the current master base. No blocking correctness, architecture, compliance, or test-coverage issues found. All current-head CI checks are green, and the base-only changes do not overlap this PR. GitHub did not permit a formal approval because the authenticated account is the PR author.

heifner (commented):
Reviewed at `020f04ad1` against WIRE-351 / CertiK WNS-25.

**The ticket's recommendation is satisfied exactly.** WNS-25 asked for `reevaluate_eligibility()` at the end of `withdraw()`, `withdrawinle()`, and `cancelwtdw()`; all three are present, and the second commit explicitly restricts the change to `sysio.opreg`. Three files, +256/-13, no drift.

Two things I verified as correct beyond the ask:

- The `flushwtdw` reorder (`queue.erase(wkey)` before `reevaluate_eligibility`) is a real bug fix — the stale queue row was still visible to `sum_pending_withdraws`, so `available_inline` subtracted the matured amount twice. `it` is advanced before the erase, so the loop stays safe.
- The `has_terminal_status` guards in `reevaluate_eligibility` and `process_eligibility_change` are sound, and `meets_role_min` still early-returns on `is_bootstrapped`, so the genesis seed set cannot be de-activated by this path.

**One thing I checked and am deliberately NOT raising**, so it doesn't get re-litigated: reservation-time de-activation shrinks the consensus electorate, because `evalcons` sizes `group_size` from `eligible_batch_operators()`, which intersects the seated group with live opreg ACTIVE status. That is not a stall — the threshold adapts — and a depot-side floor is the explicitly accepted trade in `opp-consensus.md` (CertiK WNS-08 / #573), since flooring re-creates an unrecoverable stall and the outposts carry no floor either.

The two findings below both come from the same shift the fix introduces: de-activation now lands at **reservation** time rather than at **flush** time. That is the correct behaviour per the ticket — these are the two places the rest of the system has not caught up with it.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -974,6 +987,9 @@ void opreg::withdraw(name account, uint64_t amount) {
    auto action = build_withdraw_request_action(account, kWireChainCode, kWireTokenCode, amount,
                                                result.request_id);
    append_action_log(ops, op_pk, action, result.success, std::move(result.error_message));
+   if (result.success) {
+      reevaluate_eligibility(ops, op_pk, get_self(), account);
```
heifner: **MEDIUM — a routine partial withdrawal now silently arms a delayed termination.**

`sysio.epoch::advance` records a delivery observation for **every** member of the expiring group regardless of opreg status, and its own comment says so: *"`recorddel` remains an audit record even for a newly slashed operator, while `termcheck` safely skips non-ACTIVE rows."* `termcheck` does skip them (`op.status != ACTIVE` returns early) — but the `dellog` rows are still written, and the rolling window retains them.

So for a non-bootstrapped batch operator:

1. It requests a partial withdrawal that dips `available_inline` below `req_batchop_collat`, and this change flips it to `UNKNOWN` in the same transaction.
2. It can no longer `deliver` — `sysio.msgch::deliver` hard-checks `op.status == ACTIVE`.
3. Each epoch it remains seated, `recorddel(delivered=false)` accrues.
4. It tops up collateral, or the flush completes and it re-deposits, and flips back to `ACTIVE`.
5. The very next `termcheck` walks the window, sees `worst_consecutive > terminate_max_consecutive_misses`, and calls `terminate_inline`.

The operator is terminated for having used the withdrawal path. Before this change it stayed `ACTIVE` through the wait window and kept delivering, so no misses accrued.

Scope, for the record: `termcheck` returns early for `is_bootstrapped` and for any non-BATCH type, so this reaches non-bootstrapped batch operators only — producers are not terminable this way.

Either skip `recorddel` for members that are not `ACTIVE`, or have `termcheck` ignore misses accrued while the operator was de-activated. The first is simpler; the second preserves the audit trail the comment is protecting.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -1300,11 +1328,14 @@ void opreg::flushwtdw(uint32_t current_epoch) {
...
       append_action_log(ops, op_pk, remit_action, true, "");
 
+      // Remove the matured reservation before eligibility is recomputed. The
+      // balance debit already accounts for this withdrawal; leaving the queue
+      // row visible would make available_inline subtract the same amount twice.
+      queue.erase(wkey);
+
       // Re-check eligibility — this withdraw may have dropped the operator
       // below the role minimum.
       reevaluate_eligibility(ops, op_pk, get_self(), row.account);
```
heifner: **LOW — the `flushwtdw` shortfall branch erases the reservation without re-evaluating, which is the exact staleness this PR fixes here.**

This success path now does the right thing: erase the matured row, then recompute. But the `"insufficient balance at flush (rollup mismatch)"` branch about twenty lines above (`:1305-1311`) does `queue.erase(wkey); continue;` with no `reevaluate_eligibility`.

That leaves the operator stranded. Its reservation is gone, so `available_inline` immediately rises back above `req_*_collat` — but nothing recomputes eligibility, so it sits in `UNKNOWN` until some unrelated `deposit` / `withdraw` / `cancelwtdw` happens to trigger a re-check. With de-activation now landing at reservation time, that stranded window is reachable in ordinary operation rather than only after a rollup bug.

The `SLASHED` / `TERMINATED` branches above it are correctly exempt (terminal, and `has_terminal_status` would no-op anyway), and the `!ops.contains` branch is moot. This one wants the same call the success path makes.

contracts/sysio.opreg/src/sysio.opreg.cpp
huangminghuang: Addressed in 42850d7. `recorddel` now ignores observations for every non-`ACTIVE`, non-terminal operator, while preserving audit records for `SLASHED` and `TERMINATED` states. The new `recorddel_ignores_withdrawal_ineligibility` regression verifies that two misses during withdrawal-induced `UNKNOWN` status create no delivery-log rows and cannot terminate the operator after cancellation/reactivation.

contracts/sysio.opreg/src/sysio.opreg.cpp
huangminghuang: Addressed in 42850d7. The defensive rollup-shortfall branch now erases the stale reservation and immediately calls `reevaluate_eligibility`, matching the other reservation-removal paths.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -1582,6 +1620,21 @@ void opreg::claimremit(name account) {
...
+   // those operators permanently.
+   operators_t ops(get_self());
+   auto op_pk = operator_key{account.value};
+   if (ops.contains(op_pk)) {
+      auto status = ops.get(op_pk).status;
+      if (status != OperatorStatus::OPERATOR_STATUS_ACTIVE &&
+          !has_terminal_status(status)) {
+         return;
+      }
+   }
```
heifner: **MEDIUM — this guard stops the parked interval from *contributing* misses, but not from *welding together* the miss runs on either side of it.**

`termcheck` counts consecutive **rows**, not consecutive **epochs**, and resets the run only on a materialized `delivered == true` row:

```cpp
// :1696-1706 — unchanged by this PR
for (auto it = idx.lower_bound(lower_key); it != idx.end() && it->by_account_ts() <= upper_key; ++it) {
   if (it->account != account) break;
   total_in_window++;
   if (!it->delivered) { total_misses++; consecutive_misses++; ... }
   else                { consecutive_misses = 0; }
}
```

Absence is invisible to that loop. Writing *nothing* during the de-activated interval is therefore not equivalent to writing deliveries — the pre-withdrawal run survives the gap intact and the first post-reactivation miss continues it.

`sysio.epoch::advance` emits one `recorddel` per **(active outpost × expiring-group member)** (`sysio.epoch.cpp:590-603`; header comment at `sysio.epoch.hpp:147-149`), so with the launch outpost set `{ETH, SOL}` a single fully-missed duty epoch already writes **two** miss rows, against a production `terminate_max_consecutive_misses = 5` (`sysio.opreg.hpp:89`).

All of the following inside one 24h `terminate_window_ms`:

| Step | Rows added | `worst_consecutive` |
|---|---|---|
| 2 fully-missed duty epochs while ACTIVE | 4 misses | 4 — survives |
| `withdraw` / `withdrawinle` drops below `req_batchop_collat` → UNKNOWN | — | — |
| parked; every `recorddel` dropped by this guard | 0 | 4 (unchanged) |
| `cancelwtdw` or top-up → ACTIVE, re-seated | — | — |
| **1** fully-missed duty epoch | 2 misses | **6 > 5 → `terminate_inline`** |

On master the operator stayed ACTIVE through the wait window, so any delivery in it reset the run to 0. The PR description says "withdrawal-induced downtime cannot become termination misses after reactivation" — the downtime adds no misses of its own, but it no longer clears the run either, so the operator is still terminated for having used the withdrawal path, just with a shorter fuse.

`recorddel_ignores_withdrawal_ineligibility` only exercises a clean pre-withdrawal history, so this case isn't covered.

### Suggested fix — floor the scan at the last reactivation

`process_eligibility_change` already stamps `o.available_at = now` on every ineligible→eligible transition (`:1367-1368`), and the only other writer is registration (`:363`). In `termcheck`:

```cpp
uint64_t scan_open  = std::max(window_open, op.available_at);
uint128_t lower_key = (static_cast<uint128_t>(account.value) << 64) | scan_open;
```

No schema change, no new field, and a no-op for an operator that never demotes (its `available_at` is the original activation, far below `window_open`). For one returning from a park it starts clean — which is the semantics this guard was reaching for, and it gets there without dropping the audit rows at all.

Worth adding the regression the current test is missing: miss → `termcheck` (survives) → `withdrawinle` → dropped `recorddel`s → `cancelwtdw` → miss → `termcheck` must leave the operator ACTIVE. `activate_batch_operator(op, /*max_consec_misses=*/1, ...)` makes that about a dozen lines.

heifner:
Re-reviewed `42850d7`. Both earlier threads: the `flushwtdw` rollup-shortfall one is correctly fixed; the `recorddel` one is only half addressed — details in the thread on `sysio.opreg.cpp`.

The rest of the change verified clean, and I'd rather record these three than block on them.

### Verified correct

- All four re-evaluation sites match WNS-25's recommendation — `withdraw:990-992`, `withdrawinle:1019-1021`, `cancelwtdw:1035-1039`, plus the shortfall branch `:1310`. Each gap was real on master.
- The `flushwtdw` erase-before-re-evaluate reorder (`:1335` ahead of `:1339`) is the right fix. `sum_pending_withdraws` is a live `byaccount` scan of `wtdwqueue`, so a row still present when `meets_role_min` runs gets subtracted a second time on top of the balance debit. `flushwtdw_rechecks_eligibility_after_erase` genuinely discriminates — balance 2, min 1, withdraw 1 flips to UNKNOWN under the old order.
- Both terminal-state guards are load-bearing rather than belt-and-braces: `reevaluate_eligibility` dispatches an **inline action** (`:1073-1077`), so status can change between send and execution. Defending at the sink (`:1362`) as well as at the caller (`:1061`) is correct.
- `has_terminal_status` is a faithful extraction of the existing `available_inline` guard.
- No `dellog` id gap — the early return precedes `next_dellog_id()`.
- ABI unchanged; the `.wasm` moves alongside a real source change.

### LOW — the dropped observations also shrink the percent rail's denominator

```cpp
// :1709-1710
bool exceeds_percent = total_in_window > 0 &&
                       (total_misses * 100u / total_in_window) > cfg.terminate_max_pct_misses_24h;
```

`total_in_window` is *rows present*, not seated epochs. Dropping observations drains the sample while the operator is parked, and `prune_dellog` ages out the pre-park rows, so an operator can return to a near-empty window where a single missed duty epoch reads as 100% against a production threshold of 5%.

The sharp edge is pre-existing — `termcheck` already terminates on one miss row even at the most permissive 99% setting — but this PR adds a routine, operator-triggered way to reach an empty window. The `available_at` floor suggested in the thread makes a returning operator behave like a freshly activated one rather than fixing the small-sample problem; a real fix is a minimum-sample guard on the percent rail, which is its own ticket.

### Observation — parking is now a clean way to shed miss history

`termcheck` early-returns on non-ACTIVE (`:1664`) and `recorddel` now writes nothing, so a batch operator approaching the consecutive threshold can queue a withdrawal large enough to drop below `req_batchop_collat`, sit out the termination window, then `cancelwtdw` and come back with a pruned log. It costs nothing: `try_enqueue_withdraw` writes only a queue row and never debits a balance, and `cancelwtdw` is free. Set against permanent termination plus collateral remit, parking is strictly the better move for a bad operator.

This is inherent to "de-activate at reservation time and don't punish the downtime" rather than to any line of this diff. Flagging it because WNS-25's remediation now interacts with the termination machinery in a way the audit didn't scope — probably belongs on the ticket.

### Scope note — lock paths still move `available_inline` without re-evaluating

`available_inline` is `balance − sum_locks_inline − sum_pending_withdraws`. After this PR every *withdraw-queue* mutation re-evaluates, but no *lock* mutation does: `releaselock` explicitly early-returns for a healthy operator (`:1464-1469`, "uwrit::release just erases the lock row; opreg has no work") and lock acquisition lives in `sysio.uwrit`. Same staleness class as WNS-25, entirely pre-existing, outside this ticket (which names only the three withdraw actions), and confined to underwriters, whom `termcheck` doesn't terminate (`:1677`). Follow-up ticket, not this PR.

contracts/sysio.opreg/src/sysio.opreg.cpp
huangminghuang: Addressed in `6009e89a4c`. `termcheck` now floors its delivery-log scan at `max(window_open, op.available_at)`, so rows from an earlier ACTIVE interval cannot join a consecutive-miss run after withdrawal cancellation reactivates the operator. The regression records a miss before parking, verifies observations are ignored while `UNKNOWN`, reactivates the operator, then records a post-reactivation miss and proves `termcheck` keeps it ACTIVE. The exact committed head passed the full 663-case source suite, the 663-case artifact-copy suite, and the monitored `flow-operator-collateral-deposit` run.

contracts/sysio.opreg/src/sysio.opreg.cpp
```diff
@@ -1631,9 +1685,14 @@ void opreg::termcheck(name account) {
...
 
+   // A transition back to ACTIVE starts a new duty interval. Rows from an
+   // earlier active interval must not join a consecutive-miss run across time
+   // when the operator was ineligible and could not deliver.
+   uint64_t scan_open = std::max(window_open, op.available_at);
+
    dellog_t log(get_self());
    auto idx = log.get_index<"byaccountts"_n>();
-   uint128_t lower_key = (static_cast<uint128_t>(account.value) << 64) | window_open;
+   uint128_t lower_key = (static_cast<uint128_t>(account.value) << 64) | scan_open;
```
heifner: **MEDIUM — the floor narrows *both* rails, so every reactivation now empties the percent-rail sample. This lands harder than the bug it closes.**

Thanks for taking the `available_at` floor — the consecutive rail is genuinely fixed. But my suggestion was imprecise about *where* to apply it, and as written `scan_open` feeds `lower_key` (`:1695`), so the single loop accumulates `total_in_window` and `total_misses` from the narrowed scan too (`:1704-1712`). A reactivation therefore doesn't just reset the consecutive run — it resets the miss-rate sample to empty.

At the production `terminate_max_pct_misses_24h = 5` (`sysio.opreg.hpp:90`), with `sysio.epoch::advance` emitting one `recorddel` per **(active outpost × expiring-group member)** — two rows per duty epoch on the launch set `{ETH, SOL}`:

| Post-reactivation state | `total_misses * 100 / total_in_window` | outcome |
|---|---|---|
| first duty epoch fully missed | `2*100/2` = 100 | **terminated** |
| first duty epoch missed on one outpost only | `1*100/2` = 50 | **terminated** |
| 1 miss among 16 rows | `100/16` = 6 | **terminated** |
| 1 miss among 17 rows | `100/17` = 5 | survives |

So a returning operator must accumulate **~17 clean observations (≈9 duty epochs)** before it can afford a single miss.

Before `6009e89` the pre-park rows stayed inside the 24 h sample, so an operator with a healthy history returned with a large denominator — only a park longer than `terminate_window_ms` drained it. Now every reactivation drains it, and this fires on the **first** missed observation regardless of prior history, where the original finding needed a pre-existing 4-miss run plus one more. It is the same failure WIRE-351 is about — a routine partial withdrawal ending in termination — relocated from the consecutive rail to the percent rail.

`recorddel_ignores_withdrawal_ineligibility` can't see it: `activate_batch_operator` defaults to `kMaxAcceptedPctMisses24h = 99` and the test appends a delivered row, so the rate lands at 50 % and clears a 99 % ceiling. No test exercises the percent rail at the production 5 % — `termcheck_terminates_at_duty_rotation_cadence` explicitly notes its delivered record "keeps the percent rail below its ceiling for the rest of the run."

### Suggested correction — floor the run, not the sample

Keep `lower_key` on `window_open` and reset the run at the boundary inside the loop:

```cpp
   total_in_window++;
   if (!it->delivered) total_misses++;

   // A transition back to ACTIVE starts a new duty interval. Rows from an
   // earlier interval still count toward the miss RATE, but must never join a
   // consecutive run spanning time the operator could not deliver.
   if (it->ts_ms < op.available_at) { consecutive_misses = 0; continue; }

   if (!it->delivered) {
      consecutive_misses++;
      if (consecutive_misses > worst_consecutive) worst_consecutive = consecutive_misses;
   } else {
      consecutive_misses = 0;
   }
```

One `uint64` comparison per row, and `scan_open` goes away.

The split is the point: the consecutive rail measures an unbroken run of failures *while the operator was able to deliver*, and a reactivation legitimately breaks it. The percent rail measures a rate over a time window, and needs the whole window to mean anything — a 5 % threshold evaluated against a 2-row sample is not a rate.

For the regression, pin a case at the production `kDefaultMaxPctMisses24h = 5` rather than the 99 % ceiling; that is the only setting under which this class is observable.

One minor note while you are in here: `ts_ms` is second-granular (`current_time_ms()` is `sec_since_epoch() * 1000`), so a pre-park row written in the same second as the reactivation still compares `>= available_at` and joins the run either way. Your `produce_block(fc::seconds(2))` and the `pre_withdrawal_miss_ts < reactivated_at` assertion already guard the test against it, and in production duty epochs are seconds apart — noting it only so the boundary is not mistaken for exact.

huangminghuang: Addressed in 194d0759b4. termcheck now scans from the rolling window boundary so pre-reactivation rows remain in the percentage sample; rows before available_at only reset the consecutive-miss sequence. Added a production-threshold regression proving 19 healthy observations plus the first post-reactivation miss remains exactly 5% rather than becoming a one-row 100% sample. Both exact-head source and artifact-copy contract suites passed all 664 cases.

heifner (approved):
Re-reviewed at `194d0759b4`. **All four findings from the three prior rounds are addressed, I found nothing new, and CI is green on exactly this head** (run `33522743676`, `head_sha` verified — asan / asserton / gcc / ubsan / ubuntu24 + package verification all pass). Approving.

The last commit applies the split-rail correction exactly as specified:

```cpp
total_in_window++;
if (!it->delivered) total_misses++;          // percent rail keeps the whole window

if (it->ts_ms < op.available_at) {           // consecutive rail restarts at reactivation
   consecutive_misses = 0;
   continue;
}
```

That is the right division: the percent rail measures a rate over the rolling window and needs the whole window to mean anything; the consecutive rail measures an unbroken run of failures *while the operator was able to deliver*, and a reactivation legitimately breaks it. Dropping `<algorithm>` with `std::max` is safe — no other `std::max`/`min`/`sort` in the file.

### What I verified beyond the delta

- **The `available_at` floor is reliable.** `process_eligibility_change` (`sysio.opreg.cpp:1367`) is the only non-bootstrap writer of `OPERATOR_STATUS_ACTIVE`, and it stamps `available_at` inside the same `modify`. There is no second path to ACTIVE that could leave the floor stale. `available_at` also has no other consumer — opreg and its tests only, and the TS side is a generated field with no logic behind it — so making it load-bearing for termination introduces no coupling.
- **Both new regressions genuinely discriminate**, which is the part worth checking on a test-heavy fix:
  - `recorddel_ignores_withdrawal_ineligibility` — without the `ts_ms < available_at` guard the pre-park miss welds onto the post-reactivation miss, `worst_consecutive` reaches 2 against `max_consec_misses = 1`, and the operator terminates. The `get_dellog_entry(kThirdDeliveryLogId).is_null()` assertion is a nice touch: it pins that the parked observations were actually dropped rather than merely ignored.
  - `termcheck_preserves_percent_sample_across_withdrawal_ineligibility` — under the previous `scan_open` floor this is a one-row, 100% sample and terminates; with the fix it is 1 miss in 20 rows = exactly 5%, at the production ceiling. Pinning it at `kDefaultMaxPctMisses24h` rather than the 99% test ceiling is what makes it observable at all, and the explicit `healthy_history_end < reactivated_at` assertion keeps the boundary from being accidental.
- **`flushwtdw`** — `ops.contains` guard first, SLASHED/TERMINATED exempt, the shortfall branch now erases *and* re-evaluates, and the success path erases before re-evaluating. `flushwtdw_rechecks_eligibility_after_erase` is discriminating for the reorder: without it `available_inline` double-subtracts and the operator flips to UNKNOWN.
- **The terminal-state invariant is guarded at both ends** — in `reevaluate_eligibility` before the fan-out and in `process_eligibility_change` at the sink. Not redundant: the fan-out is an inline action, so the two guards run at different points in the transaction.
- **Artifact hygiene** — SHA-256 of the committed `sysio.opreg.wasm` is `51be26956d3062e1bb31302ec3e9b418980ff11ad70cebcab572f816d568843a`, matching the PR body, and the wasm is paired with a contract-source change in every commit that touched it (`e9d3647e` is the scope revert and touches no contract source). ABI untouched, net diff is three files.

One thing worth recording: `sysio.opreg.hpp:58-60` already described the 2-epoch wait as *"long enough that an operator who would drop below the role minimum is demoted before the funds physically leave."* That was intent the code did not actually deliver until this change — the header is now true.

### Two things I checked and am deliberately NOT raising

Noting them so they don't get re-litigated in a later round.

**Cold percent-rail sample.** `termcheck`'s percent rail has no minimum sample size (`total_in_window > 0` only), so at the production 5% ceiling a single miss in a ≤16-row window terminates. This is pre-existing and hits fresh registrations and roster-benched operators identically — `sysio.opreg.hpp:118-131` already acknowledges that *"operators benched by a surplus roster accrue no records while benched, so no finite window observes them."* The withdrawal path reaches it only after a **>24h voluntary absence**: `WITHDRAW_WAIT_EPOCHS = 2`, so a real round-trip is two epochs, far inside `terminate_window_ms`, which this PR now preserves exactly. A minimum-sample threshold would need a new `op_config` field, i.e. the ABI change this PR explicitly excludes. Not a finding here — worth its own ticket if we want the guard.

**Self-parking as miss-ladder evasion.** An operator could dip below the minimum to stop accruing misses. It gains nothing: below the minimum it is not ACTIVE, so `advance` will not seat it in a new tail group and it holds no roster slot. The only exposure is the remaining already-seated window, which is the electorate-shrink trade already accepted in `opp-consensus.md` (WNS-08 / #573).

PR #583, opened 2026-08-22
Docs: ROA resource model overview

heifner:
## Summary

Adds `docs/roa-overview.md`, a user-facing page on Resource Owner Allocation. Written for developers arriving from EOS/Antelope, and intended to be handed out directly — the immediate use is answering community questions about what a transaction actually costs on Wire.

Documentation only; no code changes.

## What it covers

- **The payer model.** `action::payer()` returns `authorization[0].actor` when that entry carries the reserved `sysio.payer` permission, and the called contract otherwise. Since ordinary transactions do not name a payer, the contract is billed and the signer is neither charged nor limit-checked. Explicit self-pay is covered as the exception, including the requirement that the payer sit at index 0, carry a real permission on the same action, and sign.
- **What an unprovisioned account or contract looks like.** New accounts start at 0 CPU / 0 NET with 1,144 bytes of RAM. Those zeros are never consulted for a user. For a contract they are fatal *under default billing* — the contract is the payer, so an ordinary call fails rather than falling through to any other billing path. They are not consulted at all when an action names an explicit `sysio.payer`, so a provisioned caller or relayer can drive a contract that has no policy of its own.
- **Policies.** Fields and semantics, who may issue (any registered node owner; there is no tier gate), tier budgets and caps, the four policy actions, and multi-issuer stacking, including the frozen `bytes_per_unit` and the RAM-reclaim interaction between issuers.
- **Throughput.** The proportional-share formula, and specifically that its denominator is total ROA-allocated weight rather than staked tokens.
- **Spam control.** Two layers: the contract's own CPU/NET share as an always-on consensus cap, and `nodeop`'s subjective billing metering the signer. The defaults described are the post-#591 ones — billing on, the contract payer excluded.
- **Scenarios** for a token transfer, contract deployment sizing, trading, CPU exhaustion, onboarding, and self-pay.
- **Comparison** to EOS staking, REX, and PowerUp.

## Verification

The behavioural claims were checked against the implementation with throwaway tests run against a full-system tester, then removed. The transcript blocks in the doc reproduce that output:

- Unprovisioned contract called by a funded user → `account payloadless net usage is too high: 132 > 0`
- Funding only the contract → same call succeeds; contract CPU `0 → 2001`µs, caller's CPU used stays `0`
- Explicit self-pay with no allocation → `account alice net usage is too high: 148 > 0`
- Token transfer RAM → sender `1068 → 1068`, receiver `1068 → 1068` even for a newly created balance row, `sysio` `411951 → 412239`
[...]

PR #590, opened 2026-08-26
chain: discover payer keys by the paired permission, not always active

heifner:
## The mismatch

`authorization_manager` has two places that reason about an explicit `sysio.payer` entry, and they disagreed.

**Consensus** (`check_authorization`) pairs the marker with **any** real permission the payer declares on the same action, then satisfies that entry:

```cpp
if (auth.actor == payer && auth.permission != config::sysio_payer_name) {
   foundPayer = true;
   break;
}
```

**Discovery** (`get_required_keys`) hard-coded `active`:

```cpp
if (declared_auth.permission == config::sysio_payer_name) {
   auto active_auth = permission_level{declared_auth.actor, config::active_name};
   SYS_ASSERT( checker.satisfied(active_auth), unsatisfied_authorization, … );
}
```

So a self-pay transaction paired under `owner`, or under a custom permission linked to the action, is **accepted by consensus but rejected by discovery**. Since `/v1/chain/get_required_keys` is what clio, kiod and every wallet call to learn which keys to sign with, such a transaction could not be signed through the standard path even though the chain would have taken it.

It stayed hidden because a freshly created account usually carries the same key on `owner` and `active`. It bites the accounts that separated them — which is the case the doc for this feature actively recommends.

## The fix

The marker needs no check of its own. It is virtual: no `permission_object` backs it and it holds no keys. Consensus already requires it to be paired with a real permission, and **that entry is checked on its own iteration of the same loop** — so skipping the marker makes discovery agree with consensus by construction, rather than by keeping a second copy of the pairing rule in sync.

In the common case the reported key set is unchanged. Where it differs, it was previously wrong.

**Not a consensus path.** The only callers are the read-only RPC and `producer_plugin` signing its own `votesnaphash` transaction, so no protocol feature applies.

## Test coverage

`get_required_keys_explicit_payer_tests` in `unittests/api_tests.cpp`, on `payloadless::doit` — chosen because it takes no arguments and asserts nothing about its authorization, so the authorization path is the only thing that can reject these transactions.

For each of the three pairings (`active`, `owner`, and a linked custom permission) it asserts **both** halves: the key set discovery reports, **and** that a transaction signed with exactly those keys is accepted. Agreement between the two is the property at issue — either half alone passed before the fix.

It also pins:

- discovery stays **strict** — candidates that cannot satisfy the paired permission still throw, including the neighbouring permissions in the hierarchy (`active` satisfies neither its parent `owner` nor its child `custom`), and an empty candidate set;
- the marker is attributed **no key**, via a two-action transaction pairing the same payer under two different permissions, whose result is exactly the union of those two.

## Validation

- **Guard verified**: reinstating the `@active` check fails the `owner` and custom cases (2 of the discovery assertions, plus the strictness case whose message differs). The test fails without the fix and passes with it.
- **Full suite**: `unit_test --sys-vm` — **1516 test cases, `*** No errors detected`**.

## Note for reviewers

`create_account` already gives `owner` and `active` distinct keys, so the bug was reachable in the default tester and simply never exercised.

This blocks #583 (the ROA overview doc), which documents self-pay under `owner`/custom permissions as working end to end. That is true at consensus but was not discoverable before this change, so this should land first.

### DIFF
--- a/libraries/chain/authorization_manager.cpp
+++ b/libraries/chain/authorization_manager.cpp
@@ -652,16 +652,24 @@ namespace sysio { namespace chain {
          // duplicated rather than shared with the consensus validators -- this function is not on
          // the apply path, and refactoring the consensus copies to serve it would put a
          // non-consensus caller in a position to change consensus behaviour.
+         // Marker presence is tracked separately from the actor rather than using an empty
+         // account_name as the sentinel. An empty actor is valid JSON and decodes to
+         // account_name{}, so a `{"": sysio.payer}` entry would leave the sentinel looking unset
+         // and skip the pairing check below. Consensus is not exposed to that -- it rejects the
+         // empty actor earlier, in validate_referenced_accounts ("action's paying actor '' does
+         // not exist") -- but this function never runs that pass.
+         bool         has_payer = false;
          account_name payer;
          for (size_t i = 0; i < act.authorization.size(); ++i) {
             const auto& declared_auth = act.authorization[i];
 
             if (declared_auth.permission == config::sysio_payer_name) {
-               SYS_ASSERT( payer.empty(), irrelevant_auth_exception,
+               SYS_ASSERT( !has_payer, irrelevant_auth_exception,
                            "Multiple payers specified for action" );
                SYS_ASSERT( i == 0, irrelevant_auth_exception,
                            "Explicit payer must be the first declared authorization" );
-               payer = declared_auth.actor;
+               has_payer = true;
+               payer     = declared_auth.actor;
                continue;
             }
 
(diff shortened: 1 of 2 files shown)


huangminghuang (commented):
The original finding is addressed. Please also refresh the PR description so it captures the complete current PR: it still says the marker needs no check and that simply skipping it ensures agreement, which contradicts the follow-up implementation. It also omits the new structural/CFA coverage and retains the old validation count.

huangminghuang (approved):
Re-reviewed at 42a44037c6. The previously reported empty-actor and rejection-test issues are resolved; no remaining code findings.

PR #591, opened 2026-08-26
Producer: enable subjective billing by default, exclude the payer

heifner:
## Summary

Turns subjective CPU billing on by default, and excludes the contract payer from it.

| Option | Was | Now |
|---|---|---|
| `disable-subjective-p2p-billing` | `true` | `false` |
| `disable-subjective-api-billing` | `true` | `false` |
| `disable-subjective-payer-billing` | `false` | `true` |

These are node-local producer settings. No consensus effect, no protocol feature, no replay impact; operators can restore prior behaviour with the same flags.

## Why enable it

Antelope could afford to ship subjective billing off because the signer had staked CPU as a deterrent. Wire has no such backstop: every user account holds zero objective CPU by design, since the contract is the payer.

And objective billing never sees a failure. `add_transaction_usage` runs from `finalize()`, which the exception path does not reach. So today a failed transaction costs the signer nothing, the contract nothing, and the network nothing — while the producer still executes and reverts it. Subjective billing is the only mechanism that prices that work, and both `disable-subjective-p2p-billing` and `disable-subjective-api-billing` defaulting `true` means `producer_plugin` disables the feature outright.

Enabling it also enables the per-account failure limiter, which is gated behind the same flag: `subjective-account-max-failures` is 3 with a 1-block window, and the counter clears every block — so it takes a sustained six failures per second to stay throttled.

Six per second is cheap, which is why that threshold is only meaningful once the failures are attributed correctly. See **Attribution** below.

## Why exclude the payer

`disable-subjective-payer-billing` defaulted `false`, meaning the payer accrues subjective CPU too. Under contract-pays the payer is the called contract, so enabling subjective billing as-is would charge a contract for its callers' failures.

That points the cost at the wrong account. The contract is the party deliberately provisioned to absorb traffic; the account generating failures is the signer, and the signer is still billed through `auth_cpu`. Left as-is, a popular contract with an ordinary failure rate — asserts, races, front-run swaps — could be throttled node-locally through no fault of its own, and would look to an operator like the contract misbehaving.

## Attribution

Raised in review as a P1. `_account_fails` blamed the accounts named in `first_authorizers()`, which is read straight off the packed transaction and is attacker-chosen until `check_authorization` runs. `failure_limit()` then gates *before* `chain.push_transaction`, and `add()` counted every non-exhausted, non-duplicate failure — `unsatisfied_authorization` included, since its exception classification only builds the log line's "reason" string.

Enabling subjective billing by default is what made that reachable. Naming a victim on a transaction one cannot sign throttled the victim's own transactions on the receiving node. It did nothing to a real spammer: `validate_referenced_accounts` requires only that the actor and permission exist, so they name a different existing account each time. `add()` blames every per-action first authorizer, so one multi-action transaction reached several victims at once.

`controller` now records on `transaction_metadata` whether the signatures satisfied the declared authorizations, and the producer counts a failure only when they did. Execution failures still count exactly as before: `check_authorization` precedes `exec()`, and `schedule_action` records an action trace before executing, so an asserting first action is still blamed. The flag is cleared per attempt, since a transaction retried from the unapplied queue may name a permission that has been updated since — removing that one line makes the regression test fail.

Spring carries the same flaw (`_account_fails.add(first_auth, e)` on the untrusted declared authorizer, plus `subjective_bill_failure` on the same value) but is equally unreachable there, since it also defaults both flags off. Wire's `transaction_context` already attributed subjective CPU from real action traces, so only the plugin-side counter was affected.

## Note on history

Wire inherited the `true` defaults from Leap at `85486b769f init`; `bad4bb49eb GH-1173` later removed the `disable-subjective-billing` master switch, leaving this pair as the only control.

Upstream shipped the feature **on**: `9c254a9c35` ("Consolidated Security Fixes for 2.0.10", 2021-02-25) added `disable-subjective-billing` as a `bool_switch()->default_value(false)`, which is where the awkward name comes from. `8eb0915971` ("Consolidated Security Updates for 2.0.12", 2021-04-19) flipped it to `true` eight weeks later, and it has defaulted off through every Leap and Spring release since, including `spring/main` today. The 2.0.12 rationale is not recorded here.

## Harness option arity

Separate commit, raised in review as a P2. The `NodeopPluginArgs` generator typed each field by matching the option's *name* against a word list (`disable`, `enable`, `force`, `delete`, …), so the type tracked spelling rather than whether nodeop consumes a value. `disable-subjective-account-billing` is `value<vector<string>>()->multitoken()` but was typed `bool` because its name contains "disable"; `finality-data-history` is a `bool_switch()` but was typed `str`. The earlier CI fix on this branch emitted from that type, so the first rendered as a bare flag and the second as `--finality-data-history 1`.

`boost::program_options` already reports arity in the help output the generator parses: the value placeholder sits one space after the option name, and the description column starts after two or more spaces. That is now read before whitespace is collapsed and recorded as explicit `_<key>NodeopArgTakesValue` metadata, which the emitter uses instead of the field type. Two details matter — matching the literal token `arg` is wrong because an option with a `value_name()` prints that name instead (`--wasm-runtime runtime (=sys-vm-jit)`), and the match is anchored to the two-space option column so a wrapped description mentioning an option cannot overwrite its entry. `--delete-all-blocks` does exactly that.

Also fixed: `supportedNodeopArgs` matched `NodeopArg` unanchored, so the new sibling field would have silently joined the supported-option list.

Regenerating against `nodeop` changes exactly those two options and nothing else across the eight plugin files. Cross-checked against the source: all 19 `bool_switch()` declarations classify as switches, and all 23 options classified as switches are ones.

## Tests

`plugin_args_emission_test.py` covers the shapes that were rendered wrongly: a switch, a boolean-valued option, a value option typed `bool`, a plain value option, and the `supportedNodeopArgs` anchoring.

`auth_tests/auth_verified_tracks_authorization` covers the attribution flag: cleared when authorization fails, set on success, still set when authorization passes and execution then fails, and cleared again when the same metadata is re-pushed after the permission it names is re-keyed.

`subjective_billing_test.py` no longer passes `--disable-subjective-api-billing`, so it exercises the shipped configuration; if the default is flipped back it fails rather than silently covering nothing. Verified against a build carrying this change.

The mechanism is already covered by `subjective_billing_test/disable_payer_billing_test`, which asserts a disabled payer is skipped in both `subjective_bill` and `subjective_bill_failure` while `auth_cpu` still accumulates.

`ProducerPluginArgs.py` is generated at build time from `nodeop --help`, so the PerformanceHarness defaults follow automatically — confirmed regenerating to `payer=1, p2p=0, api=0`. The harness passes both flags explicitly from its own `--disable-subjective-billing` argument, so its behaviour is unchanged.


plugins/producer_plugin/src/producer_plugin.cpp
```diff
@@ -1465,11 +1465,13 @@ void producer_plugin::set_program_options(
...
          ("disable-subjective-account-billing", boost::program_options::value<vector<string>>()->composing()->multitoken(),
           "Account which is excluded from subjective CPU billing")
-         ("disable-subjective-payer-billing", bpo::value<bool>()->default_value(false),
-          "Disable subjective CPU billing for all contract payer accounts")
-         ("disable-subjective-p2p-billing", bpo::value<bool>()->default_value(true),
+         ("disable-subjective-payer-billing", bpo::value<bool>()->default_value(true),
+          "Disable subjective CPU billing for all contract payer accounts. On by default: under "
+          "contract-pays the payer is the called contract, so billing it for its callers' failures "
+          "charges the party provisioned to absorb traffic rather than the account causing it")
+         ("disable-subjective-p2p-billing", bpo::value<bool>()->default_value(false),
```
huangminghuang: [P1] Prevent spoofed account throttling

Enabling both billing paths also activates `_account_fails`, which attributes failures using untrusted declared `first_authorizers()` before authorization succeeds. A transaction naming a victim but carrying no valid victim signature reaches `unsatisfied_authorization`, and `handle_push_result()` still calls `_account_fails.add(auths, e)`. After three such transactions in a block, `failure_limit()` rejects the victim’s legitimate transactions before checking authorization; a multi-action transaction can target several victims at once. Repeating roughly six forged failures per second sustains the throttle on a default node. Please only count failures after successful authorization, or exclude pre-authorization/authentication failures before enabling these defaults.

tests/PerformanceHarness/NodeopPluginArgs/BasePluginArgs.py
```diff
@@ -24,9 +24,17 @@ def __str__(self) -> str:
...
                 if current is not None and current != default:
-                    if type(current) is bool:
-                        args.append(f"{getattr(self, f'_{field.name}NodeopArg')}")
+                    arg = getattr(self, f'_{field.name}NodeopArg')
+                    # Emit a bare flag only for options nodeop declares as a switch. The generator
+                    # types those `bool`; an option that takes a value is typed from its default
+                    # (`int`, `str`, ...). Keying on the runtime type instead would emit a bare flag
+                    # for any field merely *assigned* a Python bool -- which nodeop then rejects,
+                    # because boost consumes the following argument as the missing value.
+                    if field.type is bool:
```
huangminghuang: [P2] Do not infer option arity from field type

`field.type` reflects the generator’s naming/default heuristics, not whether Boost consumes a value. For example, `disableSubjectiveAccountBilling` is generated as `bool` because its name contains `disable`, although nodeop declares `--disable-subjective-account-billing` as `value<vector<string>>()->multitoken()`; assigning `account1` now emits a bare option. Conversely, `finalityDataHistory` is generated as `str` although `--finality-data-history` is a `bool_switch()`, so assigning `True` emits the invalid form `--finality-data-history 1`. Both cases rendered correctly under the previous runtime-type dispatch. Please carry explicit switch/value metadata from the generator and add regression coverage for switches, boolean-valued options, and non-boolean value options.

huangminghuang (commented):
Re-reviewed at `0675cd6d33`; no new commits have been pushed since the previous review. The existing findings remain unresolved:

- [P1: unauthenticated transactions can throttle arbitrary accounts](https://github.com/Wire-Network/wire-sysio/pull/591#discussion_r3873109598)
- [P2: generated CLI option arity is inferred incorrectly](https://github.com/Wire-Network/wire-sysio/pull/591#discussion_r3873113623)

I am not approving this revision until both issues are addressed.

plugins/producer_plugin/src/producer_plugin.cpp
heifner: Confirmed and fixed in `71e86dc9a3`. I traced each step rather than taking the report on faith, and it holds exactly as described: `auths` comes from `first_authorizers()` off the packed transaction, `failure_limit()` gates before `chain.push_transaction`, and `add()` counts every non-exhausted, non-duplicate failure — `unsatisfied_authorization` included, since `account_failure::add` classifies exception codes only to build the log line's "reason" string.

Two things sharpen it further. `validate_referenced_accounts` requires only that the actor and permission *exist*, so a real spammer cycles account names and is never throttled — the pre-authorization counter is close to pure offense. And `add()` blames every per-action first authorizer, so one multi-action transaction reaches several victims at once.

I did not take the "exclude pre-authorization failures" route as a denylist of exception codes. The forgeable set is not the authorization set: `expired_tx_exception`, `invalid_ref_block_exception`, the `validate_referenced_accounts` throws, and init-time `tx_net_usage_exceeded` all fire before `check_authorization` and are equally attacker-chosen, and no exception hierarchy separates pre- from post-authorization. A denylist would have left the victim reachable through a different code.

Instead `controller` records on `transaction_metadata` whether the signatures satisfied the declared authorizations, and the producer counts a failure only when they did. I considered keying on `!trace->action_traces.empty()` — equivalent today, since `check_authorization` precedes `exec()` — but an explicit flag states the rule at the read site and moves with the check if anyone reorders it. `transaction_metadata` already carries `elapsed`, `prev_accounts_billing` and `prev_succeeded` as controller-to-plugin writebacks, so this fits the existing pattern; `transaction_trace` would have been a wire-format change.

The flag is cleared per attempt, guarded by `check_auth` so nothing is written from the read-only thread pool, with an assert pinning that. Removing the clear alone makes the regression test fail, which is what pins the retry case.

Execution failures are unaffected: `schedule_action` records an action trace before executing, so a first action that asserts is still blamed. `auth_tests/auth_verified_tracks_authorization` covers authorization failure, success, authorization-passes-then-execution-fails, and a re-push after the named permission is re-keyed.

Worth noting for context: Spring has the same flaw — `_account_fails.add(first_auth, e)` and `subjective_bill_failure` both on the untrusted declared authorizer — but is equally unreachable there, since it also defaults both flags off. Wire's `transaction_context` already attributed subjective CPU from real action traces, so only the plugin-side counter needed the fix.

tests/PerformanceHarness/NodeopPluginArgs/BasePluginArgs.py
heifner: Correct on both counts, and fixed in `553e622e8c` as a separate commit. I verified each example against the declarations: `disable-subjective-account-billing` is `value<vector<string>>()->composing()->multitoken()` and was typed `bool` purely because its name contains "disable"; `finality-data-history` is `bpo::bool_switch()->default_value(false)` and was typed `str`.

The generator now carries explicit `_<key>NodeopArgTakesValue` metadata and the emitter uses it instead of `field.type`. The arity comes from what nodeop itself prints: `boost::program_options` puts the value placeholder one space after the option name and starts the description column after two or more spaces. That has to be read before the existing pipeline collapses whitespace, which erases the distinction.

Two details cost me a wrong revision each, so they are worth recording:

- Matching the literal token `arg` is not sufficient. An option that set a `value_name()` prints that name instead — `--wasm-runtime runtime (=sys-vm-jit)` and `--producer-name arg` are both value options, but only one says "arg". Keying on the placeholder's *presence* rather than its spelling is what works.
- Matching `--option` anywhere on an indented line picks up wrapped description text. `--delete-all-blocks` is mentioned inside its own description continuation line, which silently overwrote its real entry and flipped it to a value option. The match is now anchored to the two-space option column, which also handles the short-form spelling `-e [ --enable-stale-production ]`.

Separately, `supportedNodeopArgs` matched `NodeopArg` unanchored, so the new sibling field would have joined the supported-option list. Now anchored.

Verification: regenerating against `nodeop` and diffing field types against the previous generator across all eight plugin files changes exactly the two options you named and nothing else. Cross-checked independently against the source — all 19 `bool_switch()` declarations classify as switches, and all 23 options classified as switches are genuinely switches, so no misclassification in either direction.

Regression coverage is `plugin_args_emission_test.py`, registered with ctest: a switch, a boolean-valued option rendering 1/0 rather than True/False, a value option typed `bool`, a plain non-boolean value option, and the `supportedNodeopArgs` anchoring.

The name-based type heuristic is gone except for one narrow case: value options nodeop prints without a default and whose names match `sizegb|maxage|retainblocks` keep their `int` annotation. That is documentation for harness users only — arity decides rendering — and it keeps the regeneration diff limited to the two genuine corrections.

huangminghuang (approved):
Re-reviewed the full cumulative diff at 31d9a10d43d64d9c484773adad125b27e91785e3. The prior authorization-attribution and CLI arity findings are resolved, the added regression coverage is appropriate, and current-head CI is green.

PR #593, opened 2026-08-28
fix(outpost_solana_client): carry the custody mint and token program on reserve remits

valthon:
Every SPL swap remit currently aborts on the Solana outpost and wedges the epoch. This restores the relay's dispatch manifest to the contract the program has required since SOL-396.

## What broke

wire-solana `896ed2fa` (SOL-396, landed 2026-08-27) changed the SPL settlement path. `reserve_vault_transfer` went from:

```rust
- let token_prog_ai = require_remaining_account(remaining, &token::ID, "SPL token program")?;
+ require_remaining_account(remaining, &custody_token_program, "custody token program")?;
+ let mint_ai        = require_remaining_account(remaining, &custody_mint, "custody mint")?;
```

so the **mint is newly required**, and the destination ATA is now derived with the reserve's own `custody_token_program` rather than the SPL-Token default.

This plugin's manifest never moved in lock-step. `swap_remit` was the only reserve-backed shape that never carried the mint — `swap_revert` and `reserve_create_cancelled` already did — which is why the breakage is confined to remits.

The plugin's own comment predicted this exactly:

> LOCK-STEP: … a handler that `require_remaining_account`s an account this manifest omits aborts the dispatch round and re-packs the identical window from the same cursor on every retry.

## Observed impact

e2e run `33187348198`: `flow-swap-non-native-tokens` and `flow-swap-private-reserves` both failed. Each stalled on a single epoch (12 and 9), retrying the dispatch drain 55 times, with `custom program error: 0x17b6` (`EffectAccountMissing`) appearing **165 times in exactly those two flows and zero times in the other twelve**. Both timed out after 840s waiting for a Solana SPL destination that could never be credited.

Reproduced across three separate branch combinations, two of which had wire-solana at the plain `next` tip — so this is the relay, not any one branch.

## The change

- `swap_remit` passes `custody_mint`.
- All three reserve-backed shapes derive the ATA and the token-program account from the Reserve's `custody_token_program`. The same `(owner, mint)` pair has a different canonical ATA under Token-2022, so the legacy default names an account the program never asks for — indistinguishable from a missing one.
- `custody_token_program` is read off the Reserve alongside creator / mint / decimals and required in the IDL shape check, on the same reasoning the other three already carry: it is written at reserve creation, and a manifest built by guessing it is guaranteed to abort on chain.

**Collateral paths are deliberately untouched** — `resolve_collateral_vault_transfer` still requires `token::ID` and no mint, and `CollateralPosition` has no token program of its own.

## Verification

- Plugin test suite green.
- **`flow-swap-non-native-tokens` SUCCEEDED** on a local cluster: `EffectAccountMissing` 165 → 0, dispatch-drain failures 55 → 0, epoch advanced past 18 instead of freezing at 12, and the `usdc-permit-to-usdtsol` payout completed instead of timing out. The run also reached `SwapSolanaToEthereumSpl`, which the failing runs never got to.

## Tests

The existing SPL manifest case asserted the ATA and token program but never the mint — which is precisely how the omission shipped. It now asserts the mint, and two new cases pin token-program-aware derivation for all three reserve-backed shapes. Both were confirmed to fail when the derivation is reverted, rather than being assumed to bite.

The `Reserve` IDL stub fixture was pre-SOL-396 and gains the field.

## Deploy note

The new field is required with no fallback, per `no-back-compat-before-release`. The plugin therefore refuses to run against a pre-SOL-396 program rather than silently deriving a wrong ATA — the right failure direction, but it does mean wire-sysio and wire-solana `next` need to land together.

https://claude.ai/code/session_01NoaXitAvpNSNj7114S9338

### DIFF
--- a/plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
+++ b/plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
@@ -123,14 +123,16 @@ namespace epoch_deliveries {
 } // namespace epoch_deliveries
 
 /// Field identifiers of the per-`(token_code, reserve_code)` `Reserve`
-/// account. `custody_mint` / `custody_decimals` are pinned at reserve creation
+/// account. `custody_mint` / `custody_decimals` / `custody_token_program` are
+/// pinned at reserve creation
 /// and are what the on-chain terminal handlers branch on, so the relay's
 /// manifest must read custody from HERE and nowhere else.
 namespace reserve_account {
-   constexpr auto account_name           = "Reserve";
-   constexpr auto field_creator          = "creator";
-   constexpr auto field_custody_mint     = "custody_mint";
-   constexpr auto field_custody_decimals = "custody_decimals";
+   constexpr auto account_name                = "Reserve";
+   constexpr auto field_creator               = "creator";
+   constexpr auto field_custody_mint          = "custody_mint";
+   constexpr auto field_custody_decimals      = "custody_decimals";
+   constexpr auto field_custody_token_program = "custody_token_program";
 } // namespace reserve_account
 
 /// Field identifiers of the per-`(operator, token_code)` collateral position.
@@ -337,9 +339,10 @@ void assert_reserve_shape(const fc::network::solana::idl::program& program) {
    namespace idl = fc::network::solana::idl;
    const auto& fields = declared_account_fields(program, reserve_account::account_name);
 
-   bool has_creator  = false;
-   bool has_mint     = false;
-   bool has_decimals = false;
+   bool has_creator       = false;
+   bool has_mint          = false;
+   bool has_decimals      = false;
+   bool has_token_program = false;
    for (const auto& field : fields) {
       // Compare against the optional member directly, as the sibling shape
       // checks do: a malformed type object must reach the per-field diagnostic
@@ -359,18 +362,24 @@ void assert_reserve_shape(const fc::network::solana::idl::program& program) {
                    "Reserve '{}' must be declared u8, got '{}'",
                    reserve_account::field_custody_decimals, describe_idl_type(field.type));
          has_decimals = true;
+      } else if (field.name == reserve_account::field_custody_token_program) {
+         FC_ASSERT(is_pubkey, "Reserve '{}' must be declared pubkey, got '{}'",
+                   reserve_account::field_custody_token_program, describe_idl_type(field.type));
+         has_token_program = true;
       }
    }
-   // One message for all three: they are written together at reserve creation,
+   // One message for all four: they are written together at reserve creation,
    // and any one missing costs the manifest the same way -- the effect account
    // the program's branch requires cannot be derived, so its dispatch window
    // aborts and the cursor never moves past it.
-   FC_ASSERT(has_creator && has_mint && has_decimals,
-             "Reserve IDL missing '{}' / '{}' / '{}' (found {} / {} / {}); the terminal manifest "
-             "resolves custody and the cancel-refund target from these, and a reserve-backed "
-             "dispatch window cannot be built without them",
+   FC_ASSERT(has_creator && has_mint && has_decimals && has_token_program,
+             "Reserve IDL missing '{}' / '{}' / '{}' / '{}' (found {} / {} / {} / {}); the terminal "
+             "manifest resolves custody, the cancel-refund target and the ATA-deriving token "
+             "program from these, and a reserve-backed dispatch window cannot be built without them",
              reserve_account::field_creator, reserve_account::field_custody_mint,
-             reserve_account::field_custody_decimals, has_creator, has_mint, has_decimals);
+             reserve_account::field_custody_decimals,
+             reserve_account::field_custody_token_program,
+             has_creator, has_mint, has_decimals, has_token_program);
 }
 
 /// Keep only the candidate IDLs whose declared address matches the deployed
@@ -772,6 +781,10 @@ reserve_terminal_info reserve_info_from_account(const fc::variant_object& reserv
              reserve_account::field_custody_mint);
    FC_ASSERT(reserve.contains(reserve_account::field_custody_decimals),
              "Reserve account missing '{}' field", reserve_account::field_custody_decimals);
+   FC_ASSERT(reserve.contains(reserve_account::field_custody_token_program),
+             "Reserve account missing '{}' field — the canonical ATA differs per token program, "
+             "so deriving with the legacy default would name an account the program never asks for",
+             reserve_account::field_custody_token_program);
 
    const auto decimals = reserve[reserve_account::field_custody_decimals].as_uint64();
    FC_ASSERT(decimals <= std::numeric_limits<uint8_t>::max(),
@@ -783,7 +796,9 @@ reserve_terminal_info reserve_info_from_account(const fc::variant_object& reserv
          reserve[reserve_account::field_creator].as_string()),
       fc::network::solana::solana_public_key::from_base58_string(
          reserve[reserve_account::field_custody_mint].as_string()),
-      static_cast<uint8_t>(decimals)};
+      static_cast<uint8_t>(decimals),
+      fc::network::solana::solana_public_key::from_base58_string(
+         reserve[reserve_account::field_custody_token_program].as_string())};
 }
 
 std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manifests(
@@ -925,11 +940,13 @@ std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manif
          //
          // The accounts are still passed because the reserve may be created
          // between this read and the crank landing. If it is, this manifest is
-         // complete only for a NATIVE swap remit/revert (Reserve PDA +
-         // recipient); an SPL one is short the recipient ATA, and a
-         // RESERVE_CREATE_CANCELLED is short the `creator` on EITHER custody
-         // kind, since the creator is read off the Reserve we could not
-         // decode. Those windows abort and the next tick rebuilds them against
+         // complete only for a NATIVE swap remit/revert (Reserve PDA + recipient);
+         // an SPL one is short the recipient ATA, the custody mint and the custody
+         // token program — all three are read off the Reserve we could not decode,
+         // and the ATA is not even derivable without the token program (SOL-396) —
+         // and a RESERVE_CREATE_CANCELLED is short the `creator` on EITHER custody
+         // kind, since the creator is read off the Reserve we could not decode.
+         // Those windows abort and the next tick rebuilds them against
          // a now-readable Reserve.
          //
          // The degrade is PER-ATTESTATION: the manifests are built for every
@@ -944,19 +961,41 @@ std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manif
               log_label, token_code, reserve_code, effect.attestation_index);
          add(vault_pda, true);
          if (effect.recipient) add(*effect.recipient, true);
+         // Best-effort only, and right solely for a legacy-SPL reserve: the real
+         // token program is pinned on the Reserve we could not read. Not the
+         // sanctioned derivation — see `custody_token_program` below.
          add(token_program_id, false);
          continue;
       }
       const auto& info = *info_opt;
+
+      // SOL-396 LOCK-STEP: every reserve-backed SPL settlement routes through the
+      // program's `reserve_vault_transfer`, which requires BOTH the reserve's
+      // `custody_token_program` AND its `custody_mint` in remaining_accounts, and
+      // derives the destination ATA with that token program. Both facts come from
+      // the Reserve — never the legacy SPL-Token default — because the same
+      // (owner, mint) pair has different canonical ATAs under each program.
+      const auto& custody_token_program = info.custody_token_program;
+      const auto  custody_ata           = [&](const fc::network::solana::solana_public_key& owner) {
+         return fc::network::solana::system::get_associated_token_address(
+            owner, info.custody_mint, custody_token_program);
+      };
 
       switch (effect.shape) {
          case effect_shape::swap_remit: {
             if (!effect.recipient) break;
             if (is_native_custody(info.custody_mint)) { add(*effect.recipient, true); break; }
             add(vault_pda, true);
-            add(fc::network::solana::system::get_associated_token_address(
-                   *effect.recipient, info.custody_mint), true);
-            add(token_program_id, false);
+            add(custody_ata(*effect.recipient), true);
+            // The mint is REQUIRED here, not optional: `reserve_vault_transfer`
+            // `require_remaining_account`s it, and the CPI's `transfer_checked`
+            // needs the mint account to verify the decimals against — the decimals
+            // themselves come from the Reserve record, not from this account.
+            // Omitting it is what aborted every SPL swap remit with
+            // `EffectAccountMissing` after SOL-396 — the one reserve-backed shape
+            // that had never carried it.
+            add(info.custody_mint, false);
+            add(custody_token_program, false);
             break;
          }
          case effect_shape::swap_revert: {
@@ -965,9 +1004,8 @@ std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manif
             add(vault_pda, true);
             add(info.custody_mint, false);
             add(*effect.recipient, true);
-            add(fc::network::solana::system::get_associated_token_address(
-                   *effect.recipient, info.custody_mint), true);
-            add(token_program_id, false);
+            add(custody_ata(*effect.recipient), true);
+            add(custody_token_program, false);
             add(associated_token_program_id, false);
             add(system_program_id, false);
             break;
@@ -976,10 +1014,9 @@ std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manif
             if (is_native_custody(info.custody_mint)) { add(info.creator, true); break; }
             add(vault_pda, true);
             add(info.creator, false);
-            add(fc::network::solana::system::get_associated_token_address(
-                   info.creator, info.custody_mint), true);
+            add(custody_ata(info.creator), true);
             add(info.custody_mint, false);
-            add(token_program_id, false);
+            add(custody_token_program, false);
             add(associated_token_program_id, false);
             add(system_program_id, false);
             break;
--- a/plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
+++ b/plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
@@ -333,7 +333,7 @@ void assert_collateral_position_shape(const fc::network::solana::idl::program& p
              has_operator, has_token_code, has_mint, has_amount);
 }
 
-/// Assert the loaded IDL declares `Reserve` with the three fields the terminal
+/// Assert the loaded IDL declares `Reserve` with the four fields the terminal
 /// manifest resolves from it. Full contract on the header declaration.
 void assert_reserve_shape(const fc::network::solana::idl::program& program) {
    namespace idl = fc::network::solana::idl;
@@ -661,7 +661,6 @@ constexpr uint8_t  META_DISC_PUBKEY_DATA    = 2;
 constexpr uint8_t  META_DISC_INDEX_BASE     = 0x80;  // U8_TOP_BIT
 constexpr uint8_t  SEED_UNINITIALIZED       = 0;
 constexpr uint8_t  SEED_LITERAL             = 1;
-constexpr uint8_t  SEED_INSTRUCTION_DATA    = 2;
 constexpr uint8_t  SEED_ACCOUNT_KEY         = 3;
 
 /// First 8 bytes of sha256("spl-transfer-hook-interface:execute").
@@ -717,23 +716,45 @@ derive_extra_account_metas_pda(const fc::network::solana::solana_public_key& hoo
 
 std::vector<extra_account_meta>
 parse_extra_account_metas(const std::vector<uint8_t>& validation_account_data) {
+   // EVERY malformed shape below throws rather than returning a short list.
+   // A partial parse is the worst possible outcome: `mint_transfer_hook_for`
+   // would wrap it as authoritative, the manifest would carry the hook program
+   // and validation PDA while silently omitting declared metas, and the CPI
+   // would then fail with `IncorrectAccount` -- window aborted, cursor pinned,
+   // nothing in the log naming the cause. That is exactly what the resolver
+   // below refuses to do, and the parser must not undercut it.
    std::vector<extra_account_meta> metas;
-   if (validation_account_data.size() < 12) return metas;
+   const size_t size = validation_account_data.size();
+   FC_ASSERT(size >= 12,
+             "ExtraAccountMetaList account is {} bytes, too short to hold even one TLV header",
+             size);
    const auto* bytes = validation_account_data.data();
 
    size_t cursor = 0;
-   while (cursor + 12 <= validation_account_data.size()) {
+   while (cursor + 12 <= size) {
       const bool is_execute =
          std::memcmp(bytes + cursor, EXECUTE_TLV_DISCRIMINATOR.data(), 8) == 0;
       const uint32_t entry_len = le_u32(bytes + cursor + 8);
       const size_t   value     = cursor + 12;
-      if (value + entry_len > validation_account_data.size()) break;
+      FC_ASSERT(value + entry_len <= size,
+                "ExtraAccountMetaList TLV entry at {} claims {} bytes but only {} remain",
+                cursor, entry_len, size - value);
       if (is_execute) {
-         if (entry_len < 4) break;
+         FC_ASSERT(entry_len >= 4,
+                   "ExtraAccountMetaList Execute entry is {} bytes, too short for its count",
+                   entry_len);
          const uint32_t count = le_u32(bytes + value);
+         // `count` comes from account data we do not control. Bound the walk by
+         // the ENTRY's extent, never by the account's: a count that overruns the
+         // entry would otherwise parse the FOLLOWING TLV bytes as metas and push
+         // arbitrary pubkeys into the manifest as writable accounts.
+         const size_t declared = static_cast<size_t>(count) * EXTRA_ACCOUNT_META_SIZE;
+         FC_ASSERT(4 + declared <= entry_len,
+                   "ExtraAccountMetaList declares {} metas ({} bytes) but its entry holds {}",
+                   count, declared, entry_len);
+         metas.reserve(count);
          for (uint32_t i = 0; i < count; ++i) {
             const size_t at = value + 4 + static_cast<size_t>(i) * EXTRA_ACCOUNT_META_SIZE;
-            if (at + EXTRA_ACCOUNT_META_SIZE > validation_account_data.size()) break;
             extra_account_meta meta;
             meta.discriminator = bytes[at];
             std::memcpy(meta.address_config.data(), bytes + at + 1, 32);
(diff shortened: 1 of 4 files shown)


heifner:
Reviewed against the diff, the surrounding plugin code, and the `wire-solana` `origin/next` program this must lock-step with.

**Verified correct:** `require_remaining_account` matches by pubkey, so ordering is not load-bearing; `fc::network::solana::system::get_associated_token_address`'s 3-arg overload uses the correct `[owner, token_program, mint]` seed order; the collateral paths genuinely still require `token::ID` with no mint (`inbound.rs:637`); the SPL branches of `handle_swap_remit` / `handle_swap_revert` return before the wallet-writable asserts, so omitting the bare wallet from the SPL manifest is right; and the IDL fixture inserts `custody_token_program` between `custody_decimals` and `name_len`, matching `opp_states.rs:528`.

Three findings.

---

### 1. HIGH — the lock-step restoration is still incomplete for transfer-hook mints, which is the exact mint SOL-396 was written for

`plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp:985`

`reserve_vault_transfer` routes through `spl_token_2022::onchain::invoke_transfer_checked(..., remaining, ...)`, and its own doc-comment states the extras are "resolved from `remaining`". When the mint carries the TransferHook extension, `add_extra_accounts_for_execute_cpi` hard-errors (`TransferHookError::IncorrectAccount`) unless the hook **program id** and the mint's **extra-account-metas validation PDA** — plus any accounts the hook itself declares — are present in `remaining_accounts`. This manifest builder adds none of them; `grep -n "hook\|extra_account"` over the plugin returns nothing.

`init_liqsol_reserve.rs` pins liqSOL reserves under Token-2022 with `TRANSFER_HOOK_PROGRAM_ID` and a per-vault hook record. So a `SWAP_REMIT` / `SWAP_REVERT` / `RESERVE_CREATE_CANCELLED` out of a liqSOL reserve aborts the dispatch window and re-packs from the unadvanced cursor on every retry — the identical wedge this PR is fixing, one layer deeper.

Neither side currently covers it: `wire-solana`'s `tests.rs:2295` says the fixture uses a legacy mint "so the hook lookup finds none and the plain transfer path runs."

Either extend the manifest (derive the validation PDA from mint + hook program, and carry the hook program id), or scope it out explicitly with a follow-up ticket — as written, the PR body's claim that this "restores the relay's dispatch manifest to the contract the program has required since SOL-396" is not yet true for liqSOL.

### 2. LOW — the new boot gate is untested

`plugins/outpost_solana_client_plugin/test/test_outpost_solana_client_plugin.cpp:2517`

`reserve_shape_rejects_drifted_declarations` still enumerates only the five pre-existing cases (creator / mint / decimals dropped, mint-as-byte-array, decimals-as-u32). None has creator + mint + decimals present with `custody_token_program` absent — i.e. the pre-SOL-396 IDL this PR's boot check exists to reject — and none declares it with a wrong type.

Deleting `has_token_program` from the `FC_ASSERT`, or dropping its `is_pubkey` branch, leaves the suite green (`reserve_shape_accepts_the_deployed_declaration` still passes against the updated fixture). That is the same coverage hole the PR body identifies as "precisely how the omission shipped," reproduced on the shape-check side. `reserve_info_requires_every_custody_field` got its new case; `assert_reserve_shape` did not.

### 3. LOW — `assert_reserve_shape`'s contract doc is now wrong

`plugins/outpost_solana_client_plugin/include/sysio/outpost_solana_client_plugin/outpost_solana_client.hpp:209`

Still reads "carries the **three** fields ... `creator` and `custody_mint` (pubkeys) and `custody_decimals` (u8)", and line 222's `@throws` still says "any of the **three** fields is absent". The function now requires four and throws on a missing or wrong-typed `custody_token_program`. The sibling docs for `reserve_terminal_info` / `reserve_info_from_account` were correctly updated to "ALL FOUR" in this same diff, so this is a missed occurrence — and this header is described elsewhere in the file as the authoritative full contract on the declaration.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -944,19 +961,41 @@ std::vector<std::vector<fc::network::solana::account_meta>> build_dispatch_manif
...
+            add(custody_ata(*effect.recipient), true);
+            // The mint is REQUIRED here, not optional: `reserve_vault_transfer`
+            // `require_remaining_account`s it, and the CPI's `transfer_checked`
+            // needs the mint account to verify the decimals against — the decimals
+            // themselves come from the Reserve record, not from this account.
+            // Omitting it is what aborted every SPL swap remit with
+            // `EffectAccountMissing` after SOL-396 — the one reserve-backed shape
+            // that had never carried it.
+            add(info.custody_mint, false);
+            add(custody_token_program, false);
```
huangminghuang: **[P1] Include Token-2022 TransferHook accounts.** This manifest now supplies the program-aware ATA, custody mint, and custody token program, but SOL-396 also changed reserve settlement to `spl_token_2022::onchain::invoke_transfer_checked(..., remaining, ...)`. For a mint with `TransferHook`—including canonical liqSOL—that helper first looks up the hook program and then resolves the `ExtraAccountMetaList` accounts. Neither this manifest nor the fixed `dispatch_attestations` accounts supplies them, so the helper returns `IncorrectAccount` before the transfer; the transaction aborts and the dispatch cursor remains pinned. The `swap_revert` and `reserve_create_cancelled` branches have the same omission. Please resolve and include the hook program, validation PDA, and referenced metas with their required writability for all three shapes, and cover a real hook-enabled Token-2022 mint rather than only a program-aware ATA derivation.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
```diff
@@ -359,18 +362,24 @@ void assert_reserve_shape(const fc::network::solana::idl::program& program) {
...
-   // One message for all three: they are written together at reserve creation,
+   // One message for all four: they are written together at reserve creation,
    // and any one missing costs the manifest the same way -- the effect account
    // the program's branch requires cannot be derived, so its dispatch window
    // aborts and the cursor never moves past it.
-   FC_ASSERT(has_creator && has_mint && has_decimals,
-             "Reserve IDL missing '{}' / '{}' / '{}' (found {} / {} / {}); the terminal manifest "
-             "resolves custody and the cancel-refund target from these, and a reserve-backed "
-             "dispatch window cannot be built without them",
+   FC_ASSERT(has_creator && has_mint && has_decimals && has_token_program,
```
huangminghuang: **[P2] Add a negative test that isolates `custody_token_program`.** The existing `reserve_shape_rejects_drifted_declarations` cases all omit this new field while testing some other missing or type error, and there is no case where the other three required fields are valid and only `custody_token_program` is absent or non-pubkey. Removing this new check would therefore leave the suite green. Please build each rejection case from a valid four-field declaration and mutate one field at a time.

valthon:
All three findings addressed in `5d68f88b` — thanks, the P1 was a real gap and the boot-gate one was the same coverage hole reproduced on the shape-check side.

## Transfer-hook accounts (P1 / finding 1)

Built rather than scoped out. The manifest now carries the hook program, the `ExtraAccountMetaList` validation PDA, and every account that PDA declares, for **all three** reserve-backed shapes.

It reads chain rather than mirroring liqSOL's layout, because hard-coding one hook's shape reproduces exactly the failure mode this PR exists to fix — a relay-side assumption that silently diverges when the program changes:

- the mint's TLV is walked for extension 14, taking `program_id` at value+32 and treating an all-zero id as a disabled hook;
- the validation PDA is `["extra-account-metas", mint]` under that program, and its Execute TLV entry (discriminator `692565c5 4bfb661a`) is parsed into the declared metas;
- each meta resolves by discriminator — literal, PDA of the hook program, or PDA of the account at `discriminator - 0x80` — with `Literal` and `AccountKey` seeds evaluated against the Execute account order `[source, mint, destination, authority, validation_pda]`. Resolved metas append to that list as they go, since a later meta may key off an earlier one.

Layout constants are from spl-token-2022 6.0.0 and spl-tlv-account-resolution 0.9.0 rather than memory.

**Two paths deliberately throw** instead of degrading, because the quiet alternative emits a manifest short the accounts the CPI is about to demand — invisible until an outpost stalls: an unimplemented seed or discriminator form (`InstructionData`, `AccountData`, pubkey-from-account-data), and a mint that *declares* a hook whose validation account cannot be read.

Coverage is a hook-enabled mint modelled on the real liqSOL meta list — a literal plus two `user_record` PDAs seeded on the source and destination token accounts, which is what makes the Execute account *order* load-bearing rather than decorative — asserting derivation and per-meta writability. A companion case pins that a hook-free mint's manifest is unchanged at exactly five accounts, so this is provably additive for every mint that exists today.

## Boot gate (P2 / finding 2)

You were both right that deleting the check left the suite green. Every rejection case is now one mutation off a *valid four-field* declaration — including the field absent (the pre-SOL-396 IDL) and mis-typed. Verified by regression: removing `has_token_program` now fails on `custody_token_program dropped (pre-SOL-396 IDL)`, and disabling hook appending fails 7 assertions.

## Doc (finding 3)

`assert_reserve_shape`'s contract and `@throws` both say four fields now — a missed occurrence when the sibling docs were updated in the same diff.

## One limitation worth stating

No flow exercises a hook mint, so the hook path is covered by unit tests only. That is also true program-side — `tests.rs:2295` notes the fixture uses a legacy mint "so the hook lookup finds none and the plain transfer path runs" — so neither side has end-to-end coverage of it yet.

The PR body's claim about restoring the SOL-396 contract is now accurate rather than aspirational.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
valthon: Built in `5d68f88b`. All three reserve-backed shapes now carry the hook program, the `ExtraAccountMetaList` validation PDA, and the accounts that PDA declares.

Resolution reads chain rather than mirroring liqSOL: mint TLV walked for extension 14, validation PDA derived as `["extra-account-metas", mint]`, its Execute entry parsed, and each meta resolved by discriminator (literal / hook PDA / PDA of the account at `disc - 0x80`) with `Literal` and `AccountKey` seeds evaluated against `[source, mint, destination, authority, validation_pda]` — appending as it goes, since a later meta may key off an earlier one.

Unimplemented seed forms and an unreadable validation account both throw rather than degrading, since a silently-short manifest is the wedge itself.

Coverage is a hook-enabled mint modelled on the real liqSOL meta list (literal + two `user_record` PDAs seeded on source and destination), plus a hook-free case pinning the manifest unchanged at five accounts. Verified by regression: disabling hook appending fails 7 assertions.

plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp
valthon: Fixed in `5d68f88b`. You were right that the check was unexercised — every case omitted `custody_token_program` while testing some other defect.

Each rejection case is now one mutation off a valid four-field declaration: each field dropped in turn (including `custody_token_program`, i.e. the pre-SOL-396 IDL the gate exists to reject) and each mis-typed. Verified by regression: removing `has_token_program` from the assert now fails on `custody_token_program dropped (pre-SOL-396 IDL)`, where before the suite stayed green.

heifner:
Re-reviewed at `5d68f88`. Both earlier findings are resolved: the boot-gate drift suite now builds every case as the valid four-field declaration with exactly one field dropped or mis-typed — including `custody_token_program dropped (pre-SOL-396 IDL)` — and the header contract reads four fields.

The new transfer-hook resolution checks out against the on-chain counterpart in `wire-solana` `d47150a6` (PR #439, `programs/liqsol-core/src/instructions/opp/inbound.rs`): `reserve_vault_transfer`'s `require_remaining_account` order, the vault being its own transfer authority (so `add_hook_accounts(vault, ata, vault)` is right), the Execute prefix order `[source, mint, destination, authority, validation]`, the `ExecuteInstruction` TLV discriminator bytes, `ExtensionType::TransferHook == 14`, and the real hook (`programs/transfer-hook/src/lib.rs:248-295`) declaring only `Literal` + `AccountKey` seeds and external-PDA discriminators — all forms the new resolver implements. Collateral settlement was correctly left pinned to `token::ID`.

Five findings, all in `plugins/outpost_solana_client_plugin/src/outpost_solana_client.cpp`.

---

### 1. `:719` — `parse_extra_account_metas` drops metas silently on every malformed shape

`size < 12` (721), `value + entry_len > size` (730), `entry_len < 4`, no Execute entry found, and a truncated meta (736 — which `break`s the inner loop and then returns a **partial** list) all yield a short list rather than an error.

`mint_transfer_hook_for` wraps whatever comes back as an authoritative `mint_transfer_hook`, so `add_hook_accounts` emits the hook program + validation PDA while silently omitting some or all declared metas. `add_extra_accounts_for_execute_cpi` then fails with `IncorrectAccount`: the dispatch window aborts, the cursor stays pinned, and nothing in the log names the cause.

That is precisely what the header comment on `resolve_hook_metas` says is worth throwing to prevent — "silently dropping a meta produces exactly the wedged epoch this resolution exists to prevent, and it would be invisible until an outpost stalled". The parser should `FC_ASSERT` on each of these, consistent with the resolver.

### 2. `:736` — the meta loop is bounded by the account size, not the Execute entry's extent

The per-meta bound is checked against `validation_account_data.size()` rather than `value + entry_len`. `count` (line 733) is read from untrusted account data; if it claims more metas than `entry_len` covers, the loop keeps parsing 35-byte records out of the *following* TLV bytes and pushes them as real `extra_account_meta`s. Those resolve to arbitrary pubkeys that enter the manifest as writable accounts.

Bound the loop with `std::min(size, value + entry_len)`, and treat a `count` that disagrees with `entry_len` as an error rather than parsing past it.

### 3. `:1190` — `add_hook_accounts` re-reads the custody mint per effect, uncached

`read_transfer_hook(info.custody_mint)` is called inline for every effect, unlike its two siblings in the same function, which are explicitly cached ("One read per DISTINCT reserve for the whole build").

Every reserve-backed SPL attestation now costs an extra `get_account_info(custody_mint)` — two when the mint has a hook, since `mint_transfer_hook_for` also reads the validation PDA — even though the reader returns `nullopt` for every mint in the system today. An envelope with 30 SPL remits over 3 reserves goes from 3 reads to 33, inside a build gated by `throw_if_past_deadline()` per effect, and multiplies exposure to a transient RPC throw aborting the whole build.

A `std::map<solana_public_key, std::optional<mint_transfer_hook>>` alongside `reserve_cache` covers it.

### 4. `:1224` — hook accounts put a single attestation at 15 of the 16-account budget

For `swap_revert` on a hook mint the manifest is: reserve PDA, vault, mint, recipient, recipient ATA, custody token program, hook program, validation PDA, the real liqSOL hook's 5 declared metas, ATA program, system program = **15** of `MAX_TERMINAL_DYNAMIC_ACCOUNTS = 16`. `reserve_create_cancelled` is the same 15.

The greedy packer at line 1316 always takes the first attestation regardless of size, so one hook meta more — or one more effect account added program-side — makes a single attestation's manifest unserializable. `transaction::serialize` then throws locally, which the header itself documents as "an alarm, not a retry", and the epoch cannot settle at all.

Nothing pins this: `build_manifests_carry_transfer_hook_accounts_for_a_hook_mint` only checks membership and writability, and the existing packet-limit test measures a non-hook manifest.

### 5. `:1671` — an absent custody-mint read degrades to "no hook"

`mint_transfer_hook_for` treats an absent/empty custody-mint read as `nullopt`, and `mint_transfer_hook_program:694` likewise `break`s to `nullopt` on a truncated or desynced mint TLV. The function's own docstring refuses exactly that two lines later for the validation PDA: "A hook that IS configured but whose validation PDA cannot be read is NOT degraded to `nullopt`."

Scenario: an RPC returns null for the mint at the default commitment while the confirmed bank lags; the relay builds a hook-free manifest for a hook mint and the window aborts inside the CPI with no local signal. It self-heals on the next tick, so this is low severity — but the mint is a required effect account for a reserve that demonstrably exists, so "not found" is anomalous rather than benign, and the asymmetry with the sibling path looks unintentional.

---

Minor, not counted: the `.cpp` docstring at `:336` still says "the three fields" while the header was updated to four; and `SEED_INSTRUCTION_DATA` (~`:660`) is declared but never referenced.

valthon:
All five addressed in `9a773a55` (pushed). Two of them were genuine defects in the hook code from the previous commit, and on the first one your point stands on its own terms — the parser was undercutting the resolver's own stated contract.

### 1. Parser dropped metas silently

Right, and inconsistent with `resolve_hook_metas` two functions down, which throws on exactly this reasoning. All five shapes now `FC_ASSERT`: short account, entry overrunning the account, undersized entry, count disagreeing with the entry, and no Execute entry.

### 2. Meta loop bounded by account size, not entry extent

This was the serious one. `count` is untrusted, so a count larger than its entry walked into the following TLV bytes and pushed arbitrary pubkeys in as **writable** accounts. The walk is now bounded by the entry's extent, and a count that does not fit its entry is an error rather than something to parse past.

Pinned by `extra_account_meta_parser_refuses_malformed_accounts`, including a case whose trailing bytes are deliberately walkable — it fails when the entry-extent bound is removed, which I checked rather than assumed.

### 3. Uncached per-effect mint read

Fixed with a `std::map<solana_public_key, std::optional<mint_transfer_hook>>` beside the two sibling caches, memoising `nullopt` as well. Your 30-remits-over-3-reserves example goes from 33 reads back to 3.

### 4. 15 of 16 accounts

Confirmed your count independently — `hook_manifest_worst_case_fits_the_dispatch_budget` measures the real liqSOL five-meta declaration and gets exactly 15 for both `swap_revert` and `reserve_create_cancelled`.

Pinned **exactly**, not bounded: a `<=` check would still pass at 16 with zero headroom, and since the packer always takes at least one attestation regardless of size, going over does not defer — it throws locally and the epoch cannot settle. I did not change the packer; the failure is already loud, and what was missing was a test that fails before it can happen.

### 5. Absent custody-mint read degraded to "no hook"

Fixed — it throws now, consistent with the validation-PDA path in the same function. Agreed it is low severity and self-healing; the asymmetry was unintentional.

Also fixed: the `.cpp` docstring at `:336` and the unused `SEED_INSTRUCTION_DATA`.

---

**Verification.** Full plugin suite green. `flow-swap-non-native-tokens` run locally against a `nodeop` built from `9a773a5573` — SUCCEEDED, with `EffectAccountMissing` 0, custom program errors 0, and epochs advancing past 18 rather than freezing at 12. The one drain failure in that log is a connection-refused during nodeop shutdown after the validator is gone.

**Still not covered end-to-end:** no flow exercises a hook mint, so the hook path remains unit-tested only. The flow's mints are classic SPL, which makes that run a regression check on the non-hook path — worth having given #3 moved when the read happens for every SPL attestation, but it is not evidence about hook mints. Same gap program-side, per `tests.rs:2295`.

heifner:
Re-reviewed at `9a773a55`. Everything from the last round is resolved: the meta parser is strict, the mint read is cached, the absent-mint path asserts instead of degrading, both doc/dead-symbol nits are gone — and the new `hook_manifest_worst_case_fits_the_dispatch_budget` test is a good addition, since it turns the account budget into something measured rather than assumed.

Re-verified the hook resolution against `wire-solana` `896ed2fa`: the Execute prefix `[source, mint, destination, authority, validation_pda]` matches `spl_transfer_hook_interface::instruction::execute`; discriminators `0` / `1` / `>=0x80` and seed tags `0/1/3` match `ExtraAccountMeta` / `Seed`; `EXECUTE_TLV_DISCRIMINATOR` = `69 25 65 c5 4b fb 66 1a`; TLV base length 12 and mint TLV start 166 are right; and `accounts` growing as metas resolve matches `add_to_cpi_instruction`. The resolver was checked against the real liqSOL hook's five metas.

**Two worth changing before merge, three optional** — the rest of this is scoped so it can be one pass.

---

## Blocking

Both end the same way, which is what makes them worth the round trip: the dispatch window aborts, `drive_dispatch_rounds` repacks the next window from the on-chain cursor, the cursor never advanced, and the next attempt rebuilds the identical failing manifest. Not a retryable drop — a pinned cursor, with nothing behind it able to settle, and the cause visible only inside a Solana CPI.

### 1. `is_signer` is resolved and then discarded — `src/outpost_solana_client.cpp:853` → `:1242`

`resolve_hook_metas` preserves all three properties:

```cpp
resolved.push_back(fc::network::solana::account_meta{key, meta.is_signer, meta.is_writable});
```

but its only caller unpacks two:

```cpp
for (const auto& meta : resolve_hook_metas(...)) {
   add(meta.key, meta.is_writable);   // is_signer dropped
}
```

`ExtraAccountMeta::new_with_pubkey(&key, /*is_signer*/ true, ...)` is a legal validation-account entry. If one appears, `add_extra_accounts_for_execute_cpi` builds the Execute meta with `is_signer: true`, and `de_escalate_account_meta` only de-escalates against accounts already in the CPI instruction — a freshly appended extra keeps the flag, and `invoke_signed` aborts with "unauthorized signer".

The reason this one stands out is consistency, not likelihood: an `InstructionData` / `AccountData` seed gets `FC_ASSERT(false, "...unsupported seed form...")` a few lines up, and `PubkeyData` gets its own refusal, each commented as preferring a loud failure to a manifest short an account. A signer meta is the same "this builder cannot satisfy this shape" category and is the only one that passes through quietly. One `FC_ASSERT(!meta.is_signer, ...)` in the resolver turns an unexplained on-chain abort into a build-time error naming the meta index.

### 2. Two unused accounts consuming the last of the budget — `:1273` and `:1285`

`swap_revert` and the SPL branch of `reserve_create_cancelled` both end with:

```cpp
add(associated_token_program_id, false);
add(system_program_id, false);
```

Neither handler ever creates an ATA — both log-and-skip on `data_is_empty() && owner == system_program` (`inbound.rs:2721`, `:3023`), and `associated_token::create*` appears nowhere in `inbound.rs`. Pre-existing and harmless until now.

What changed is the headroom. This PR's own test pins the worst case at `15u` against `MAX_TERMINAL_DYNAMIC_ACCOUNTS = 16` — the real liqSOL hook with five metas, one slot spare. The packer always takes at least one attestation regardless of size, so exceeding 16 isn't split or deferred: `transaction::serialize` throws locally, which the header calls "an alarm, not a retry". liqSOL declaring a sixth extra meta — a program-side change nobody would think to check against this plugin — wedges the outpost. Dropping these two takes the worst case 15 → 13.

---

## Optional

**3. `:693` — two silent "no hook" fall-throughs in `mint_transfer_hook_program`.** A truncated TLV (`value + ext_len > size` → `break`) and a `TransferHook` entry with `ext_len < 64` (falling through the `ext_len >= 64` guard to `cursor = value + ext_len`) both return `nullopt`. For a mint that genuinely declares a hook that yields a hook-free manifest and the same pinned cursor. This is the outcome `parse_extra_account_metas` was just made strict to prevent ("a partial parse is the worst possible outcome") — a malformed `TransferHook` entry deserves the same `FC_ASSERT` rather than a fall-through to hook-free.

**4. `:1709` — the mint read runs for legacy-SPL reserves, where a Token-2022 extension is impossible.** `add_hook_accounts` calls `mint_transfer_hook_for` for every reserve-backed SPL effect regardless of `info.custody_token_program`. Two costs: a `get_account_info` per distinct mint per build, inside a loop whose deadline is only probed per effect; and the new `FC_ASSERT(mint_info.has_value() && !mint_info->data.empty(), ...)` failing the **entire** build — every attestation, not just this one — on a transient absent/empty read for a mint the relay previously never touched. Gating on `info.custody_token_program == fc::network::solana::system::program_ids::TOKEN_2022_PROGRAM` (`solana_system_programs.hpp:25`) makes it a true no-op for legacy mints, which is what the surrounding comments already describe it as.

**5. `include/sysio/outpost_solana_client_plugin/outpost_solana_client.hpp:330` — the header contradicts the implementation.** It says *"Returns empty when the account carries no `Execute` entry"*; the implementation ends with `FC_ASSERT(false, "ExtraAccountMetaList carries no Execute entry; ...")` (`.cpp:770`) and can never return empty. The sibling `resolve_hook_metas` doc correctly says it throws.

valthon:
Both blocking items fixed, all three optional ones taken, in `e8c1759c9b`.

**1. `is_signer` dropped.** Correct, and the consistency argument is the right one — it was the only "this builder cannot satisfy this shape" case that passed through quietly while the seed forms and the unknown discriminator already asserted. `resolve_hook_metas` now refuses a signer meta by index. Verified non-vacuously: with the assert removed the new test exits 201, so it catches the regression rather than passing by construction.

**2. Two unused accounts.** Confirmed before removing them — `associated_token` appears in `inbound.rs` only as `get_associated_token_address*`, never a create CPI, and both handlers log-and-skip on an uninitialized ATA (`:2721`, `:3023`). Dropped from `swap_revert` and the SPL branch of `reserve_create_cancelled`; `associated_token_program_id` had no other use so its binding went too. The pinned worst case moves **15 → 13**, so liqSOL could declare three more metas before the cliff instead of one. Test and its comment updated to the new number and why it moved.

**3. Silent "no hook" fall-throughs.** Agreed, and it is the same asymmetry `parse_extra_account_metas` was made strict for. Both now assert: the truncated entry, and a `TransferHook` entry below the 64 bytes its authority and program id require (named `TRANSFER_HOOK_EXT_MIN_LEN`). The `ext_type == 0` terminator and running off the end of the TLV stay as normal termination.

**4. Mint read for legacy SPL.** Taken, gated on `info.custody_token_program == TOKEN_2022_PROGRAM`. The read cost was the smaller half — the real problem is the one you named second: that assert fails the *entire* build, every attestation in the envelope, on a transient read for a mint that could not have carried an extension.

This one had a consequence worth flagging: **all five existing hook tests used a synthetic `measurement_pubkey` as the custody token program**, so the gate would have made them silently vacuous. They now use the real Token-2022 id — which means they were never actually asserting hook behaviour against Token-2022 custody until this commit.

**5. Header doc.** Fixed, and the same contradiction was in `mint_transfer_hook_for`'s own comment, which claimed an absent mint returns `nullopt` where the code asserts. Both corrected.

**Tests.** 88/88 cases, 495/495 assertions. Three added: the signer refusal, the legacy-SPL gate asserting zero reads through a counter on the reader, and direct coverage of the mint TLV walk — which had none at all, so #3 was untested in either direction before this.

**Still not covered end-to-end.** Unchanged from my last comment: no flow uses a hook mint, so the hook path remains unit-tested only, same gap as `tests.rs:2295` program-side. What I did re-run is `flow-swap-non-native-tokens` against a `nodeop` rebuilt from this commit (`v1.0.0-e8c1759c9b...`) — **SUCCEEDED, 197 steps, 0 failed**, epochs to 17, all four OPP directions balanced (18/17/18/17 artifacts) and zero fatal log signatures. That is a regression check on the **non-hook** path, which is exactly what #4 moved, since the mint read now does not happen at all for those classic-SPL mints.

heifner (approved):
Approving at `e8c1759c`. All five findings from the last round are addressed, and each one landed with a test.

Both blockers are exactly right: the `FC_ASSERT(!meta.is_signer, ...)` sits at the top of the `resolve_hook_metas` loop alongside the seed and discriminator refusals it belongs with, and the unused `associated_token_program_id` / `system_program_id` are gone from both SPL branches. The budget test moving `15u` -> `13u` is the part I'd point at: it proves the removal and keeps the ceiling measured rather than assumed, which is what made the one-slot margin visible in the first place.

The three optional items are correct too. `mint_transfer_hook_program` now fails against a mint it cannot read instead of reporting "no hook" — the same posture `parse_extra_account_metas` already took — and `TRANSFER_HOOK_EXT_MIN_LEN` names the 64 rather than leaving it a literal. The `TOKEN_2022_PROGRAM` gate makes the legacy path genuinely free and, more importantly, scopes the hard assert to mints where a hook is representable at all. `legacy_spl_custody_never_reads_the_mint_for_a_hook` asserting the *absence* of the read, rather than just the resulting manifest, is the right shape for that test.

Checked for stale bindings after the removals: `token_program_id`, `token_2022_program_id` and `system_program_id` are all still referenced (`system_program_id` being the all-zero native-custody marker).

